PR #4, opened 2019-11-04
Replace sympy dependency with ast

phargogh:
This PR replaces `sympy` with a combination of stdlib `ast` and `eval()`.  This approach allows us to continue to provide symbolic evaluation without the need for any 3rd-party dependencies, and so any expression supported by `eval` is supported by `pygeoprocessing.symbolic`.  Note that while we do analyze the parsed expression's AST to find the variable names, we don't do any other checking.  The usual warnings about using `eval` would apply here.  If the execution safety of the expression is really needed, I would suggest that we use `asteval` (https://github.com/newville/asteval) which acts as an interpreter as well as a syntax-tree walker.

I'm marking this PR as a draft until @richpsharp finishes getting travis-ci up and running for this repo.

A few extra notes about this PR: 
* Tests pass locally, but convolution tests have been updated to have a more tolerant number of places that they assert to.  I assume the difference on my workstation is due to the use of a different python architecture.
* I didn't comment about this change in `HISTORY.rst` as the symbolic evaluation has not yet been officially released.


phargogh:
[Compare `numexpr` and `asteval` for use in `pygeoprocessing`](https://trello.com/c/zfnpcPII/260-compare-numexpr-and-asteval-for-use-in-pygeoprocessing)

src/pygeoprocessing/symbolic.py
```diff
@@ -72,30 +70,31 @@ def evaluate_raster_calculator_expression(
...
-        'evaluating: %s\nactive symbols: %s\nraster_op:\n%s',
-        expression, active_symbols, raster_op_source)
+        'evaluating: %s\nactive symbols: %s\n',
+        expression, sorted(active_symbols))
     symbol_list, raster_path_band_list = zip(*[
         (symbol, raster_path_band) for symbol, raster_path_band in
         sorted(symbol_to_path_band_map.items()) if symbol in active_symbols])
 
+    missing_symbols = set(active_symbols) - set(symbol_list)
+    if len(missing_symbols) > 0:
```
richpsharp: Can I nitpick here and request a Pythonic

```suggestion
    if missing_symbols:
```

src/pygeoprocessing/symbolic.py
```diff
@@ -170,10 +171,11 @@ def _generic_raster_op(*arg_list):
     target_dtype = numpy.result_type(*[x.dtype for x in array_list])
     result = numpy.empty(arg_list[0].shape, dtype=target_dtype)
     nodata_list = arg_list[n:2*n]
-    func = arg_list[2*n]
+    expression = arg_list[2*n]
     target_nodata = arg_list[2*n+1]
     default_nan = arg_list[2*n+2]
     default_inf = arg_list[2*n+3]
+    kwarg_names = arg_list[-1]
```
richpsharp: Can I request the following to be consistent with the rest of the nodation of the indexes, the docstring, and avoiding a nasty future bug caused by a change in the length of the list?

```suggestion
    kwarg_names = arg_list[2*n+4]
```

src/pygeoprocessing/symbolic.py
```diff
@@ -188,10 +190,18 @@ def _generic_raster_op(*arg_list):
             raise ValueError(
                 "`target_nodata` is undefined (None) but there are nodata "
                 "values present in the input rasters.")
-        func_result = func(*[array[valid_mask] for array in array_list])
+        user_symbols = dict((symbol, array[valid_mask]) for (symbol, array) in
+                            zip(kwarg_names, array_list))
     else:
         # there's no nodata values to mask so operate directly
-        func_result = func(*array_list)
+        user_symbols = dict((symbol, array) for (symbol, array) in
```
richpsharp: Since we're Python 3 now, can I request dictionary comprehension here?

```suggestion
        user_symbols = {(symbol, array) for (symbol, array) in
                        zip(kwarg_names, array_list)}
```

src/pygeoprocessing/symbolic.py
```diff
@@ -188,10 +190,18 @@ def _generic_raster_op(*arg_list):
...
                 "`target_nodata` is undefined (None) but there are nodata "
                 "values present in the input rasters.")
-        func_result = func(*[array[valid_mask] for array in array_list])
+        user_symbols = dict((symbol, array[valid_mask]) for (symbol, array) in
+                            zip(kwarg_names, array_list))
     else:
         # there's no nodata values to mask so operate directly
-        func_result = func(*array_list)
+        user_symbols = dict((symbol, array) for (symbol, array) in
+                            zip(kwarg_names, array_list))
```
richpsharp: Won't need this if you like the above.

```suggestion
```

tests/test_geoprocessing.py
```diff
@@ -2587,7 +2587,8 @@ def test_convolve_2d_single_thread(self):
         # the sides and realizing diagonals got subtracted twice
         expected_result = test_value * (
             n_pixels ** 2 * 9 - n_pixels * 4 * 3 + 4)
-        self.assertAlmostEqual(numpy.sum(target_array), expected_result)
+        self.assertAlmostEqual(numpy.sum(target_array), expected_result,
+                              places=3)
```
richpsharp: I can't quite tell, is the indent off by one space?

```suggestion
                               places=3)
```

tests/test_geoprocessing.py
```diff
@@ -2624,7 +2625,8 @@ def test_convolve_2d_multiprocess(self):
...
```
richpsharp: Is this indent off by one space too?

```suggestion
                               places=3)
```

tests/test_geoprocessing.py
```diff
@@ -2692,7 +2695,8 @@ def test_convolve_2d_ignore_nodata(self):
 
         # calculate by working on some graph paper
         expected_result = 9*9*.5
-        self.assertAlmostEqual(numpy.sum(target_array), expected_result)
+        self.assertAlmostEqual(numpy.sum(target_array), expected_result,
+                               places=5)
```
richpsharp, line 2699: How are you choosing the decimal places here? There's a 3, 4, and 5. Is there a reason for that that's worth considering? ... Like, I'm worried if there's a numerical error happening.

richpsharp (changes requested):
Okay I finished my review. And I LOVE github's review process! Lemme know what you think. We don't even have to do that "back to you" thing that we used to do now.

tests/test_geoprocessing.py
phargogh: I selected those numbers because they were the maximum values I could select to achieve the same tolerance.  The differences were only observed on tests that used the 2d convolution, and I was observing them in my 64-bit conda environment (on windows).  I'd be happy to dig into this further if you'd like!  I do know that the tests were failing on my machine before I changed anything in the `symbolic` module, so these shouldn't have anything to do with the functional changes in this PR.

src/pygeoprocessing/symbolic.py
phargogh: Sure, happy to use a dict comprehension!  But the notation you mentioned will create a set of tuples and not a dict:

```
>>> type({(a, a) for a in range(5)})
<class 'set'>
>>> type({a: a for a in range(5)})
<class 'dict'>
```
I'll patch this shortly.

src/pygeoprocessing/symbolic.py
phargogh: Nearly!  The `{ ... }` notation doesn't quite work here because `{zip(...)}` yields a `set` with  `zip` object in it.

Instead, I've updated this to be `dict(zip(...))` and that works great and avoids the unnecessary comprehension.  Updated in 05b0195

src/pygeoprocessing/symbolic.py
phargogh: Updated in 05b0195

tests/test_geoprocessing.py
phargogh: Huh, no idea how that happened!  Patched in d916244.

tests/test_geoprocessing.py
phargogh: It sure is!  Not sure how that happened.  Patched in d916244.

PR #6, opened 2019-11-06
add travis CI testing 

richpsharp:
This PR adds a .travis-ci.yml configuration file to automatically test the PyGeoprocessing repository using a docker image that builds PyGeoprocessing and runs the test suite on the current commit.

The Docker image only uses Python 3.7 for now. If we were to extend it I think I'd like to make a docker image per Python version type, and for different architectures. But all of that seems secondary to having a functional test suite running on PRs.

Also note the rough versioning scheme I've set up for the docker image. Inside the `pygeoprocessing-docker` directory there is a file called `VERSION`. This will hold *alll* the version numbers of the docker image and will serve as a reminder to the next version to bump and what already exists. In turn the commit is tied to a particular version of the docker image through the `travis.yml` file. This is also very manual but I think quite functional and totally fine for most of our current needs.

Note I've also removed the Appveyor and Bitbucket Pipelines configuration file in this PR. I think Travis CI will serve all those needs for the future.


phargogh:
I agree, having a functional test suite takes precedent.  I was also looking through the travis docs and their support for [various python versions looks interesting](https://docs.travis-ci.com/user/languages/python/), so long as we can figure out the GDAL dependency.

Thanks for adding this!

PR #7, opened 2019-11-11
Release/1.9.0

phargogh:
As far as I can tell, this should match the 1.9.0 release on PyPI.  The 1.9.0 release was accidentally deleted as a part of the github migration (sorry!  I think I'm to blame for that one).

Unlike with pull requests with mercurial, tags don't come along for free.  If you approve of this PR as-is, we'll need to re-tag the pygeoprocessing mege rev with `1.9.0` accordingly.


richpsharp:
I'm approving, but is it weird the Travis-CI tests didn't run?

phargogh:
No, it doesn't seem weird that Travis didn't run.  I knowingly based this off of the revision of `develop` that didn't have the travis stuff in it in order to match the release onPyPI.

Also FYI @richpsharp , I'm still learning about how tags work in git, and so I just tagged `master` with `1.9.0` manually and pushed the new tags table.  The `1.9.0` release is now reflected in the repo.

We should look into if there are any better ways to do tagging as part of our release processes.

Issue #18, opened 2020-01-09
multiple processes sending logging messages to Qt’s dialog box on MacOS


davemfish:
There's an issue detailed over at https://bitbucket.org/natcap/invest/issues/3931/invest-370-sdr-crashes-hard-on-mac, which seems to boil down to multiple processes started by pygeoprocessing sending logging messages to Qt’s dialog box on MacOS. 

Pygeoprocessing does not use any type of Queue handler for logging, and it does use multiprocessing by default in some operations such as `align_and_resize`. Presumably a queue handler for the logger would resolve this and be generally a good idea, even though this problem only occurs in one very specific instance (that we know of).

```
natcap.invest.sdr.sdr 3.7.0.post839+h89ab3f9d1349
pygeoprocessing 1.9.1
MacOS 10.15.2 (Catalina)
```

```
01/08/2020 15:57:47  pygeoprocessing.geoprocessing INFO     starting stats_worker
The process has forked and you cannot use this CoreFoundation functionality safely. You MUST exec().
Break on __THE_PROCESS_HAS_FORKED_AND_YOU_CANNOT_USE_THIS_COREFOUNDATION_FUNCTIONALITY___YOU_MUST_EXEC__() to debug.
The process has forked and you cannot use this CoreFoundation functionality safely. You MUST exec().
Break on __THE_PROCESS_HAS_FORKED_AND_YOU_CANNOT_USE_THIS_COREFOUNDATION_FUNCTIONALITY___YOU_MUST_EXEC__() to debug.
The process has forked and you cannot use this CoreFoundation functionality safely. You MUST exec().
Break on __THE_PROCESS_HAS_FORKED_AND_YOU_CANNOT_USE_THIS_COREFOUNDATION_FUNCTIONALITY___YOU_MUST_EXEC__() to debug.
Fatal Python error: Illegal instruction

Current thread 0x000070000d4de000 (most recent call first):
  File "env/lib/python3.7/site-packages/natcap/invest/ui/inputs.py", line 291 in write
  File "env/lib/python3.7/logging/__init__.py", line 1028 in emit
  File "env/lib/python3.7/logging/__init__.py", line 894 in handle
  File "env/lib/python3.7/logging/__init__.py", line 1586 in callHandlers
  File "env/lib/python3.7/logging/__init__.py", line 1524 in handle
  File "env/lib/python3.7/logging/__init__.py", line 1514 in _log
[...]

PR #19, opened 2020-01-12
Bugfix/geoprocessing 18 remove multiprocessing

davemfish:
This PR removes the `multiprocessing` dependency in pygeoprocessing. The motivating issue was to avoid an occasional deadlock  that occurred on Mac OS X during ``align_and_resize_raster_stack``. See issue #18 .   

Also, in `convolve_d2`, some `multiprocessing` utilities were replaced with analogous `queue` and `threading` library objects. 

I'm not sure why the README.rst appears in this diff -- that change should already be on develop -- but it probably has to do with a rebase I did on my fork in order to fix incorrect issue numbers in my commit messages.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,7 +1,7 @@
 Release History
 ===============
 
-Unreleased Changes
+1.9.1 (2019-12-19)
 ------------------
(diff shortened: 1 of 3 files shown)


richpsharp (approved):
Awesome thanks @davemfish! I no longer have to carry that burden with me.

Issue #25, opened 2020-01-31
Windows tests crash with shapely 1.7.0


davemfish:
Shapely 1.7.0 was just released and we're seeing a variety of issues. I triggered the Actions test suite with a dummy commit over in my fork and observed errors like `ValueError: GEOSGeom_createLinearRing_r returned a NULL pointer` on Windows . 

See full logs here:
https://github.com/davemfish/pygeoprocessing/runs/418379492?check_suite_focus=true

Shapely does not describe any backwards incompatibility with this release, so presumably these are bugs that will be fixed by shapely. Probably the thing to do in the meantime is avoid `shapely 1.7.0`

davemfish:
resolved by #26

davemfish:
Re-opening because we have a better solution than avoiding this version of shapely entirely. We need only avoid a certain build on a certain PyPI.

dcdenu4:
@davemfish I posted to the Python 3.8 issue, but thought I'd bring it back over here. For Python 3.8 support we'll need a version of Shapely to match and it doesn't look like Golhke is hosting any at the moment. I just installed the `Shapely‑1.7.0‑cp38‑cp38‑win_amd64.whl` (I'm running Windows Python 3.8 64-bit) and all the Pygeo tests pass for me. I wonder if the dll issues have been sorted?

davemfish:
@dcdenu4 , it looks that whl for 3.8 is available from Gohlke. And in the end we figured out that it was only the 1.7.0 wheels on PyPI that were problematic. The Gohlke builds worked all along, it was just a matter of making sure pip installed them instead of the ones from PyPI.

PR #26, opened 2020-01-31
avoiding a problematic shapely version. #25.

davemfish:
A PR to avoid dependency on `shapely 1.7.0` Please see issue #25 for a description of the problem.

And this PR also now modifies the github Actions pipeline to install dependencies from requirements.txt

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -5,8 +5,7 @@ on: [push, pull_request]
 jobs:
   Test:
     runs-on: ${{ matrix.os }}
-    env:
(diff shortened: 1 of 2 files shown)


davemfish:
It looks like the github actions environment is not actually built from `requirements.txt`, since it was designed to be flexible to `pip` and `conda` package management, which `requirements.txt` is not. 

So I'll update that actions pipeline in this PR so there is only one source of truth for dependencies.

phargogh (approved):
Ah, so much nicer to just be able to use `conda` to install from `requirements.txt`!

richpsharp (approved):

PR #30, opened 2020-02-19
force a specific package index url for shapely on Windows.

davemfish:
It's not Shapely version 1.7.0 that's a problem for Windows, it's just the build on PyPI. So here's a solution for forcing the installation from a different package index. See issue #25

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -35,7 +35,10 @@ jobs:
       shell: bash
       # Replace numpy and scipy with PyPI versions to circumvent import issue.
       # https://stackoverflow.com/a/37110747/299084
+      #
+      # Shapely 1.7.0 on PyPI has GEOS cdll issues in some environments. Gohlke builds work.
       run: |
+          $CONDA/python -m pip install shapely --index-url http://pypi.naturalcapitalproject.org/simple/ -c requirements.txt
           $CONDA/python -m pip install -r requirements.txt
           $CONDA/python -m pip install $PACKAGES
           $CONDA/python setup.py install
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -6,7 +6,7 @@
 GDAL>=2.2.0,<3.0
 numpy>=1.10.1
 scipy>=0.14.1,!=0.19.1
-Shapely>=1.6.4,!=1.7.0
+Shapely>=1.6.4
```
phargogh, line 9: Great!  So nice to lift that version constraint.

.github/workflows/pythonpackage.yml
```diff
@@ -35,7 +35,10 @@ jobs:
       shell: bash
       # Replace numpy and scipy with PyPI versions to circumvent import issue.
       # https://stackoverflow.com/a/37110747/299084
+      #
+      # Shapely 1.7.0 on PyPI has GEOS cdll issues in some environments. Gohlke builds work.
       run: |
+          $CONDA/python -m pip install shapely --index-url http://pypi.naturalcapitalproject.org/simple/ -c requirements.txt
```
phargogh, line 41: Oh awesome ... this is so much better than only providing the `shapely` index.  Thanks for working this out!

phargogh (approved):
Looks great to me, Dave ... thanks!

phargogh:
@davemfish will we need to do this for InVEST as well?

davemfish:
@phargogh Yes, I think we'll do this for InVEST as well. I figured I would get your review on this one first, in case there were issues to resolve. I'll either do the invest PR after your Actions PR gets merged. Or if you have reason to update that PR and want to include this fix, go for it!

PR #31, opened 2020-02-24
Migration to GDAL 3

dcdenu4:
This PR migrates Pygeoprocessing to GDAL 3 and drops GDAL 2. 

GDAL 3 comes with some major changes which can be viewed here: 
- https://trac.osgeo.org/gdal/wiki/Release/3.0.0-News
- https://gdalbarn.com/#why
- https://trac.osgeo.org/gdal/wiki/rfc73_proj6_wkt2_srsbarn

The main points from the barn raising:

> “Coordinate systems in GDAL, PROJ, and libgeotiff are missing modern capabilities and need a thorough refactoring:
The dreaded ad hoc CSV databases in PROJ_LIB and GDAL_DATA are frustrating for users, pose challenges for developers, and impede interoperability of definitions.
GDAL and PROJ are missing OGC WKT2 support.
PROJ 5.0+ no longer requires datum transformation pivots through WGS84, which can introduce errors of up to 2m, but the rest of the tools do not take advantage of it.”

The major changes that actively affect Pygeoprocessing is the datum transformation through WGS84. Coordinate transformations now expect lat,lon instead of lon,lat. The recommendation is to use `osr.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)` which will allows for GIS friendly usage.

The proposed solution is to check Spatial References when creating coordinate transformations and if the spatial reference is unprojected, apply the migration Axis recommendation.

One hard coded test was updated to reflect slight variance in extents.


richpsharp:
Hey Doug, just FYI can you include me on pull requests on PyGeoprocessing in the future?

HISTORY.rst
```diff
@@ -8,6 +8,7 @@ Unreleased Changes
   pygeoprocessing source distribution.  This fixes an issue where files
   matching a variety of extensions anywhere in the pygeoprocessing directory
   might be included with the source distribution.
+* Adding GDAL 3 support and dropping GDAL 2 support.
```
richpsharp: Can you specify the things that will be different between GDAL 2 and 3? Seems like it's on the coordinate transformation order for unprojected SRSs. I remember in Slack you'd linked to a mailing list discussion or somesuch about why they decided to make this change. I think that'd be good to include here too.

tests/test_geoprocessing.py
```diff
@@ -2448,8 +2448,7 @@ def test_transform_box(self):
         # plotting it in a GIS polygon, so the expected result below is
         # regression data
         expected_result = [
-            453189.3366727062, 4918131.085894576,
-            468484.1637522648, 4952660.678869661]
+            453188.671769, 4918131.799327, 468483.727558, 4952661.553935]
```
richpsharp: I assume the difference here is because of numerical roundoff or something? It does seem like it's within `1e-5` or so. Maybe rather that putting the exact decimal to round to the expected cutoff accuracy? (I know it wasn't like that before but worth saying what's really expected since you're in there)...

richpsharp: Might be related to this update? Seems like a more accurate transform? "Off by a couple of meters" caught my eye. https://trac.osgeo.org/gdal/wiki/rfc73_proj6_wkt2_srsbarn#WGS84Pivot

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1630,8 +1630,18 @@ def reproject_vector(
...
+    # Create a coordinate transformation [GDAL 2 NOT supported]
+    # In GDAL 3.0 spatial references have been updated. One difference is 
+    # how transforms work with lat/lon coordinate systems. Transforms
+    # expect (lat, lon) order in GDAL 3.0 as opposed to thge GIS friendly
+    # (lon, lat). Thus if using a (lat, lon) srs set for backwards 
+    # compatability. 
+    if not base_sr.IsProjected():
+        base_sr.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)
+    if not target_sr.IsProjected():
+        target_sr.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)
```
richpsharp: ~~If I understand the issue correctly, in this function everything is contained inside of `osr.Geometry` objects which should be internally consistent with whatever the axis order is for whatever the projection is. We shouldn't encounter an x vs. y vs lat vs long because it's not exposed at this level. Is that possible? Or did you make this change because of an issue in InVEST or something?~~

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2722,10 +2732,23 @@ def transform_bounding_box(
...
+    # Create a coordinate transformation [GDAL 2 NOT supported]
+    # In GDAL 3.0 spatial references have been updated. One difference is 
+    # how transforms work with lat/lon coordinate systems. Transforms
+    # expect (lat, lon) order in GDAL 3.0 as opposed to thge GIS friendly
+    # (lon, lat). Thus if using a (lat, lon) srs set for backwards 
+    # compatability. 
+    if not base_ref.IsProjected():
+        base_ref.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)
+    if not target_ref.IsProjected():
+        target_ref.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)
```
richpsharp: Reading through https://trac.osgeo.org/gdal/wiki/rfc73_proj6_wkt2_srsbarn#Axisorderissues, I think the test for not [...].IsProjected(): might not be necessary, and in some cases wrong. Like some kind of exotic projected transformation that swaps its coordinates. From what I understand in the case here where we're always assuming x/y order, we should assume the osr.OAMS_TRADITIONAL_GIS_ORDER for both transforms. Is that right?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1630,8 +1630,18 @@ def reproject_vector(
...
+    # how transforms work with lat/lon coordinate systems. Transforms
+    # expect (lat, lon) order in GDAL 3.0 as opposed to thge GIS friendly
+    # (lon, lat). Thus if using a (lat, lon) srs set for backwards 
+    # compatability. 
+    if not base_sr.IsProjected():
+        base_sr.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)
+    if not target_sr.IsProjected():
+        target_sr.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)
+        
+    coord_trans = osr.CreateCoordinateTransformation(base_sr, target_sr)
```
richpsharp, line 1656: But do should keep `CreateCoordinateTransformation` because that's a GDAL 3.0 thing!

src/pygeoprocessing/geoprocessing.py
richpsharp: ~~p.s. I'm not totally sure about this, but I'd want to make sure we know why we're setting the axis mapping strategy. If there's a failed test or something that's obvious, otherwise I keep thinking "how do we know if it's lat/lng or x/y". In the bounding box routine below we really don't know. But here all the geometry should know and `ogr` should be internally consistent?~~

HISTORY.rst
richpsharp: Can you indicate that the only non-backwards compatible issues in GDAL 2 to GDAL 3 were the introduction of `osr.CreateCoordinateTransformation` (vs `CoordinateTransformation`) and the need to set `[srs_ref].osr.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)` because of https://trac.osgeo.org/gdal/wiki/rfc73_proj6_wkt2_srsbarn#Axisorderissues? I think this will only be an issue for people who manually construct

richpsharp (changes requested):
One last thing Doug, since most of these issues are all about lat vs. long or not, can you add test cases for transform_bounding_box and reproject_vector for testing transforms from utm->utm, wgs84->wgs84, wgs84->utm, and I think there's already one for utm->wgs84, but if not add that too?

richpsharp:
One last thing Doug, since most of these issues are all about lat vs. long or not, can you add test cases for `transform_bounding_box` and `reproject_vector` for testing transforms from utm->utm, wgs84->wgs84, wgs84->utm, and I think there's already one for utm->wgs84, but if not add that too?

src/pygeoprocessing/geoprocessing.py
dcdenu4: Hey Rich, I just spent a significant amount of time in the WKT2 docs, PROJ 6 docs, and GDAL 3 Docs. I _think_ I'm starting to develop a deeper understanding. I think the basics are as follows:

- GDAL 3 conforms to the authority defining coordinate reference systems (CRS). 
- For Geographic CRS (WGS84), this means no longer ignoring Axis Order but abiding by LAT,LON order
- For Projected CRS (UTM), most EPSG in the registry use Easting, Northing order
- We think of things in general as **x**,**y** because it is convenient, but now a `Transform` expects Axis Order, where  if `long=x` and `lat=y`, GDAL 3 would take `Transform(y, x)`
- Thus when transforming from Projected CRS to Geographic CRS, `Transform(x,y)` will return `y,x` (Lat,Lon)
- Thus when transforming from Geographic CRS to Projected CRS, `Transform(y,x)` will return `x,y` (Easting, Northing)
- Thus when transforming from Geographic CRS to Geographic CRS, `Transform(y,x)` will return `y,x` (Lat,Lon)
- Thus when transforming from Projected CRS to Projected CRS, `Transform(x,y)` will return `x,y` (Easting, Northing)

**note:** Most projected CRS use Easting first, Northing second convention, however, some defined in the EPSG registry use the reverse convention.

So, I think you're correct in that `not.____.IsProjected()` is probably wrong, but `_____.IsGeographic()` _could_ be OK. Basically that Axis Order setting `SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)` is saying, if we've got a Geographic CRS, use Lon,Lat like the good old days (since we think of `lon,lat` as `x,y`. We certainly can, but maybe should, **NOT** do that. 

Perhaps a better solution is to create a mapping of Axis Order to Geographic (Lat,Lon) and Projected (Easting, Northing), such that when handling the extents and bounding box points we can correctly return _x_, _y_.

dcdenu4:
Hey @richpsharp, I just spent a significant amount of time in the WKT2 docs, PROJ 6 docs, and GDAL 3 Docs. I _think_ I'm starting to develop a deeper understanding. I think the basics are as follows:

- GDAL 3 conforms to the authority defining coordinate reference systems (CRS). 
- For Geographic CRS (WGS84), this means no longer ignoring Axis Order but abiding by LAT,LON order
- For Projected CRS (UTM), most EPSG in the registry use Easting, Northing order
- We think of things in general as **x**,**y** because it is convenient, but now a `Transform` expects Axis Order, where  if `long=x` and `lat=y`, GDAL 3 would take `Transform(y, x)`
- Thus when transforming from Projected CRS to Geographic CRS, `Transform(x,y)` will return `y,x` (Lat,Lon)
- Thus when transforming from Geographic CRS to Projected CRS, `Transform(y,x)` will return `x,y` (Easting, Northing)
- Thus when transforming from Geographic CRS to Geographic CRS, `Transform(y,x)` will return `y,x` (Lat,Lon)
- Thus when transforming from Projected CRS to Projected CRS, `Transform(x,y)` will return `x,y` (Easting, Northing)

**note:** Most projected CRS use Easting first, Northing second convention, however, some defined in the EPSG registry use the reverse convention.

So, I think you're correct in that `not.____.IsProjected()` is probably wrong, but `_____.IsGeographic()` _could_ be OK. Basically that Axis Order setting `SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)` is saying, if we've got a Geographic CRS, use Lon,Lat like the good old days (since we think of `lon,lat` as `x,y`. We certainly can, but maybe should, **NOT** do that. 

Perhaps a better solution is to create a mapping of Axis Order to Geographic (Lat,Lon) and Projected (Easting, Northing), such that when handling the extents and bounding box points we can correctly return _x_, _y_.

src/pygeoprocessing/geoprocessing.py
richpsharp: I think it's always okay to use `srs.SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER)` in this case. It'll only "swap" x/y if it's in a coordinate system that needs it. UTM projections are not affected. At least that's how I read the docs, and also I tried it manually and it behaved as expected. Is that okay?

dcdenu4: Yea, I think that's right!

HISTORY.rst
dcdenu4: I definitely will, but wanted to note that `osr.CoordinateTransformation` still works in GDAL 3, but since we're not supporting GDAL 2, can move on to the newer recommended function.

phargogh:
Hey @dcdenu4 I'm sorry I haven't had a chance to take a look at this closely (hoping to do so this afternoon!) but since this is kind of a forward-facing change, could you change the PR merge target into the `release/2.0` branch instead of `master`?  That will help us to continue to get bugfixes out on the 1.9.x release cycle and allow us to collect a few features for the 2.0 release in the near future.

Thanks!

src/pygeoprocessing/geoprocessing.py
richpsharp: RE: a thread we had on slack, yes this is necessary, but not necessary to have the `if` statements in there. And I have one other overall PR comment related to this I'll put up top.

richpsharp:
Hi @dcdenu4, another thing I'm hoping you can add: `pygeoprocessing` makes a lot of decisions about global state constants but exposes all of them to the user in one form or another. A big example of this is PyGeoprocessing's default to create GeoTIFFs and with LZW compressed bigtiff creation. That's almost always fine, but when it's not we don't want to build a separate version of `pygeoprocessing` (Peter did this once and me too). That's why we exposed the default as `raster_creation_options` in every `pygeoprocessing` function that creates a raster. 

Keeping with that convention, as part of this PR could you make a global default for the axis mapping strategy to `osr.OAMS_TRADITIONAL_GIS_ORDER` but expose that parameter in every function that touches any `pygeoprocessing` function that needs to do a transform? That'd be `reproject_vector` and `transform_bounding_box` because they do it directly, but also `align_and_resize_raster_stack` and `warp_raster` because they make a call to `transform_bounding_box`? And anywhere else you see that might be missing?

src/pygeoprocessing/geoprocessing.py
dcdenu4: Yep, totally agree.

dcdenu4: Yep, totally agree.

richpsharp:
Hi @dcdenu4 and @phargogh, can I ping this PR to keep it moving? I'm assuming it's held up on those failing tests right? But those tests are failing on Ubuntu specifically where there's an EPSG code that doesn't import from `pygeoprocessing.testing`. This is not a failure of PyGeoprocessing at all, but of an arbitrary import of an EPSG code that would have the same issue whether someone was using PyGeoprocessing or not. Since we're proposing removing `pygeoprocessing.testing` entirely, could we ignore those ubuntu test failures and proceed with this PR?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -574,6 +578,9 @@ def align_and_resize_raster_stack(
             name string as the first element and a GDAL creation options
             tuple/list as the second. Defaults to a GTiff driver tuple
             defined at geoprocessing.DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS.
+        osr_axis_mapping_strategy (OSR object): The OSR Axis Mapping Strategy 
+            as the first element. Defaults to a GIS friendly Lon,Lat order
```
richpsharp: It's not an OSR object, rather just an integer, can you change the type to (int)? Also the mention of "first element" sounds like it's a tuple. Could this read simply as:

```
osr_axis_mapping_strategy (int): OSR axis mapping strategy for `SpatialReference` objects. Defaults to 
    `geoprocessing.DEFAULT_OSR_AXIS_MAPPING_STRATEGY`. This parameter should not be changed unless you 
    know what you are doing.
```

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1581,6 +1589,9 @@ def reproject_vector(
             reprojection step. If it is an iterable, it will contain the
             field names to exclusively copy. An unmatched fieldname will be
             ignored. If ``False`` no fields are copied into the new vector.
+        osr_axis_mapping_strategy (OSR object): The OSR Axis Mapping Strategy 
+            as the first element. Defaults to a GIS friendly Lon,Lat order 
+            defined at geoprocessing.DEFAULT_OSR_AXIS_MAPPING_STRATEGY.
```
richpsharp: Could you change the docstring to the one I proposed above?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -503,7 +506,8 @@ def align_and_resize_raster_stack(
         target_pixel_size, bounding_box_mode, base_vector_path_list=None,
         raster_align_index=None, base_sr_wkt_list=None, target_sr_wkt=None,
         vector_mask_options=None, gdal_warp_options=None,
-        raster_driver_creation_tuple=DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS):
+        raster_driver_creation_tuple=DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS,
+        osr_axis_mapping_strategy=DEFAULT_OSR_AXIS_MAPPING_STRATEGY):
```
richpsharp, line 516: 👍

src/pygeoprocessing/geoprocessing.py
```diff
@@ -84,6 +84,9 @@
     gdal.GDT_CFloat64: numpy.complex64,
 }
 
+# GDAL 3 OSR Coordinate Transformation Axis Order strategy Lon,Lat
+DEFAULT_OSR_AXIS_MAPPING_STRATEGY = osr.OAMS_TRADITIONAL_GIS_ORDER
```
richpsharp: 👍

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1629,9 +1640,18 @@ def reproject_vector(
...
-
     # Create a coordinate transformation
-    coord_trans = osr.CoordinateTransformation(base_sr, target_sr)
+    # In GDAL 3.0 spatial references no longer ignore Geographic CRS 
+    # Axis Order and conform to Lat first, Lon Second. Transforms
+    # expect (lat, lon) order as opposed to the GIS friendly (lon, lat). 
+    # See https://trac.osgeo.org/gdal/wiki/rfc73_proj6_wkt2_srsbarn Axis order
+    # issues. SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER) swaps 
+    # the axis order, which will use Lon,Lat order for Geographic CRS, 
+    # but otherwise leaves Projected CRS alone.
```
richpsharp: I think this comment isn't relevant here anymore since we pass in the axis mapping strategy as a parameter, but I think it may be helpful at the top of the module where you define the default. Would you mind moving it to the single default instance at the top?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1815,6 +1836,9 @@ def warp_raster(
             name string as the first element and a GDAL creation options
             tuple/list as the second. Defaults to a GTiff driver tuple
             defined at geoprocessing.DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS.
+        osr_axis_mapping_strategy (OSR object): The OSR Axis Mapping Strategy 
+            as the first element. Defaults to a GIS friendly Lon,Lat order 
+            defined at geoprocessing.DEFAULT_OSR_AXIS_MAPPING_STRATEGY.
```
richpsharp: Could you change the docstring to the one I proposed above?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2709,6 +2734,10 @@ def transform_bounding_box(
             bounding box edge to sample along. A value of 2 will sample just
             the corners while a value of 3 will also sample the corners and
             the midpoint.
+        osr_axis_mapping_strategy (OSR object): The OSR Axis Mapping Strategy 
+            as the first element. Defaults to a GIS friendly Lon,Lat order 
+            defined at geoprocessing.DEFAULT_OSR_AXIS_MAPPING_STRATEGY.
+
```
richpsharp: Are there too many lines here w/r/t PEP 257?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2722,10 +2751,23 @@ def transform_bounding_box(
...
 
-    transformer = osr.CoordinateTransformation(base_ref, target_ref)
+    # Create a coordinate transformation
+    # In GDAL 3.0 spatial references no longer ignore Geographic CRS 
+    # Axis Order and conform to Lat first, Lon Second. Transforms
+    # expect (lat, lon) order as opposed to the GIS friendly (lon, lat). 
+    # See https://trac.osgeo.org/gdal/wiki/rfc73_proj6_wkt2_srsbarn Axis order
+    # issues. SetAxisMappingStrategy(osr.OAMS_TRADITIONAL_GIS_ORDER) swaps 
+    # the axis order, which will use Lon,Lat order for Geographic CRS, 
+    # but otherwise leaves Projected CRS alone
```
richpsharp: Same comment as above

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2722,10 +2751,23 @@ def transform_bounding_box(
...
+    base_ref.SetAxisMappingStrategy(osr_axis_mapping_strategy)
+    target_ref.SetAxisMappingStrategy(osr_axis_mapping_strategy)
+    
+    transformer = osr.CreateCoordinateTransformation(base_ref, target_ref)
 
     def _transform_point(point):
         """Transform an (x,y) point tuple from base_ref to target_ref."""
+        # NOTE: In GDAL 3, if osr.OAMS_TRADITIONAL_GIS_ORDER not set then
+        # TransformPoint expects Lat,Lon and NOT Lon,Lat for Geographic CRS.
```
richpsharp: This is only true if the projection system has the {2, 1} axis order like WGS84 has. Could you either remove this comment or point out something more general like "TransformPoint's axis order is dependent on the `SetAxisMappingStrategy` set above".

src/pygeoprocessing/testing/sampledata.py
```diff
@@ -113,6 +113,11 @@ def projection_wkt(epsg_id):
     origin=(443723.127327877911739, 4956546.905980412848294),
     pixel_size=lambda x: (x, -1. * x)
 )
+SRS_WGS84 = ReferenceData(
+    projection=projection_wkt(4326),
+    origin=(-123.71107369348861, 44.76009901797307),
+    pixel_size=lambda x: (x, -1. * x)
+)
```
richpsharp: Is this used anywhere? I tried to "ctrl+f" it and nothing turned up. If not, can you remove it?

tests/test_geoprocessing.py
```diff
@@ -285,8 +294,314 @@ def test_reproject_vector(self):
...
+        self.assertTrue(layer_defn.GetFieldCount(), 0)
         layer = None
         vector = None
+
+    def test_reproject_vector_latlon_to_utm(self):
+        """PGP.geoprocessing: reproject vector from lat/lon to utm."""
+        import pygeoprocessing
+        # Create polygon shapefile to reproject
+        base_srs = osr.SpatialReference()
+        base_srs.ImportFromEPSG(4326) # WGS84
```
richpsharp: Can you add a space for the comment re: "PEP257, two spaces before inline # comment"?

tests/test_geoprocessing.py
```diff
@@ -285,8 +294,314 @@ def test_reproject_vector(self):
...
+                osr.SpatialReference(target_reference.ExportToWkt())))
+        self.assertTrue(layer_defn.GetFieldCount(), 1)
+        self.assertEqual(layer_defn.GetFieldIndex('id'), 0)
+
+    def test_reproject_vector_latlon_to_latlon(self):
+        """PGP.geoprocessing: reproject vector from lat/lon to utm."""
+        import pygeoprocessing
+        # Create polygon shapefile to reproject
+        base_srs = osr.SpatialReference()
+        base_srs.ImportFromEPSG(4326) # WGS84
```
richpsharp: two space before #

tests/test_geoprocessing.py
```diff
@@ -2466,6 +2712,101 @@ def test_transform_box(self):
             'format should be [xmin, ymin, xmax, ymax]: '
             '%s' % gibraltar_bb_wgs84)
 
+    def test_transform_box_latlon_to_utm(self):
+        """PGP.geoprocessing: test geotransforming lat/lon box to UTM19N."""
+        import pygeoprocessing
+
+        # lat/lon bounds for Coast of New England from shapefile created in
+        # QGIS 3.10.2 running GDAL 3.0.3
+        bounding_box = [-72.3638439, 40.447948, -68.0041948, 43.1441579]
```
richpsharp, line 2723: Cool

tests/test_geoprocessing.py
```diff
@@ -2466,6 +2712,101 @@ def test_transform_box(self):
...
+        result = pygeoprocessing.transform_bounding_box(
+            bounding_box, wgs84_srs.ExportToWkt(), target_srs.ExportToWkt())
+        # Expected result taken from QGIS UTM19N - WGS84 reference and
+        # converting extents from above bounding box (extents) of shapefile
+        expected_result = [
+            214722.122449, 4477484.382162, 584444.275934, 4782318.029707]
+
+        self.assertIs(
+            numpy.testing.assert_allclose(
+                result, expected_result), None)
```
richpsharp, line 2740: Ha!

richpsharp (changes requested):
Great @dcdenu4, I only have some comments on the docstrings and comments. I hope you don't mind moving those around a little bit! Let me know if anything seems unreasonable okay?

richpsharp:
> Hi @dcdenu4 and @phargogh, can I ping this PR to keep it moving? I'm assuming it's held up on those failing tests right? But those tests are failing on Ubuntu specifically where there's an EPSG code that doesn't import from `pygeoprocessing.testing`. This is not a failure of PyGeoprocessing at all, but of an arbitrary import of an EPSG code that would have the same issue whether someone was using PyGeoprocessing or not. Since we're proposing removing `pygeoprocessing.testing` entirely, could we ignore those ubuntu test failures and proceed with this PR?

Oops I guess this was actually waiting on me. I don't know how I'm supposed to have notice there were changes pushed but I just happened to look at another tab. Sorry it was waiting on me, but also what do you think about the failing ubuntu test case points I made above?

dcdenu4:
@richpsharp Thanks for taking another look! Last Friday I was able to get the GDAL_DATA and PROJ_LIB environment variables properly setting with using the [**setup-miniconda**](https://github.com/goanpeca/setup-miniconda) Github Actions service as opposed to the **setup-conda** Github Actions service. The **setup-miniconda** version has more stars and is maybe better developed? I'm thinking it might be worth switching to use that resource if subsequent Windows and Mac tests don't mind. 

But, I'm going to focus on this PR today and I will ping you when more updates get pushed.

phargogh:
> Last Friday I was able to get the GDAL_DATA and PROJ_LIB environment variables properly setting with using the setup-miniconda Github Actions service as opposed to the setup-conda Github Actions service. The setup-miniconda version has more stars and is maybe better developed? I'm thinking it might be worth switching to use that resource if subsequent Windows and Mac tests don't mind.

Personally, I can't believe I didn't find the `setup-miniconda` action.  It looks better-developed, better-documented and better-supported than the `setup-conda` action, so I'd be very happy to use it if it works well!

.github/workflows/pythonpackage.yml
```diff
@@ -12,16 +12,20 @@ jobs:
...
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
+      with:
+          # Fetch all history so that setuptool_scm can build the correct version string.
+          fetch-depth: 0
+
     - name: setup-conda
       uses: s-weigand/setup-conda@v1.0.0
```
phargogh, line 24: If `setup-miniconda` is preferable, should we be using it here as well?

src/pygeoprocessing/geoprocessing.py
phargogh: Very nice ... consistent with the existing pygeoprocessing defaults!

src/pygeoprocessing/geoprocessing.py
phargogh: I'd second this, with one *trivial* change, which is that our docstrings are RST, so inline literal code should be surrounded by double-backticks.

tests/test_geoprocessing.py
```diff
@@ -285,8 +294,314 @@ def test_reproject_vector(self):
...
+        import pygeoprocessing
+        # Create polygon shapefile to reproject
+        base_srs = osr.SpatialReference()
+        base_srs.ImportFromEPSG(3157) # NAD83(CSRS) / UTM zone 10N
+        extents = [443723.1273278, 4956276.905980, 443993.1273278, 4956546.905980]
+
+        polygon_a = shapely.geometry.Polygon([
+            (extents[0], extents[3]), (extents[0], extents[1]),
+            (extents[2], extents[1]), (extents[2], extents[3]),
+            (extents[0], extents[3])])
```
phargogh: It looks like this is constructing a box using the `extents` noted above.  What you have here totally works, but I just wanted to  mention that `shapely` has a convenience function [`shapely.geometry.box`](https://shapely.readthedocs.io/en/latest/manual.html#shapely.geometry.box) that will do the same thing in a slightly more compact notation.

tests/test_geoprocessing.py
```diff
@@ -285,8 +294,314 @@ def test_reproject_vector(self):
...
+        test_feature.SetGeometry(test_geometry)
+        test_feature.SetField(field_name, 0)
+        test_layer.CreateFeature(test_feature)
+
+        test_layer = None
+        test_vector = None
+        test_driver = None
+
+        target_reference = osr.SpatialReference()
+        target_reference.ImportFromEPSG(3157) # UTM zone 10N
```
phargogh: There's another `PEP8` instance here for a comment.  There might be others here in this file as well ... could you try running `pycodestyle` (`pip install pycodestyle`) to check?

phargogh (commented):
Hey @dcdenu4 I just went through this and added a few small comments.  Since @richpsharp has been much more involved in this in this package, I'll defer to him on the final PR approval.

.github/workflows/pythonpackage.yml
dcdenu4: This is a great question and I have some updates about this. But I'm leaning towards Rich's comments that maybe that should be a separate issue and PR, because it's currently complicated. Complicated in that `setup-miniconda` works for `mac` and `linux` but now fails on `windows` because of how Miniconda is added to Windows Path.

src/pygeoprocessing/testing/sampledata.py
dcdenu4: Good point! I think I added it when I was creating earlier tests before we talked about having the `testing` module go away. And then I refactored those tests!

tests/test_geoprocessing.py
dcdenu4: Done here: [f2c337b](https://github.com/natcap/pygeoprocessing/pull/31/commits/f2c337bce218e8c7f7d922edf0eb32f4d962f6df)
Thanks for the `pycodestyle` tool!

tests/test_geoprocessing.py
dcdenu4: Great suggestion done in commit: [f2c337b](https://github.com/natcap/pygeoprocessing/pull/31/commits/f2c337bce218e8c7f7d922edf0eb32f4d962f6df)

tests/test_geoprocessing.py
dcdenu4: I think I got pep fixes in [f2c337b](https://github.com/natcap/pygeoprocessing/pull/31/commits/f2c337bce218e8c7f7d922edf0eb32f4d962f6df)

src/pygeoprocessing/testing/sampledata.py
dcdenu4: Removed in [03b39f](https://github.com/natcap/pygeoprocessing/pull/31/commits/03b39fec8d4b72c9b1315082b19c4c171f14c79c)

src/pygeoprocessing/geoprocessing.py
dcdenu4: removed in [779ec92](https://github.com/natcap/pygeoprocessing/pull/31/commits/779ec92c5b38e30880d6d986ccdb88ddf644cbf7)

src/pygeoprocessing/geoprocessing.py
dcdenu4: Done in [c9f0d6](https://github.com/natcap/pygeoprocessing/pull/31/commits/c9f0d66d3513cb8559367280e605b008d244f30c)

src/pygeoprocessing/geoprocessing.py
dcdenu4: Done in [c9f0d6](https://github.com/natcap/pygeoprocessing/pull/31/commits/c9f0d66d3513cb8559367280e605b008d244f30c)

src/pygeoprocessing/geoprocessing.py
dcdenu4: Done in [2eba123](https://github.com/natcap/pygeoprocessing/pull/31/commits/2eba123c585913539377eeaa05ebccdf13990a80)

src/pygeoprocessing/geoprocessing.py
dcdenu4: Done in [2eba123](https://github.com/natcap/pygeoprocessing/pull/31/commits/2eba123c585913539377eeaa05ebccdf13990a80)

src/pygeoprocessing/geoprocessing.py
dcdenu4: Done in [2eba123](https://github.com/natcap/pygeoprocessing/pull/31/commits/2eba123c585913539377eeaa05ebccdf13990a80)

dcdenu4:
@richpsharp @phargogh Just went through and addressed all the changes. Note that Ubuntu tests are still failing, but they work on Mac and Windows! I think that and the fact that I can get them to work using **setup-miniconda** (mentioned above) (but `windows-latest` fails using that), hopefully is enough to ignore that build/test issue for now and that it should be addressed in Issue #44. 

Let me know what else should be done for this PR! Thanks for hanging in there with me as I get back in the groove!

requirements.txt
```diff
@@ -3,7 +3,7 @@
 # This file records the packages and requirements needed in order for
 # pygeoprocessing to work as expected.
 
-GDAL>=2.2.0,<3.0
+GDAL==3.0.4
```
richpsharp: I'm sorry I didn't notice this before, we've had problems in the past when we've pegged the version of packages to an *exact* version. Could you change this to `GDAL>=3.0`? Unless there's something in a particular version that you don't want before 3.0.4? Otherwise we'll have a pain if we have a docker image with 3.0.2 that'll need a whole rebuild etc.etc..

richpsharp (changes requested):
Hi Doug, I'm sorry I had just one more thing on there which is the pegging of version 3.0.4 of GDAL in `requirements.txt`. If that exact version is not necessary can you loosen the requirements to >=3.0?

requirements.txt
dcdenu4: Hey @richpsharp , There were some bug fixes since GDAL 3.0.2 that have effected the precision of transforms, therefore I was using 3.0.4 because that is what I created my tests in, and at least one of them would otherwise fail in 3.0.2 by a decimal. Let me know if you'd like to talk about this more.
I made the change to `GDAL>=3.0.4` in [83bd15](https://github.com/natcap/pygeoprocessing/pull/31/commits/83bd1546903dfef0eb234fdcb740cb551261e82d)

richpsharp: Hi @dcdenu4, unless you think it needs to be 3.0.4 could you change it to `GDAL>=3.0.2`? (I've got a docker image that's using 3.0.2 that'd be a pain to rebuild, but I could if it was necessary).

davemfish (approved):
I took a look at these changes and don't have anything new to add.  Full disclosure: I didn't review the tests too closely.

dcdenu4:
@richpsharp @phargogh We decided to leave the GDAL version at 3.0.4 via slack because of the mentioned bug fixes that GDAL has released. Therefore, requesting a re-review.

.github/workflows/pythonpackage.yml
phargogh: Uffda.  Sounds great to have this be in a separate PR!

tests/test_geoprocessing.py
```diff
@@ -248,26 +249,37 @@ def test_reclassify_raster_empty_value_map(self):
...
+
+        # Create polygon shapefile to reproject
+        base_srs = osr.SpatialReference()
+        # NAD83(CSRS) / UTM zone 10N
+        base_srs.ImportFromEPSG(3157)
+        extents = [
+            443723.1273278, 4956276.905980, 443993.1273278, 4956546.905980]
+
+        polygon_a = shapely.geometry.box(
+            extents[0], extents[1], extents[2], extents[3])
```
phargogh: FYI, python's argument unpacking to even further abbreviate this: `polygon_a = shapely.geometry.box(*extents)`.  What you have here will do exactly the same thing!

phargogh (approved):
Looks good to me, Doug!  I noted one other thing about the use of `shapely.geometry.box`, but that's more of an FYI than anything else.  Up to you if you'd like to change that.

tests/test_geoprocessing.py
dcdenu4: You know I've seen you guys use this and thought about it, but must not have thought about it long enough to implement it. I will update it so that it helps me practice it.

dcdenu4: Done in [0f5e91](https://github.com/natcap/pygeoprocessing/pull/31/commits/0f5e91c7e1856f9ff8055b4bd8cff1516c412d99)

richpsharp (approved):

PR #49, opened 2020-04-13
Update actions to use setup-miniconda

### LINKED ISSUE #44: setup-conda action failing for Ubuntu tests
When working on Issue #16, migrating to GDAL 3 seems to cause the Ubuntu tests to fail using `setup-conda`. It appears that `setup-conda` does not properly set up that environment for **GDAL_DATA** and **PROJ_LIB** environment variables, which then give the following test errors:

```
ERROR 1: PROJ: proj_create_from_database: Open of /usr/share/miniconda/share/proj failed
ERROR 1: PROJ: proj_create_from_database: Open of /usr/share/miniconda/share/proj failed
```

It appears that even with GDAL 2 those environment variables were not being set properly, but the tests did not fail. One theory for this, is that before GDAL 3 EPSG codes were stored in a `csv` file in `gdal/data/` somewhere, opposed to GDAL 3 that now uses sqlite DBs and let's PROJ handle the codes.

dcdenu4:
With adding GDAL 3 support there was some breaking down of how the conda environments were being handled by **setup-conda** actions. Mostly `GDAL_DATA` and `PROJ_LIB` environment variables were not being set correctly. There is a work around for this in **setup-conda**, however I found that **setup-miniconda** (https://github.com/marketplace/actions/setup-miniconda) seems better developed and supported. They just released a new version that allows for channels and the gdal environment variables are properly set.

Fixes Issue #44


richpsharp (approved):
I don't know what this is but if it works I believe you!

davemfish (approved):
I approve these changes. It seems unrelated to this PR, but the Travis builds are failing. Since I'm not the most up-to-speed on the CI of this project, I'd rather defer to @richpsharp to actually merge this if that's the right thing to do.

richpsharp (changes requested):
Can you remove the Travis CI build like James requested?

dcdenu4:
@richpsharp I went into travis-CI and deselected the options to build on branch push and pull request, under the settings options. Would you like me to push something into this PR to make sure it doesn't build?

richpsharp:
> @richpsharp I went into travis-CI and deselected the options to build on branch push and pull request, under the settings options. Would you like me to push something into this PR to make sure it doesn't build?

Yes I think so? I see there's a `.travis.yml` file in the repo. I don't know how all of that triggers but I imagine it's a combination of those hooks you shut off in TravisCI and this file that's the configuration for it.

richpsharp (approved):

PR #50, opened 2020-04-15
Test and support Python 3.8

dcdenu4:
Adding Python 3.8 into the testing rotation and making a note in history.


.github/workflows/pythonpackage.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       fail-fast: false
       max-parallel: 4
       matrix:
-        python-version: [3.6, 3.7]
+        python-version: [3.6, 3.7, 3.8]
```
richpsharp: I remember james saying we only needed to support the two most recent versions of Python. If that's still true can we do that here?

richpsharp (changes requested):
ha that was short!

richpsharp (approved):

PR #56, opened 2020-04-16
Fixing signed byte warp issue + Parameters:

### LINKED ISSUE #54: Signed bytes converted to unsigned in `warp_raster`
If a signed byte raster is passed to `warp_raster` the result is incorrectly an unsigned byte raster.

### LINKED ISSUE #55: Change "Parameters:" to "Args:" in docstrings
To be consistent with the google style guide on docstrings, convert all instances of "Parameters:" to "Args:.

richpsharp:
This PR takes @davemfish's test for warp_raster from a signed byte to demonstrate the result is an unsigned type and provides a fix by explictily testing for it in the function. This is similar to how signed byte rasters are handled in the rest of PyGeoprocessing and fixes #54 

Also sneaking in a fix to change all the "Parameters:" docstrings to "Args:" in the main PyGeoprocessing modules to fix #55


src/pygeoprocessing/geoprocessing.py
```diff
@@ -3336,8 +3340,8 @@ def _convolve_2d_worker(
         work_queue, write_queue):
     """Worker function to be used by ``convolve_2d``.
 
-    Parameters:
-        Parameters:
+    Args:
+        Args:
```
davemfish, line 3344: Duplicate `Args:` line here.

tests/test_geoprocessing.py
```diff
@@ -4877,3 +4877,55 @@ def test_iterblocks_bad_raster(self):
...
+    def test_warp_raster_signedbyte(self):
+        """PGP.geoprocessing: warp raster test."""
+        import pygeoprocessing
+        import pygeoprocessing.testing
+
+        workspace_dir = 'test_workspace'
+        try:
+            os.makedirs(workspace_dir)
+        except OSError:
+            pass
```
davemfish, line 4890: How about using `self.workspace_dir` in this test instead? We already have setup and teardown methods for it in the class.

tests/test_geoprocessing.py
```diff
@@ -4877,3 +4877,55 @@ def test_iterblocks_bad_raster(self):
...
+        new_band = None
+        new_raster = None
+
+        target_raster_path = os.path.join(workspace_dir, 'target_a.tif')
+        base_a_raster_info = pygeoprocessing.get_raster_info(base_a_path)
+
+        pygeoprocessing.warp_raster(
+            base_a_path, base_a_raster_info['pixel_size'], target_raster_path,
+            'near', target_sr_wkt=wgs84_wkt, n_threads=1,
+            gdal_warp_options=['COPYMETADATA=TRUE'])
```
davemfish, line 4919: I think our testing concluded that explicitly adding the `['COPYMETADATA=TRUE']` has no effect and is unnecessary. If so, can we remove it here before future us thinks it's important?

davemfish (changes requested):
Thanks for this patch @richpsharp . I made a couple very small requests in the test, and caught one duplicate `Args` docstring line.

But my main question now is, should this PR go to `master` instead since it's a small bugfix?

src/pygeoprocessing/geoprocessing.py
richpsharp: removed in f7b456ca6eb90cb3da11894808e344f548d2ef2d

tests/test_geoprocessing.py
richpsharp: changed in 13a9e8b8e38369d40d866f1a46aaa7f3d551009c

tests/test_geoprocessing.py
richpsharp: removed in 1b1f9abae9470570cc276da1e76bfdbcf903db4f

richpsharp:
> Thanks for this patch @richpsharp . I made a couple very small requests in the test, and caught one duplicate `Args` docstring line.
> 
> But my main question now is, should this PR go to `master` instead since it's a small bugfix?

Hi @davemfish, regarding the above, I switched the target to master and got a lot of failing tests because `master` is still supporting python 3.6 to 3.7 but the linux 3.6 targets don't have a GDAL >= 3.0.4 available in their channels so all those tests fail. Would you mind deferring this merge into master until the release so I can avoid having to selectively merge parts of the release branch into master so that we can merge this PR into master so it can be merged back into the release branch?

richpsharp:
All the tests are failing, something is very broken

Issue #43, opened 2020-03-07
Remove `pygeoprocessing.testing`


richpsharp:
Remove the `pygeoprocessing.testing` module and rewrite unit tests to be independent of it. When implementing also note that InVEST uses some functions from this module can will need to have those addressed before the newer version of `pygeoprocessing` can be integrated there.

PR #63, opened 2020-04-30
Bugfix/convolve2 d 22 near zero error

richpsharp:
Fixes convolve 2d numerical error by providing a threshold that can be set with an absolute tolerance value.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -13,6 +13,8 @@ Unreleased Changes
   pygeoprocessing source distribution.  This fixes an issue where files
   matching a variety of extensions anywhere in the pygeoprocessing directory
   might be included with the source distribution.
+* Added ``set_tol_to_zero`` to ``convolve_2d`` to allow for in-function masking
+  of near-zero results to be set to 0.0.
 
 1.9.2 (2020-02-06)
 ------------------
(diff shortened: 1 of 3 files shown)


dcdenu4 (approved):
Hey @richpsharp this looks good to me, thanks for working on it. The only thing I'll bring is up is that we talked about removing `pygeoprocessing.testing` eventually [ issue #43 ]. So we'll have to do some rewriting for this test down the road.

I'll approve but NOT merge, in-case you want to take care of refactoring that test now. But, feel free to merge if you'd rather do it in a wider issue sweep!

richpsharp:
> Hey @richpsharp this looks good to me, thanks for working on it. The only thing I'll bring is up is that we talked about removing `pygeoprocessing.testing` eventually [ issue #43 ]. So we'll have to do some rewriting for this test down the road.
> 
> I'll approve but NOT merge, in-case you want to take care of refactoring that test now. But, feel free to merge if you'd rather do it in a wider issue sweep!

Good point @dcdenu4, I patched that test so it only uses `GDAL` for creating the test rasters.

dcdenu4 (approved):
Thanks Rich!

PR #68, opened 2020-05-04
Bugfix/convolve2 d 66 nodata undefined

### LINKED ISSUE #67: Convolve_2d returns different results if nodata is None
When testing a raster where nodata was not set, therefore equal to `None`, I noticed `convolve_2d` returned different results then the SAME raster with nodata set to `-1`, where both rasters didn't actually have any nodata values.

Both function calls used: `ignore_nodata=True` and `mask_nodata=False`

I sent a gist and test data to @richpsharp for confirmation of this bug.

richpsharp:
I changed the name of that parameter to be `ignore_nodata_and_edges` since that's really what it does, I fixed the places in the code that would incorrectly skip that functionality if the signal raster's nodata was None, added a test to cover it, and updated the history accordingly.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -12,6 +12,9 @@ Unreleased Changes (2.0)
   Lat,Lon but we use osr.OAMS_TRADITIONAL_GIS_ORDER to swap to Lon,Lat.
 * Using osr.CreateCoordinateTransformation() instead of
   osr.CoordinateTransformation() as the GDAL 3 call.
+* Fixed a bug in convolve_2d that would not ``ignore_nodata`` if the signal
+  raster's nodata value was undefined. Changed the name of this flag to
+  ``ignore_nodata_and_edges`` to reflect its expected functionality.
 
 Unreleased Changes (master)
 ---------------------------
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,8 +1,8 @@
 Release History
 ===============
 
-Unreleased Changes (2.0)
-------------------------
+Unreleased Changes
+------------------
 * Adding Python 3.8 support and dropping Python 3.6 support.
 * Adding GDAL 3 support and dropping GDAL 2 support. The only non-backwards
   compatible issue in GDAL 2 to GDAL 3 is the need to handle Axis Ordering with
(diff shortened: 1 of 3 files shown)


src/pygeoprocessing/geoprocessing.py
```diff
@@ -3415,11 +3422,15 @@ def _convolve_2d_worker(
         signal_block = signal_band.ReadAsArray(**signal_offset)
         kernel_block = kernel_band.ReadAsArray(**kernel_offset)
 
-        if signal_nodata is not None and ignore_nodata:
-            # if we're ignoring nodata, we don't want to add it up in the
-            # convolution, so we zero those values out
-            signal_nodata_mask = numpy.isclose(signal_block, signal_nodata)
-            signal_block[signal_nodata_mask] = 0.0
+        if ignore_nodata:
```
dcdenu4, line 3425: I think this `ignore_nodata` is fine because it's in the worker function, and should not be `ignore_nodata_and_edges` but wanted to check to make sure.

HISTORY.rst
```diff
@@ -1,8 +1,8 @@
 Release History
 ===============
 
-Unreleased Changes (2.0)
-------------------------
+Unreleased Changes
+------------------
```
dcdenu4, line 5: I think James added the two different `Unreleased` sections to better differentiate changes made. I guess it really doesn't matter because change remarks added for `release 2.0` won't make it into `master` but all the `master` change remarks will make it into the `release 2.0`. So it seems to me it's just to try and make it more human readable for us developers to keep track of what changes are specific to this release and what changes will be going out beforehand in a bugfix release. Anyway, just wanted to make sure you meant to change this.

dcdenu4 (commented):
Hey @richpsharp thanks for this. I added a few comments and then will be happy to merge this in.

HISTORY.rst
richpsharp: Yes I meant to change it.

src/pygeoprocessing/geoprocessing.py
richpsharp: Exactly, the meaning is different inside of the worker -- there's no edges there, only nodata.

richpsharp:
> Hey @richpsharp thanks for this. I added a few comments and then will be happy to merge this in.

I responded to your comments!

dcdenu4 (approved):

PR #69, opened 2020-05-06
fixes #59 removing manual set of fill value list in all of geoprocess…

### LINKED ISSUE #59: calculate slope unnecessarily sets fills the nodata value
It is not necessary to fill a nodata value in an empty raster, but `calculate_slope` does it. This is noticeable in large rasters when filling can take up to 30% of the total runtime of the function. Easy fix: don't fill.

richpsharp:
This PR removes an inefficiency I spotted in `calculate_slope` where it first fills the target raster with nodata values. This is not necessary with any recent version of GDAL (and I wonder if it ever was?) and can otherwise be a significant step in calculation when calculating slopes, sometimes almost doubling computation time.

This PR removes the `fill_value=...` call in `calculate_slope` and a few irrelvant `fill_value=None`s that aren't useful either in `geoprocessing_core.py`. There's no need to add an additional test since the current slope testing suite includes nodata values in the base data and still behave as expected.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -15,6 +15,8 @@ Unreleased Changes
   might be included with the source distribution.
 * Added ``set_tol_to_zero`` to ``convolve_2d`` to allow for in-function masking
   of near-zero results to be set to 0.0.
+* Fixes an uncessary calculation that pre-fills slope raster GeoTIFFs with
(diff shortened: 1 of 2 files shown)


dcdenu4 (approved):
This is a good observation and catch.

PR #70, opened 2020-05-06
fixes #48 adding a test to cover a feature with no geometry defined, …

### LINKED ISSUE #48: Should zonal_statistics handle features without geometries?
In `pygeoprocessing.calculate_disjoint_polygon_set`, we load in all features of a vector in order to build an `rtree`-based spatial index.  In order to do this, we loop through all the features in the input vector, exporting each one to WKB before importing the feature as a `shapely` geometry.

When a vector feature does not have a geometry, it `GetGeometryRef()` returns `None`, which will cause an exception to be raised.

At a minimum, this contractual requirement should be documented in the docstring, but we may also want to just handle this properly, perhaps by logging a warning.

Related to https://github.com/natcap/invest/issues/64

This came up on the forums recently: https://community.naturalcapitalproject.org/t/carbon-model-error/883/3?u=jdouglass

richpsharp:
Fixing issue #48 by adding tests for features with no geometry defined, printing a warning to the log if so, and adding test coverage for this case in the test suite.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -15,6 +15,9 @@ Unreleased Changes
   might be included with the source distribution.
 * Added ``set_tol_to_zero`` to ``convolve_2d`` to allow for in-function masking
   of near-zero results to be set to 0.0.
+* Fixed an issue in ``zonal_statistics`` that would crash if an aggregate
+  vector had a feature with no geometry defined. Now the function ignores
+  such features and prints a warning to the log.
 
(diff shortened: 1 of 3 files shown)


phargogh (approved):
Thanks for patching this, @richpsharp !

phargogh:
@richpsharp looks great, there's just a merge conflict on `HISTORY` from another merged PR.

PR #76, opened 2020-05-06
closes #23 adding a check for well formed raster/path band tuples, ra…

### LINKED ISSUE #23: convolve_2d doesn't warn if raster/path band not passed
`pygeoprocessing.convolve_2d` doesn't provide a useful error message if a raw path is passed for one of the raster path band tuples. Many other pygeoprocessing functions do provide this functionality so it should probably be added here too.

richpsharp:
Adds a check for the well formed paths, raise an exception if they aren't, add a test to cover it, and note in history.

### DIFF
--- a/tests/test_geoprocessing.py
+++ b/tests/test_geoprocessing.py
@@ -2589,6 +2589,22 @@ class PyGeoprocessing10(unittest.TestCase):
             n_pixels ** 2 * 9 - n_pixels * 4 * 3 + 4)
         numpy.testing.assert_allclose(numpy.sum(target_array), expected_result)
 
+    def test_convolve_2d_bad_path_bands(self):
+        """PGP.geoprocessing: test convolve 2d bad raster path bands."""
+        import pygeoprocessing
+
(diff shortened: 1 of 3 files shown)


tests/test_geoprocessing.py
```diff
@@ -2589,6 +2589,22 @@ def test_convolve_2d_single_thread(self):
             n_pixels ** 2 * 9 - n_pixels * 4 * 3 + 4)
         numpy.testing.assert_allclose(numpy.sum(target_array), expected_result)
 
+    def test_convolve_2d_bad_path_bands(self):
+        """PGP.geoprocessing: test convolve 2d bad raster path bands."""
+        import pygeoprocessing
```
dcdenu4, line 2594: Just flagging in case you are removing these imports in another PR or branch, then this new one might not be caught.

dcdenu4 (approved):
Hey @richpsharp this looks good. The only thing I flagged was the importing of `pygeoprocessing` for the test. If you're actively working on removing all these in another branch / PR this one might slip through someday.

PR #79, opened 2020-05-07
Bugfix/routing 46 memory error

### LINKED ISSUE #46: Fix memory error in routing
James left me some sample data that's helpful. More soon, but I need an issue # to start working.

richpsharp:
Hi @phargogh  and @dcdenu4, requesting both of your reviews because this addresses the same bug you both found in different ways. @phargogh, this fixes Chris's weird DEM route and @dcdenu4, this fixes that user's forum question where there was an out of memory error during routing. 

The bug was in `flow_dir_mfd` in the section where the algorithm walks through a plateau to determine where it can drain. It does this by comparing its current height with its neighbors and seeing if the neighbor has a shorter route. Except i was comparing heights with an `is_close` rather than `==` as we have been used to for comparing nodata values. Except in this case both of these users' data had VERY small deltas between some of the dem heights, such as 12.2347328328 and 12.2347329383. `is_close` returned true for both of these values. So now I ensure both values from the DEM are cast to 64 bit floating point values then compared. If there's a numerical roundoff there it will be the same numerical roundoff and results will be correct when checking for equality.

Also during testing I was able to catch and re-create a flaky bug where a ManagedRaster will flush the cache but sometimes be unable to open the raster to do so. This can occur as a race condition of hopping around a big raster and GDAL hasn't necessarily cleared out its cache for the raster yet. I explicitly test for this now and wait 0.2 seconds up to 5 times to retry. 

I've included an okayish test that I feel covers this, though it is not large enough to cause a memory error like the user @dcdenu4 worked with. I have, however, directly routed that user's data so I can informally verify it works in this PR.

closes #46


richpsharp:
something's wrong

PR #80, opened 2020-05-07
fixes #46, adding a fix to is_close vs == for root height, an issue w…

### LINKED ISSUE #46: Fix memory error in routing

richpsharp:
Hi @phargogh and @dcdenu4, requesting both of your reviews because this addresses the same bug you both found in different ways. @phargogh, this fixes Chris's weird DEM route and @dcdenu4, this fixes that user's forum question where there was an out of memory error during routing.

The bug was in flow_dir_mfd in the section where the algorithm walks through a plateau to determine where it can drain. It does this by comparing its current height with its neighbors and seeing if the neighbor has a shorter route. Except i was comparing heights with an is_close rather than == as we have been used to for comparing nodata values. Except in this case both of these users' data had VERY small deltas between some of the dem heights, such as 12.2347328328 and 12.2347329383. is_close returned true for both of these values. So now I ensure both values from the DEM are cast to 64 bit floating point values then compared. If there's a numerical roundoff there it will be the same numerical roundoff and results will be correct when checking for equality.

Also during testing I was able to catch and re-create a flaky bug where a ManagedRaster will flush the cache but sometimes be unable to open the raster to do so. This can occur as a race condition of hopping around a big raster and GDAL hasn't necessarily cleared out its cache for the raster yet. I explicitly test for this now and wait 0.2 seconds up to 5 times to retry.

I've included an okayish test that I feel covers this, though it is not large enough to cause a memory error like the user @dcdenu4 worked with. I have, however, directly routed that user's data so I can informally verify it works in this PR.

closes #46

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -17,10 +17,12 @@ Unreleased Changes
   of near-zero results to be set to 0.0.
 * Fixed a malformed ``ValueError`` message when a corrupt raster was
   encountered in ``raster_calculator``.
-* Fixes an uncessary calculation that pre-fills slope raster GeoTIFFs with
+* Fixes an unnecessary calculation that pre-fills slope raster GeoTIFFs with
   nodata values.
 * Added a check to ``convolve_2d`` to verify that raster path/band tuples were
-  passed where expected and raise a useful Exception  if not.
+  passed where expected and raise a useful Exception if not.
+* Fixed an issue in ``flow_dir_mfd`` that would cause invalid flow directions
+  on DEMs that had very small numerical delta heights.
 
 1.9.2 (2020-02-06)
 ------------------
--- a/src/pygeoprocessing/routing/routing.pyx
+++ b/src/pygeoprocessing/routing/routing.pyx
@@ -479,6 +479,11 @@ cdef class _ManagedRaster:
                     time.sleep(0.2)
                     continue
                 break
+            if max_retries == 0:
+                raise ValueError(
+                    f'unable to open {self.raster_path} in '
+                    'ManagedRaster.flush')
+            raster_band = raster.GetRasterBand(self.band_id)
 
         block_array = numpy.empty(
             (self.block_ysize, self.block_xsize), dtype=numpy.double)
--- a/src/pygeoprocessing/routing/routing.pyx
+++ b/src/pygeoprocessing/routing/routing.pyx
@@ -468,9 +468,17 @@ cdef class _ManagedRaster:
 
         raster_band = None
         if self.write_mode:
-            raster = gdal.OpenEx(
-                self.raster_path, gdal.GA_Update | gdal.OF_RASTER)
-            raster_band = raster.GetRasterBand(self.band_id)
+            max_retries = 5
+            while max_retries > 0:
+                raster = gdal.OpenEx(
+                    self.raster_path, gdal.GA_Update | gdal.OF_RASTER)
+                if raster is None:
+                    max_retries -= 1
+                    LOGGER.error(
+                        f'unable to open {self.raster_path}, retrying...')
+                    time.sleep(0.2)
+                    continue
+                break
 
         block_array = numpy.empty(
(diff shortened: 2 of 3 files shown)


HISTORY.rst
```diff
@@ -17,10 +17,12 @@ Unreleased Changes
   of near-zero results to be set to 0.0.
 * Fixed a malformed ``ValueError`` message when a corrupt raster was
   encountered in ``raster_calculator``.
-* Fixes an uncessary calculation that pre-fills slope raster GeoTIFFs with
+* Fixes an unnecessary calculation that pre-fills slope raster GeoTIFFs with
   nodata values.
 * Added a check to ``convolve_2d`` to verify that raster path/band tuples were
-  passed where expected and raise a useful Exception  if not.
```
dcdenu4, line 23: Ha! I saw this in a previous PR, but didn't have the heart to point out the double space then, haha.

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -468,9 +468,17 @@ cdef class _ManagedRaster:
...
+            while max_retries > 0:
+                raster = gdal.OpenEx(
+                    self.raster_path, gdal.GA_Update | gdal.OF_RASTER)
+                if raster is None:
+                    max_retries -= 1
+                    LOGGER.error(
+                        f'unable to open {self.raster_path}, retrying...')
+                    time.sleep(0.2)
+                    continue
+                break
```
dcdenu4, line 481: I'm not up to speed on this set of code, but if raster is NOT `None`, do we need to get the `raster_band`, and if it times out without returning `None`, should there be an error or is it find to just it continue? Perhaps I'll answer my own questions here as I read through.

tests/test_routing.py
```diff
@@ -1147,3 +1147,52 @@ def test_distance_to_channel_mfd_with_weights(self):
 
         numpy.testing.assert_almost_equal(
             distance_to_channel_d8_array, zero_array)
+
+    def test_flow_dir_mfd_plateau(self):
+        """PGP.routing: MFD on a plateau."""
+        import pygeoprocessing.routing
+
+        driver = gdal.GetDriverByName('GTiff')
+        workspace_dir = 'flow_dir_debug'
```
dcdenu4: Should this be `workspace_dir = os.path.join(self.workspace_dir, 'flow_dir_debug') ` or something?

dcdenu4 (changes requested):
Hey @richpsharp Just had a few clarifying questions and small suggestions!

dcdenu4:
Also, the PR title!
> fixes $46, adding .... 

Dolla, dolla, bills ya'll!

HISTORY.rst
richpsharp: oh you definitely should, if you see an error and let it slide? now we're both at fault...

richpsharp: We can discuss if it should be fixed later, but please don't ignore stuff like that even if it seems trivial.

src/pygeoprocessing/routing/routing.pyx
richpsharp: oh woof thanks for catching that.. that'd be a nightmare if that got out. patched in 1fe4d3f5dfbf90db2529ee4693a43c7a817e8e5b

this wouldn't show up unless the raster were very big, so it's hard to make test cases of this size since we don't want to wait 15 minutes for  test to run. stuff like this though makes me wary.

tests/test_routing.py
richpsharp: Even better, it should be removed. I do that a lot when I'm doing a single test and want to inspect the results, then i unintentionally leave it there. removed in 0171f2d5c540901f5f144bc37deca44957244f13

richpsharp:
hi @dcdenu4, I addressed your comments and re-requested your review

dcdenu4 (approved):
Thanks for the changes!

PR #84, opened 2020-05-08
Feature/testing 43 remove testing module

### LINKED ISSUE #43: Remove `pygeoprocessing.testing`

richpsharp:
This PR removes the testing module from pygeoprocessing while retaining and expanding the functionality that was useful in pygeoprocessing and InVEST's test suites. Key features:

Three convenience functions are added to `pygeoprocessing` which should not be used for any production code, but are fine and encouraged to be used in testing and scripting:
  * `raster_to_numpy_array`
  * `numpy_array_to_raster`
  * `shapely_geometry_to_vector`

These functions have been deployed in pygeoprocessing's test suite everywhere that it makes the codebase cleaner.

These functions can be used in InVEST's testing suite where it shared some pygeoprocessing testing functionality. In the case of comparing "raster equality" the implementer (probably me) should use a combination of `numpy.isclose...all()` and `raster_to_numpy_array`.

closes #84


richpsharp:
hi @phargogh, you're gonna need a beer or 3 for this review. First of all I know you love it when PRs make the code base shorter, so notice the 4000+ lines deleted vs 800+ added. A lot of that is removing the testing module, but another bit of it is just appropriate use of an API. I thought a lot about orthogonality and overdesign when doing this and I also thought a lot about how this would integrate with InVEST. So I'm sure there will be a lot here, but I hope you can keep those in mind too while you do this PR.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -3550,3 +3540,133 @@ def _is_number(x):
...
+        shapely_geometry_list (list): a list of Shapely objects.
+        target_vector_path (str): path to target vector.
+        projection_wkt (str): WKT for target vector.
+        vector_format (str): GDAL driver name for target vector.
+        fields (dict): a python dictionary mapping string fieldname
+            to OGR Fieldtypes, if None no fields are added
+        attribute_list (list of dicts): a list of python dictionary mapping
+            fieldname to field value for each geometry in
+            `shapely_geometry_list`, if None, no attributes are created.
+        ogr_geom_type (ogr geometry enumerated type): defaults to wkbPolygon.
```
phargogh: From the docstring alone, this feels nonorthogonal since the geometry type can be inferred from the Shapely objects provided in `shapely_geometry_list`.  I see from the implementation, though, that this is really intended to be the geometry type of the layer.

Do we have a use case where the layer type needs to be different from what could be inferred from the geometry type of the Shapely objects?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -3550,3 +3540,133 @@ def _is_number(x):
...
+
+    target_layer = None
+    target_vector = None
+
+
+def numpy_array_to_raster(
+        base_array, target_nodata, pixel_size, origin, projection_wkt,
+        target_path,
+        raster_driver_creation_tuple=DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS):
+    """Create a single band raster of size `base_array.shape` at `target_path`.
```
phargogh: I think double-backticks are intended here since docstrings are interpreted as RST.

tests/test_geoprocessing.py
```diff
@@ -516,8 +483,13 @@ def test_reproject_vector_latlon_to_latlon(self):
...
         # Since projecting to the same SRS, the vectors should be identical
-        pygeoprocessing.testing.assert_vectors_equal(
-            base_vector_path, target_vector_path, 1e-3)
+        target_vector = gdal.OpenEx(target_vector_path, gdal.OF_VECTOR)
+        target_layer = target_vector.GetLayer()
+        self.assertEqual(target_layer.GetFeatureCount(), 1)
+        feature = next(iter(target_layer))
+        feature_geom = shapely.wkt.loads(
+            feature.GetGeometryRef().ExportToWkt())
+        self.assertTrue(feature_geom.almost_equals(polygon_a))
```
phargogh, line 492: I did not know that shapely had methods like this!

tests/test_geoprocessing.py
```diff
@@ -2712,79 +2415,45 @@ def test_iterblocks_unsigned_byte(self):
...
-        kernel_path = os.path.join(self.workspace_dir, 'kernel.tif')
-        target_path = os.path.join(self.workspace_dir, 'target.tif')
-
-        with self.assertRaises(ValueError) as cm:
-            pygeoprocessing.convolve_2d(
-                signal_path, kernel_path, target_path)
-        actual_message = str(cm.exception)
-        # we expect an error about both signal and kernel
-        self.assertTrue('signal' in actual_message)
-        self.assertTrue('kernel' in actual_message)
```
phargogh, line 2760: Is this test deletion intentional?  I'm only raising this because it seems inconsistent with the other edits in the PR, so just wanted to make sure it wasn't an accident and you really wanted to keep the test!

tests/test_geoprocessing.py
```diff
@@ -3986,107 +3482,55 @@ def test_merge_bounding_box_list(self):
...
-        poly_b.AddGeometry(ring)
+        poly_a = shapely.geometry.Polygon([
+            (_DEFAULT_ORIGIN[0], _DEFAULT_ORIGIN[1]),
+            (_DEFAULT_ORIGIN[0] +
+             _DEFAULT_PIXEL_SIZE[0], _DEFAULT_ORIGIN[1]),
+            (_DEFAULT_ORIGIN[0] + _DEFAULT_PIXEL_SIZE[0],
+             _DEFAULT_ORIGIN[1] + _DEFAULT_PIXEL_SIZE[1]),
+            (_DEFAULT_ORIGIN[0],
+             _DEFAULT_ORIGIN[1] + _DEFAULT_PIXEL_SIZE[1]),
+            (_DEFAULT_ORIGIN[0], _DEFAULT_ORIGIN[1])])
```
phargogh: I think this is creating a box-shaped polygon, right?

If it is, we could consider using the nicely abbreiviated `shapely.geometry.box` (https://shapely.readthedocs.io/en/latest/manual.html#shapely.geometry.box) for convenience and brevity.  Just a thought!

tests/test_geoprocessing.py
```diff
@@ -4756,41 +4121,26 @@ def test_iterblocks_bad_raster(self):
...
-        new_band.SetNoDataValue(nodata_target)
-        new_band.WriteArray(pixel_a_matrix)
-        new_raster.FlushCache()
-        new_band = None
-        new_raster = None
 
         target_raster_path = os.path.join(self.workspace_dir, 'target_a.tif')
         base_a_raster_info = pygeoprocessing.get_raster_info(base_a_path)
-
+        print(base_a_raster_info)
```
phargogh: Stray `print()`?

tests/test_routing.py
```diff
@@ -1,1176 +1,1107 @@
...
+        dem_array[0, 0] = -1.0
+        raster = driver.Create(
+            base_path, dem_array.shape[1], dem_array.shape[0], 2,
+            gdal.GDT_Float32)
+        band = raster.GetRasterBand(1)
+        band.WriteArray(dem_array)
+        band.FlushCache()
+        band = None
+        raster = None
+        fill_path = os.path.join(self.workspace_dir, 'filled.tif')
```
phargogh: Wouldn't this also be able to be subsumed by the new raster creation function?

tests/test_routing.py
```diff
@@ -1,1176 +1,1107 @@
...
+            (base_path, 1), fill_path, working_dir=self.workspace_dir)
+
+        result_raster = gdal.OpenEx(fill_path, gdal.OF_RASTER)
+        raster_count = result_raster.RasterCount
+        result_band = result_raster.GetRasterBand(1)
+        result_array = result_band.ReadAsArray()
+        result_band = None
+        result_raster = None
+        self.assertEqual(raster_count, 1, "expect 1 band output")
+        self.assertEqual(result_array.dtype, numpy.float32)
```
phargogh: Wouldn't this also be able to be converted to the new raster --> numpy function?

tests/test_routing.py
```diff
@@ -1,1176 +1,1107 @@
...
+        dem_array[1, 1] = nodata
+        raster = driver.Create(
+            base_path, dem_array.shape[1], dem_array.shape[0], 1,
+            gdal.GDT_Int32)
+        band = raster.GetRasterBand(1)
+        band.WriteArray(dem_array)
+        band.FlushCache()
+        band = None
+        raster = None
+        fill_path = os.path.join(self.workspace_dir, 'filled.tif')
```
phargogh: Same here for raster creation?

tests/test_routing.py
```diff
@@ -1,1176 +1,1107 @@
...
+        dem_array = numpy.zeros((11, 11))
+        dem_raster = driver.Create(
+            dem_path, dem_array.shape[1], dem_array.shape[0], 1,
+            gdal.GDT_Float32, options=(
+                'TILED=NO', 'BIGTIFF=YES', 'COMPRESS=LZW',
+                'BLOCKXSIZE=11', 'BLOCKYSIZE=1'))
+
+        dem_band = dem_raster.GetRasterBand(1)
+        dem_band.WriteArray(dem_array)
+        dem_band.FlushCache()
```
phargogh: Same here for raster creation?

tests/test_routing.py
```diff
@@ -1,1176 +1,1107 @@
...
-
-        flow_dir_raster = gdal.OpenEx(target_flow_dir_path, gdal.OF_RASTER)
-        flow_dir_array = flow_dir_raster.ReadAsArray()
-        flow_dir_raster = None
-        flow_dir_nodata = pygeoprocessing.get_raster_info(
-            target_flow_dir_path)['nodata'][0]
-        self.assertTrue(not numpy.isclose(
-            flow_dir_array[1:-1, 1: -1], flow_dir_nodata).any(),
-            'all flow directions should be defined')
+"""pygeoprocessing.routing test suite."""
```
phargogh, line 1: Sorry, I think the newlines got switched around, so it's tougher to see what has changed.

But in starting to go through the functions, there were quite a few places where GDAL's `Driver.Create` was called, which made me wonder about whether these could be abbreviated using the new raster-from-array functionality.

phargogh (changes requested):
This looks really exciting @richpsharp !  I had a few comments for you, and the biggest unknown to me was around `tests/test_routing.py` and whether a bunch of `driver.Create`s should be converted to use the new array-to-raster functionality.

Anyway, let me know what you think!

src/pygeoprocessing/geoprocessing.py
richpsharp: I think so but it is not a common case, it's usually when something's broken. Occasionally the geometry type of a layer will not actually be the geometry type of the features, sometimes subtle because of multipolygon in a polyon or such. pygeoprocessing does have the useful concept about exposing the underlying switches if you need them a la' `raster_driver_creation_tuple`.  We often hit those when something's broken and while not as extreme `ogr_geom_type` feels like one of those switches to me. I do agree that docstring is a little light, I explain in b8e4d7dff20fa1629f188b9f2a41b53393f2574a that it's used for setting the target layer's geometry type.

src/pygeoprocessing/geoprocessing.py
richpsharp: Added in 7e3c8c32d458262adec6521fea6f586af84476fb but also removed the "target_path" so it'd all fit in one line.

src/pygeoprocessing/geoprocessing.py
richpsharp: Aww crud my comment didn't stick. I've had cases where the underlying polygon types do not correspond to the layer and not have control over it. Say it's a multipolygon but really a polygon? To me this argument feels like the `raster_creation_tuple` option on the other functions in pgp -- it's a switch to something you don't often need but if it's not there when you do need it you have to write your own implementation. I added a better docstring in b8e4d7dff20fa1629f188b9f2a41b53393f2574a to indicate that parameter is for the target layer geometry type

src/pygeoprocessing/geoprocessing.py
richpsharp: Fixed in 7e3c8c32d458262adec6521fea6f586af84476fb and made the line shorter so it fits on one line still

tests/test_geoprocessing.py
richpsharp: Gosh I don't think it was intentional! I put it back in 99a689400b1b1722f22a3de4077e993c6c37cb90

tests/test_geoprocessing.py
richpsharp: Good idea, patched in b95d82185def71581001da45aac75f7ac7f13e2b

tests/test_geoprocessing.py
richpsharp: it was! patched in 275ba4e785c7cae65dc82e336d0cd2e72617455e

tests/test_routing.py
richpsharp: oh boy that was a doozy, looks like I skipped that entire module for raster creation. I've patched it heavily in 364b9f36856005f681f2848e80303664ceb23dc9

tests/test_routing.py
richpsharp: patched in 364b9f36856005f681f2848e80303664ceb23dc9

tests/test_routing.py
richpsharp: patched in 364b9f36856005f681f2848e80303664ceb23dc9

tests/test_routing.py
richpsharp: patched in 364b9f36856005f681f2848e80303664ceb23dc9

tests/test_routing.py
richpsharp: I did them all in 364b9f36856005f681f2848e80303664ceb23dc9

richpsharp:
I addressed all your requested changes @phargogh !

src/pygeoprocessing/geoprocessing.py
phargogh: The comment stuck!  And thanks for the updated docstring.

phargogh (approved):
Fare thee well, `pygeoprocessing.testing`!

PR #91, opened 2020-05-19
Bugfix/geoprocessing 87 nodata holes in convolve 2d

### LINKED ISSUE #87: convolve 2d gets nodata holes
Seems to happen with lots of nodata in the signal raster, here's the data: https://drive.google.com/drive/u/1/folders/1841DI6meUxa9mR2DoXJ0NKJV8nsIwTXL

richpsharp:
This PR fixes an issue where NaN holes would accumulate when `convolve_2d` was invoked with `ignore_nodata_and_edges=True` but `mask_nodata=False`. Rather than set the non-maksed nodata values to 0 they were treated as an empty value. This PR fixes that and has a test to cover that case.

Closes #87


dcdenu4 (approved):
Thanks Rich!

PR #96, opened 2020-05-19
fixing indention error in history and adding the description content …

richpsharp:
fixing history syntax error so we can release

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,7 @@ setup(
     name='pygeoprocessing',
     description="PyGeoprocessing: Geoprocessing routines for GIS",
     long_description=LONG_DESCRIPTION,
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -26,6 +26,7 @@
     name='pygeoprocessing',
     description="PyGeoprocessing: Geoprocessing routines for GIS",
     long_description=LONG_DESCRIPTION,
+    long_description_content_type='text/x-rst',
```
phargogh, line 29: Nice.  I'll add this to InVEST.

phargogh (commented):
Looks good to me!  Just waiting for tests to finish as a final confirmation.

phargogh (approved):
Tests pass.  Merging!

PR #101, opened 2020-06-05
Explicitly pulling tags as well.  RE:#45

### LINKED ISSUE #45: Tags aren't pulled when building wheels
When updating to `actions/checkout@v2`in #36, I missed the fact that tags are no longer pulled by default.  This isn't a functional problem when running tests, but it *is* a problem when building wheels, since the version string is part of the filename.

phargogh:
This fixes #45 by simply adding a step to pull tags when we build the python distributions.

This works: take a look at [this build](https://github.com/natcap/pygeoprocessing/pull/101/checks?check_run_id=743986004), and the wheel `dist/pygeoprocessing-2.0.0.post9+g6e6560b-cp38-cp38-macosx_10_14_x86_64.whl` is constructed, which has the correct version string in the filename.

### DIFF
--- a/.github/workflows/build-py-dists.yml
+++ b/.github/workflows/build-py-dists.yml
@@ -27,6 +27,9 @@ jobs:
                   # Fetch all history so that setuptools_scm can build the correct version string.
                   fetch-depth: 0
 
+            - name: Fetch git tags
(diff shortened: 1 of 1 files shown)


richpsharp (approved):
explicit!

PR #106, opened 2020-08-06
Adding custom exception for reclassification error

### LINKED ISSUE #105: Proposal: allow missing values to be accessed as attributes of a reclassification exception
In InVEST, we are looking to create more readable error messages when a table-based reclassification fails, and the way that we think these messages will be most useful will be if we can provide more contextual information about the error. The error that is raised by `pygeoprocessing.reclassify_raster`, while sensible for us as programmers who already understand the context of the exception, does not allow for the useful information to be accessed without some string parsing.

I propose that a custom exception be raised when values are missing from the `value_map`.  The purpose of this exception is to:

1. Allow the exception relating to missing values to be caught and have that error handled apart from other errors that might be raised by `reclassify_raster`
2. Allow the relevant attributes of the reclassification to be accessed as attributes of the new exception such that they can be used in a higher-level exception that provides additional context that is specific to the InVEST model.  At present this is possible, but would require that we extract this information from the string text of the `ValueError` currently raised.  The relevant attributes here would be:
    * The set of `missing_values`
    * The `value_map`  (though I'm less worried about this, because this structure is still accessible at the point where InVEST calls `reclassify_raster`).
3. Continue to be a useful exception when called directly within a script by a developer, with a useful error message.

Example of how this could be used in InVEST:

```python
try:
    pygeoprocessing.reclassify_raster(...)
except pygeoprocessing.ReclassificationError as error:
    error_message = f'The biophysical table is missing rows with LULC code {error.missing_values}'
    raise ValueError(error_message)
```
    
Including @dcdenu4 since we had talked about this in the context of reclassifications across InVEST.

Also, this is related to #103.

What do you think about this @richpsharp ?

dcdenu4:
Adding a custom exception to `pygeoprocessing/geoprocessing.py` called `ReclassificationMissingValuesError` that takes a _message_, _missing values_, and _value map_. These are set as attributes that can then be accessed via InVEST for example.

Added the importing of this custom class to `__init__.py` so it is available via `pygeoprocessing.ReclassificationMissingValuesError`.

Closes #105

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,15 @@ Release History
 
 Unreleased Changes
 ------------------
+* Adding custom exception ``ReclassificationMissingValuesError`` that can 
+  occur when mapping raster values to dictionary keys, where a raster value 
+  is not represented in the dictionary as a key. The custom exception takes 
+  3 arguments that are set as attributes:
+  * ``msg`` - a string for the error message
+  * ``missing_values`` - a list of raster values not found in the dictionary
+  * ``value_map`` - the dictionary the raster values were attempted to map to
+  This error is now raised in ``reclassify_raster`` and can be caught 
+  to better debug the mismatch when using a table to reclassify a raster. 
 * Correcting the docstring for ``pygeoprocessing.numpy_array_to_raster`` to
   specify that the ``pixel_size`` parameter must be a tuple or list, not an
   int.
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,15 +3,13 @@ Release History
 
 Unreleased Changes
 ------------------
-* Adding custom exception ``ReclassificationMissingValuesError`` that can 
-  occur when mapping raster values to dictionary keys, where a raster value 
-  is not represented in the dictionary as a key. The custom exception takes 
-  3 arguments that are set as attributes:
-  * ``msg`` - a string for the error message
-  * ``missing_values`` - a list of raster values not found in the dictionary
-  * ``value_map`` - the dictionary the raster values were attempted to map to
-  This error is now raised in ``reclassify_raster`` and can be caught 
-  to better debug the mismatch when using a table to reclassify a raster. 
+* Added a custom exception class ``ReclassificationMissingValuesError`` to 
+  ``pygeoprocessing``. ``pygeoprocessing.reclassify_raster`` raises this
+  exception instead of ``ValueError`` when a raster pixel value is not
+  represented in ``value_map``. This custom exception provides a list of 
+  missing raster pixel values in a ``missing_values`` attribute that allows
+  the caller access to the pixel values that are missing through a Python type
+  rather than indirectly through an error message.
 * Correcting the docstring for ``pygeoprocessing.numpy_array_to_raster`` to
   specify that the ``pixel_size`` parameter must be a tuple or list, not an
   int.
--- a/src/pygeoprocessing/__init__.py
+++ b/src/pygeoprocessing/__init__.py
@@ -13,7 +13,8 @@ from .geoprocessing_core import raster_band_percentile
 import pkg_resources
 __version__ = pkg_resources.get_distribution(__name__).version
 
-__all__ = ('calculate_slope', 'raster_band_percentile')
+__all__ = ('calculate_slope', 'raster_band_percentile', 
+           'ReclassificationMissingValuesError')
 for attrname in dir(geoprocessing):
     attribute = getattr(geoprocessing, attrname)
     if isinstance(attribute, types.FunctionType):
--- a/src/pygeoprocessing/__init__.py
+++ b/src/pygeoprocessing/__init__.py
@@ -7,6 +7,7 @@ import sys
 import types
 
 from . import geoprocessing
+from .geoprocessing import ReclassificationMissingValuesError
 from .geoprocessing_core import calculate_slope
 from .geoprocessing_core import raster_band_percentile
 import pkg_resources
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -36,15 +36,12 @@ class ReclassificationMissingValuesError(Exception):
         msg (str) - error message
         missing_values (list) - a list of the missing values from the raster
             that are not keys in the dictionary
-        value_map (dict) - the dictionary used to map raster values
-
     """
 
-    def __init__(self, msg, missing_values, value_map):
+    def __init__(self, msg, missing_values):
         self.msg = msg
         self.missing_values = missing_values
-        self.value_map = value_map
-        super().__init__(msg, missing_values, value_map)
+        super().__init__(msg, missing_values)
 
 LOGGER = logging.getLogger(__name__)
 
@@ -1737,7 +1734,7 @@ def reclassify_raster(
     Raises:
         ReclassificationMissingValuesError if ``values_required`` is ``True``
         and a pixel value from ``base_raster_path_band`` is not a key in
-        ``attr_dict``.
+        ``value_map``.
 
     """
     if len(value_map) == 0:
@@ -1767,7 +1764,7 @@ def reclassify_raster(
                     f'The following {missing_values.size} raster values'
                     f' {missing_values} from "{base_raster_path_band[0]}"'
                     ' do not have corresponding entries in the ``value_map``:'
-                    f' {value_map}.', missing_values, value_map)
+                    f' {value_map}.', missing_values)
         index = numpy.digitize(original_values.ravel(), keys, right=True)
         return values[index].reshape(original_values.shape)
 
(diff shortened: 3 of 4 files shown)


tests/test_geoprocessing.py
```diff
@@ -113,13 +116,11 @@ def test_reclassify_raster_missing_pixel_value(self):
             test_value: 100,
         }
         target_nodata = -1
-        with self.assertRaises(ValueError) as cm:
+        with self.assertRaises(reclass_error) as cm:
```
richpsharp: In combination with line 102 above, I'd tend to do an `from .geoprocessing import ReclassificationMissingValuesError` then `self.assertRaises(ReclassificationMissingValuesError)` rather than obscure the exception through a variable. Or just `pygeoprocessing.ReclassificationMissingValuesError`. It's one of the reasons I kinda like custom exceptions when they make sense -- you can infer their meaning from their class name.

tests/test_geoprocessing.py
```diff
@@ -113,13 +116,11 @@ def test_reclassify_raster_missing_pixel_value(self):
...
         target_nodata = -1
-        with self.assertRaises(ValueError) as cm:
+        with self.assertRaises(reclass_error) as cm:
             pygeoprocessing.reclassify_raster(
                 (raster_path, 1), value_map, target_path, gdal.GDT_Float32,
                 target_nodata, values_required=True)
-        expected_message = (
-            'The following 1 raster values [-0.5] from "%s" do not have ' %
-            (raster_path,))
+        expected_message = ('The following 1 raster values [-0.5]')
```
richpsharp: I think you can lose the parens here now that it's on one line.

richpsharp (changes requested):
Easy Doug! I had a couple stylistic comments, but also can you add an entry in `HISTORY.rst` since this is a user-facing change? I think a brief statement of guidance of the expected use of that exception would be helpful there too.

tests/test_geoprocessing.py
dcdenu4: Yep, agreed. This was me trying to appease the `line too long` linting gnomes

tests/test_geoprocessing.py
dcdenu4: Parens removed in https://github.com/natcap/pygeoprocessing/pull/106/commits/8509356492a7b45beaa058e8184e09d1087603a7

tests/test_geoprocessing.py
dcdenu4: Now using `pygeoprocessing.ReclassificationMissingValuesError` straight up in https://github.com/natcap/pygeoprocessing/pull/106/commits/8509356492a7b45beaa058e8184e09d1087603a7

src/pygeoprocessing/__init__.py
```diff
@@ -7,6 +7,7 @@
 import types
 
 from . import geoprocessing
+from .geoprocessing import ReclassificationMissingValuesError
```
phargogh, line 10: Could you add `ReclassificationMissingValuesError`  to `__all__` as well here?  Doing so will enable `from pygeoprocessing import *` to also expose `ReclassificationMissingValueError`.

phargogh (changes requested):
Looks like a good solution to me!  I had one comment about including this in `__all__`, and then since this is a user-facing change, could you include a note in `HISTORY` about this?  Thanks!

richpsharp (changes requested):
Thanks for making those changes Doug. Did you see my note about `HISTORY.rst`? Can you update that to show the user facing changes and a description about how this exception could occur and why a user may want to catch this exception?

src/pygeoprocessing/__init__.py
dcdenu4: Added to `__all__` here: https://github.com/natcap/pygeoprocessing/pull/106/commits/5eecc74c5c71675b465c97261efedacbe8ce3f8e

dcdenu4:
@richpsharp @phargogh I took a stab at a useful History blurb here: https://github.com/natcap/pygeoprocessing/pull/106/commits/da813ea879abf03582c072a77c29b1946dcca299

src/pygeoprocessing/geoprocessing.py
```diff
@@ -29,6 +29,23 @@
...
 import shapely.wkb
 
+class ReclassificationMissingValuesError(Exception):
+    """Raised when a raster value is not a valid key to a dictionary.
+
+    Attributes:
+        msg (str) - error message
+        missing_values (list) - a list of the missing values from the raster
+            that are not keys in the dictionary
+        value_map (dict) - the dictionary used to map raster values
```
richpsharp: Hey Doug, I'm sorry I didn't notice this before until you made a note of it in your HISTORY.rst. Presumably the caller is in the same context as the exception that's raised here so the `value_map` should be in scope of handling that exception right? The exception should be providing information the caller does not have, which are the actual missing values. It's not necessarily bad to include this, but naively it seems unnecessary and just another bit of code that could be dropped. Is there a reason you're including `value_map` as an attribute in the exception?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1718,8 +1735,9 @@ def reclassify_raster(
         None
 
     Raises:
-        ValueError if ``values_required`` is ``True`` and a pixel value from
-           ``base_raster_path_band`` is not a key in ``attr_dict``.
+        ReclassificationMissingValuesError if ``values_required`` is ``True``
+        and a pixel value from ``base_raster_path_band`` is not a key in
+        ``attr_dict``.
```
richpsharp: Sorry, I didn't notice tihs earlier either, but I think ``attr_dict`` should be ``value_map``. I'm guessing that variable changed a while ago and we didn't change the docstring.

HISTORY.rst
```diff
@@ -3,6 +3,15 @@ Release History
...
+* Adding custom exception ``ReclassificationMissingValuesError`` that can 
+  occur when mapping raster values to dictionary keys, where a raster value 
+  is not represented in the dictionary as a key. The custom exception takes 
+  3 arguments that are set as attributes:
+  * ``msg`` - a string for the error message
+  * ``missing_values`` - a list of raster values not found in the dictionary
+  * ``value_map`` - the dictionary the raster values were attempted to map to
+  This error is now raised in ``reclassify_raster`` and can be caught 
+  to better debug the mismatch when using a table to reclassify a raster. 
 * Correcting the docstring for ``pygeoprocessing.numpy_array_to_raster`` to
```
richpsharp: I've got a proposed re-word of this section below. The ``msg`` is part of the base exception so I don't think that needs mentioning and if you're up for my suggestion below I think you can drop ``value_map``. 

`Added a custom exception ``ReclassificationMissingValuesError`` to ``pygeoprocessing.reclassify_raster`` that raises when a raster pixel value is not represented in ``value_map``. This custom exception provides a list of missing raster pixel values in a ``missing_values`` attribute that allows the caller access to the pixel values that are missing through a Python type rather than indirectly through an error message.`

richpsharp (changes requested):
Sorry I have more review questions here Doug. I didn't look at this as carefully as I should have on a first pass.

src/pygeoprocessing/geoprocessing.py
dcdenu4: Nope, that's a fair assessment and does seem redundant. I'll remove it.

src/pygeoprocessing/geoprocessing.py
dcdenu4: Good catch.

HISTORY.rst
dcdenu4: Yea, I think this is better stated then what I have. I'll update it.

dcdenu4:
Thanks for the comments Rich, I also noticed a test is failing because of the `__all__` inclusion. Looks like that test excepts all the attributes to be functions, which this is not.

HISTORY.rst
dcdenu4: Updated statement here: https://github.com/natcap/pygeoprocessing/pull/106/commits/d0bcb5436d271f8f1c58a6d44853cdbd782fb185

src/pygeoprocessing/geoprocessing.py
dcdenu4: Removed `value_map` here: https://github.com/natcap/pygeoprocessing/pull/106/commits/0ed7d528c4fe110c9333a5877e5d9bc00b902482

src/pygeoprocessing/geoprocessing.py
dcdenu4: Removed `value_map` here: https://github.com/natcap/pygeoprocessing/pull/106/commits/0ed7d528c4fe110c9333a5877e5d9bc00b902482

phargogh (approved):
👍

richpsharp (approved):

PR #108, opened 2020-08-11
Feature/multiprocess raster calculator

### LINKED ISSUE #107: create multiprocess `raster_calculator`
Create a new submodule `multiprocessing` that has an API-equivalent `raster_calculator` function. Generally this function should use additional Process objects to parallelize computation where efficient to do so.

richpsharp:
This PR adds a `pygeoprocessing.multiprocessing.raster_calculator` function which matches the API and results of the traditional `raster_calculator` operation. High level changes to note:

* Python 3.8 introduced a shared memory module which allows one to pass memory references rather than pickle large data structures between processes. This makes the calculation of `raster_stats` possible at scale by passing the memory reference to numpy arrays rather than pickling results and passing them down a queue, or loading them directly from a raster. When not used in practice, with 4 CPUs the multiprocessing manager process would max out computation with pickling/unpickling data between processes. In practice runtime for the *entire* calculation increased twofold by using shared memory blocks between raster calculation workers and the stats worker. Note this changes makes this future version of PyGeoprocessing >=3.8 compatible only. Hence I've taken out testing for Python 3.7 in the workflow.
* I've modified the stats worker module as described above which also modifies it for the primary `raster_calculator`.
* I've added a largish test for the multiprocessing raster calculator that takes more time than any test, most of this is the construction and destruction of the worker processes.
* Note updated in history.

### DIFF
--- a/.github/workflows/build-py-dists.yml
+++ b/.github/workflows/build-py-dists.yml
@@ -12,7 +12,7 @@ jobs:
                 python-arch: [x86, x64]
                 include:
                     - python-version: 3.7
-                      numpy: "numpy~=1.15"
+                      numpy: "numpy~=1.17"
                     - python-version: 3.8
                       numpy: "numpy~=1.17"
 
--- a/.github/workflows/build-py-dists.yml
+++ b/.github/workflows/build-py-dists.yml
@@ -8,12 +8,9 @@ jobs:
             fail-fast: false
             matrix:
                 os: [windows-latest, macos-latest]
-                python-version: [3.7, 3.8]
+                python-version: [3.8]
                 python-arch: [x86, x64]
                 include:
-                    - python-version: 3.7
-                      numpy: "numpy~=1.15"
-
                     - python-version: 3.8
                       numpy: "numpy~=1.17"
 
@@ -62,7 +59,7 @@ jobs:
             - name: Set up python
               uses: actions/setup-python@v1
               with:
-                  python-version: 3.7
+                  python-version: 3.8
 
             - name: Install python dependencies
               run: python -m pip install --upgrade pip numpy setuptools_scm cython
--- a/.github/workflows/build-py-dists.yml
+++ b/.github/workflows/build-py-dists.yml
@@ -8,9 +8,11 @@ jobs:
             fail-fast: false
             matrix:
                 os: [windows-latest, macos-latest]
-                python-version: [3.8]
+                python-version: [3.7, 3.8]
                 python-arch: [x86, x64]
                 include:
+                    - python-version: 3.7
+                      numpy: "numpy~=1.15"
                     - python-version: 3.8
                       numpy: "numpy~=1.17"
 
@@ -59,7 +61,7 @@ jobs:
             - name: Set up python
               uses: actions/setup-python@v1
               with:
-                  python-version: 3.8
+                  python-version: 3.7
 
             - name: Install python dependencies
               run: python -m pip install --upgrade pip numpy setuptools_scm cython
--- a/.github/workflows/build-py-dists.yml
+++ b/.github/workflows/build-py-dists.yml
@@ -12,7 +12,7 @@ jobs:
                 python-arch: [x86, x64]
                 include:
                     - python-version: 3.7
-                      numpy: "numpy~=1.17"
+                      numpy: "numpy~=1.15"
                     - python-version: 3.8
                       numpy: "numpy~=1.17"
 
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -360,9 +360,11 @@ def raster_calculator(
             exception_queue = queue.Queue()
 
             if sys.version_info >= (3, 8):
+                # The stats worker keeps running variables as a float64, so
+                # all input rasters are dtype float64 -- make the shared memory
+                # size equivalent.
                 block_size_bytes = (
-                    numpy.dtype(_GDAL_TYPE_TO_NUMPY_LOOKUP[
-                        datatype_target]).itemsize *
+                    numpy.dtype(numpy.float64).itemsize *
                     block_offset_list[0]['win_xsize'] *
                     block_offset_list[0]['win_ysize'])
 
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -500,8 +500,9 @@ def raster_calculator(
                 stats_worker_thread.join(_MAX_TIMEOUT)
                 if stats_worker_thread.is_alive():
                     raise RuntimeError("stats_worker_thread.join() timed out")
-                shared_memory.close()
-                shared_memory.unlink()
+                if sys.version_info >= (3, 8):
+                    shared_memory.close()
+                    shared_memory.unlink()
 
             # check for an exception in the workers, otherwise get result
             # and pass to writer
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -29,6 +29,7 @@ import shapely.ops
 import shapely.prepared
 import shapely.wkb
 
+
 class ReclassificationMissingValuesError(Exception):
     """Raised when a raster value is not a valid key to a dictionary.
 
@@ -42,6 +43,7 @@ class ReclassificationMissingValuesError(Exception):
         self.msg = msg
         self.missing_values = missing_values
         super().__init__(msg, missing_values)
+
 
 LOGGER = logging.getLogger(__name__)
 
@@ -354,6 +356,10 @@ def raster_calculator(
             stats_worker_queue = None
             exception_queue = None
 
+        block_offset_list = list(iterblocks(
+            (target_raster_path, 1), offset_only=True,
+            largest_block=largest_block))
+
         if calc_raster_stats:
             # To avoid doing two passes on the raster to calculate standard
             # deviation, we implement a continuous statistics calculation
@@ -375,9 +381,7 @@ def raster_calculator(
         n_pixels = n_cols * n_rows
 
         # iterate over each block and calculate local_op
-        for block_offset in iterblocks(
-                (target_raster_path, 1), offset_only=True,
-                largest_block=largest_block):
+        for block_offset in block_offset_list:
             # read input blocks
             offset_list = (block_offset['yoff'], block_offset['xoff'])
             blocksize = (block_offset['win_ysize'], block_offset['win_xsize'])
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -360,8 +360,10 @@ def raster_calculator(
             exception_queue = queue.Queue()
 
             if sys.version_info >= (3, 8):
+                # assume largest GDAL float size just in
                 block_size_bytes = (
-                    numpy.dtype(numpy.float64).itemsize *
+                    numpy.dtype(_GDAL_TYPE_TO_NUMPY_LOOKUP(
+                        datatype_target)).itemsize *
                     block_offset_list[0]['win_xsize'] *
                     block_offset_list[0]['win_ysize'])
 
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -4,11 +4,11 @@ import collections
 import functools
 import logging
 import math
-import multiprocessing.shared_memory
 import os
 import pprint
 import queue
 import shutil
+import sys
 import tempfile
 import threading
 import time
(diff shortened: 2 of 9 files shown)


phargogh:
@richpsharp after looking into python 3.8 support in pyinstaller (issue: https://github.com/pyinstaller/pyinstaller/issues/4311, outstanding PR: https://github.com/pyinstaller/pyinstaller/pull/4986), it looks like there's some work in progress on this (and the pyinstaller dev community has really stepped up in a way I wasn't expecting), but it isn't complete yet.

The other side of things is that (as far as I can tell), pyinstaller still seems like it might be the best option for the time being, though it is worth noting that cx_freeze (another python binary freezing project) [claims to have added support for python 3.8](https://cx-freeze.readthedocs.io/en/latest/releasenotes.html?highlight=3.8#version-6-1-january-2020), so this might be another option for us if pyinstaller isn't worth keeping around.

Given our past experiences debugging pyinstaller builds and the uncertainty around the alternative (finding a binary build utility that does work), I think it'd be safest for us to be python 3.7-compatible for the moment and then plan to migrate to 3.8+ when we can safely do so.

Oh, but also, FWIW, I'd like to _try_ `cx_freeze` and see how hard it is to assemble a set of binaries for InVEST.

richpsharp:
> @richpsharp after looking into python 3.8 support in pyinstaller (issue: [pyinstaller/pyinstaller#4311](https://github.com/pyinstaller/pyinstaller/issues/4311), outstanding PR: [pyinstaller/pyinstaller#4986](https://github.com/pyinstaller/pyinstaller/pull/4986)), it looks like there's some work in progress on this (and the pyinstaller dev community has really stepped up in a way I wasn't expecting), but it isn't complete yet.
> 
> The other side of things is that (as far as I can tell), pyinstaller still seems like it might be the best option for the time being, though it is worth noting that cx_freeze (another python binary freezing project) [claims to have added support for python 3.8](https://cx-freeze.readthedocs.io/en/latest/releasenotes.html?highlight=3.8#version-6-1-january-2020), so this might be another option for us if pyinstaller isn't worth keeping around.
> 
> Given our past experiences debugging pyinstaller builds and the uncertainty around the alternative (finding a binary build utility that does work), I think it'd be safest for us to be python 3.7-compatible for the moment and then plan to migrate to 3.8+ when we can safely do so.
> 
> Oh, but also, FWIW, I'd like to _try_ `cx_freeze` and see how hard it is to assemble a set of binaries for InVEST.

Okay, no problem on this. I added backwards comparable non shared memory support in an update.

richpsharp:
Okay @phargogh, I refactored this to be 3.7/3.8 compliant. I did change one thing that I don't know what it was for in a github action. It's in `build-py-dists.yml` and looks like this:

```
- python-version: 3.7
         numpy: "numpy~=1.17"
```

That was at 1.15 before and I had a weird error in the build that I didn't have on my side about a `ndarray`. I'm not sure what all that infrastructure is there so just FYI if that's something specifically to review.

.github/workflows/build-py-dists.yml
```diff
@@ -12,8 +12,7 @@ jobs:
                 python-arch: [x86, x64]
                 include:
                     - python-version: 3.7
-                      numpy: "numpy~=1.15"
-
+                      numpy: "numpy~=1.17"
                     - python-version: 3.8
                       numpy: "numpy~=1.17"
```
phargogh: Yeah, this is a weird one and I'm not sure offhand why `numpy~=1.15` was causing such errors.  I've found over the years that compiling our wheel-distributed cython modules against the earliest-possible version of `numpy` will avoid failed imports later on when we get a `ValueError: numpy.dtype has the wrong size, try recompiling`.  Updating to 1.17 should be fine, 1.16 might work, too.  The main point of all this is that we probably shouldn't compile against the very latest version of numpy because we'll be way way more likely to get this exception and pygeoprocessing doesn't literally need the latest version of numpy in order to work.

If memory serves, `numpy~=1.15` was the first version to support python 3.7, and `numpy~=1.17` was the first to support python 3.8.  So that's where those versions came from.  I'll make a note to add a note about these versions to eh yml file.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -345,14 +349,27 @@ def raster_calculator(
     try:
         last_time = time.time()
 
+        block_offset_list = list(iterblocks(
+            (target_raster_path, 1), offset_only=True,
+            largest_block=largest_block))
+
```
phargogh, line 355: I almost had a comment about not needing the length attribute of the list here, but I see that you've updated the worker queues to only expect this many blocks, which seems like a neat solution!

src/pygeoprocessing/geoprocessing.py
```diff
@@ -345,14 +349,27 @@ def raster_calculator(
...
         if calc_raster_stats:
             # if this queue is used to send computed valid blocks of
             # the raster to an incremental statistics calculator worker
             stats_worker_queue = queue.Queue()
             exception_queue = queue.Queue()
+
+            if sys.version_info >= (3, 8):
+                block_size_bytes = (
+                    numpy.dtype(numpy.float64).itemsize *
```
phargogh, line 367: Could you perhaps add a note here about why `numpy.float64` was chosen?  I suspect it's because this is the largest value we'll see from GDAL rasters, but future us might wonder.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -3642,6 +3666,7 @@ def numpy_array_to_raster(
         None
     """
     numpy_to_gdal_type = {
+        numpy.dtype(numpy.bool): gdal.GDT_Byte,
```
phargogh, line 3686: This is kind of a user-facing change ... would it be worth noting in `HISTORY`?

src/pygeoprocessing/multiprocessing/raster_calculator.py
```diff
@@ -0,0 +1,589 @@
...
+                callback_state['total_blocks']*100:.2f}% complete""")
+        callback_state['last_time'] = time.time()
+
+
+class RasterPathBand():
+    """Abstraction for the (str, int) tuple that represents a raster/band."""
+    def __init__(self, path, band_id):
+        """Copy path and band_id directly."""
+        self.path = path
+        self.band_id = band_id
```
phargogh: Is there an advantage to using a class here over a `collections.namedtuple`?  If attribute access is the only thing we're really looking for, `namedtuple` would still allow attribute access while still being immutable, which I think might be more in line with what we're trying to express here but I'd love to hear what you think!

src/pygeoprocessing/multiprocessing/raster_calculator.py
```diff
@@ -0,0 +1,589 @@
...
+
+
+def _build_raster_calc_error_handler(pool):
+    def _raster_calc_error_handler(exception):
+        """Error handler for raster_calculator."""
+        pool.terminate()
+        raise Exception(
+            f"error on raster calculator worker '{exception}'").with_traceback(
+                exception.__traceback__)
```
phargogh, line 63: Well that `with_traceback` is slick.  Learned something new today!

src/pygeoprocessing/multiprocessing/raster_calculator.py
```diff
@@ -0,0 +1,589 @@
...
+        stats_worker_queue, nodata_target, target_raster_path, write_lock,
+        processing_state, result_array_shared_memory):
+    """Process a single block of an array for raster_calculation.
+
+    Args:
+        block_offset (dict): contains 'xoff', 'yoff', 'xwin_size', 'ywin_size'
+            and can be used to pass directly to Band.ReadAsArray.
+        base_canonical_arg_list (list): list of RasterPathBand, numpy arrays,
+            or 'raw' objects to pass to the ``local_op``.
+        stats_worker_queue (queue): pass a shared memory object `local_op`
```
phargogh: Hey, sorry, looks like `local_op` ended up being markdown instead of RST.

phargogh: It also looks like the `local_op` argument is not described in this docstring.

src/pygeoprocessing/multiprocessing/raster_calculator.py
```diff
@@ -0,0 +1,589 @@
...
+
+    return n_cols, n_rows
+
+
+def _validate_raster_input(
+        base_raster_path_band_const_list, raster_info_list,
+        target_raster_path):
+    """Check for valid raster/arg inputs and output.
+
+    Args:
```
phargogh, line 289: Looks like the `raster_info_list` parameter is missing from this docstring.

src/pygeoprocessing/multiprocessing/raster_calculator.py
```diff
@@ -0,0 +1,589 @@
...
+    if datatype_target not in _VALID_GDAL_TYPES:
+        raise ValueError(
+            'Invalid target type, should be a gdal.GDT_* type, received '
+            '"%s"' % datatype_target)
+
+    # create target raster
+    raster_driver = gdal.GetDriverByName(raster_driver_creation_tuple[0])
+    try:
+        os.makedirs(os.path.dirname(target_raster_path))
+    except OSError:
```
phargogh: Could you describe a circumstancs where we'd have an `OSError` that won't cause the rest of this `raster_calculator` call to crash?

phargogh (changes requested):
This looks really neat @richpsharp !  I had a few fairly minor comments, but overall the structure of this looks really great and I don't have any major suggestions.  Let me know what you think!

.github/workflows/build-py-dists.yml
richpsharp: Okay, I reverted this in `1ebed4d5aa0263740b2230ffa2652f3b3a13002a`

src/pygeoprocessing/geoprocessing.py
richpsharp: Figured I might as well since I'd need it to generate reliable logging messages.

src/pygeoprocessing/geoprocessing.py
richpsharp: That was the original reason, but thinking about it for a second, we already know what the target type is so why not just allocate that much ram only? changed in 1f19d0d4851df3d338709130ab5193695bc5634a

src/pygeoprocessing/geoprocessing.py
richpsharp: Noted in 6822131dd298b54e88b6504656bd29e62f48b61a

src/pygeoprocessing/multiprocessing/raster_calculator.py
richpsharp: Sure, no reason for the class, I just didn't think of using a named tuple. Changed in 1d77b160eb39f43df2dfc880e88c467cd989a9f4

src/pygeoprocessing/multiprocessing/raster_calculator.py
richpsharp: Not just slick but critical to early debugging of this method!

src/pygeoprocessing/multiprocessing/raster_calculator.py
richpsharp: Fixed in 3c5930b2714967c37bd122737536d6adc1bce5ec

src/pygeoprocessing/multiprocessing/raster_calculator.py
richpsharp: Fixed in 3c5930b2714967c37bd122737536d6adc1bce5ec

src/pygeoprocessing/multiprocessing/raster_calculator.py
richpsharp: It's such a habit to write that pair I don't think about it anymore! I added an explicit check for the kind of exception in 65abdd59dd126c5273f321898de08c423bc6fe28

src/pygeoprocessing/geoprocessing.py
richpsharp: Actually the real reason was because it was supposed to be a `float64`! The stats worker accumulates variables in that type since it's high precision, so we gotta make sure the underlying memory can hold it too. I added a comment for that.

richpsharp:
Hi @phargogh, I think I addressed all of your issues, ready for re-review.

.github/workflows/build-py-dists.yml
phargogh: Oh sorry, I wasn't trying to say that it should be changed back, only provide some context as to what this version means.  If it builds as it was, then great!

src/pygeoprocessing/multiprocessing/raster_calculator.py
phargogh: Looks like this was patched in 3c5930b2714967c37bd122737536d6adc1bce5ec as well.

phargogh (approved):
Thanks @richpsharp !  Looks great to me.

PR #110, opened 2020-08-20
fixes #109 guarding against "ignore edge/nodata" but "don't mask nodata"

### LINKED ISSUE #109: convolve 2d bug?
This is from @dcdenu4: https://drive.google.com/file/d/1R3QsQGQCud6-QNjtbodaz6fm5mawliu6/view?usp=sharing

Summary
=======

When running ``pygeoprocessing.convolve_2d`` on a raster with patched values 
surrounded my nodata values ( like spots on a cow ) the output is corrupt. 
When replacing the nodata values for that raster with 0s, the output is as 
expected.

Files
=====

* convolution-bug-script.py
    A python script with the paths defined to replicate the issue. This calls
    ``convolve_2d`` on ``urban_c_aligned.tif`` and ``kernel_urban_c.tif``. 
    The output is ``output_convolution.tif``.
* kernel_urban_c.tif
    The exponential kernel for convolution.
* urban_c_aligned.tif
    The raster with patchy values of 1 and otherwise nodata values. Nodata 
    value is 0
* urban_c_aligned_zeros.tif
    The same raster as ``urban_c_aligned.tif`` except nodata has been set to -9
    and original nodata values are now 0s

richpsharp:
Addresses a nonsensical combination of parameters for `convolve_2d` where `ignore_nodata_and_edges=True` and `mask_nodata=False`. Rather than allow this combination and return numerical noise, the function now raises a `ValueError`. Updated docstring, tests, and history to reflect this change.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,10 +3,14 @@ Release History
 
 Unreleased Changes
 ------------------
-* Added a custom exception class ``ReclassificationMissingValuesError`` to 
+* Modified ``pygeoprocessing.convolve_2d`` to guard against nonsensical queries
+  to both ``ignore_nodata_and_edges=True`` but also ``mask_nodata=False``.
+  A query of this combination now raises a ``ValueError`` to guard against
+  programmer error.
+* Added a custom exception class ``ReclassificationMissingValuesError`` to
   ``pygeoprocessing``. ``pygeoprocessing.reclassify_raster`` raises this
   exception instead of ``ValueError`` when a raster pixel value is not
-  represented in ``value_map``. This custom exception provides a list of 
+  represented in ``value_map``. This custom exception provides a list of
   missing raster pixel values in a ``missing_values`` attribute that allows
   the caller access to the pixel values that are missing through a Python type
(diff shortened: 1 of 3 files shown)


dcdenu4 (changes requested):
Hey @richpsharp I think this makes sense to me. If we're ignoring `nodata` values but not masking them, then those values are going to be nonsensical. 

Is it worth having a shell test that makes sure the added `ValueError` is caught? I know it seems pretty trivial but maybe good to cover those instances?

richpsharp:
> Hey @richpsharp I think this makes sense to me. If we're ignoring `nodata` values but not masking them, then those values are going to be nonsensical.
> 
> Is it worth having a shell test that makes sure the added `ValueError` is caught? I know it seems pretty trivial but maybe good to cover those instances?

Oh totally -- and it's not trival at all, it's necessary. And ALSO I'd already written that test and I don't know what happened as to why it's not in there. But I made it again and did a push.

dcdenu4 (approved):
I see it now! Thanks!

PR #111, opened 2020-08-23
fixes #102, better logging message and update to history

### LINKED ISSUE #102: Logging on `new_raster_from_base` is not informative
Logging from `new_raster_from_base` is not helpful w/r/t a user understanding what function is running. Rather it says "% complete" without further info. Make that logging message give enough information that the user can infer what's happening. I feel there are other logging messages like this across pygeoprocessing, so when addressing this issue check any other log messages too.

richpsharp:
This PR fixes an issue that would report `"pygeoprocessing.geoprocessing [<lambda>:911]  3.21%% complete"` messages when a new raster was being filled with a value. This process can take a long time in cases with a large raster and the above message was confusing when used in the context of another function, like `convolve_2d`. This PR changes the log message to report the target raster, the fill value, and the percent complete rounded to two decimal points:

```
2020-08-22 17:35:37,884 (5994) INFO pygeoprocessing.geoprocessing [<lambda>:911] filling new raster convolve_out.tif with 0 -- 3.58% complete
```

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -23,6 +23,9 @@ Unreleased Changes
   This parameter defaults to ``False``.  This is a change from prior behavior,
   when the diagnostic vectors were always created, which could occupy
   significant computational time under large outflow geometries.
+* Modified logging message for ``pygeoprocessing.new_raster_from_base`` when
+  filling a raster such that an informative error message is printed with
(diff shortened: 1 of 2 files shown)


dcdenu4 (approved):
Looks pretty straight forward to me!

PR #113, opened 2020-08-23
Modifications to improve performance of convolve_2d

### LINKED ISSUE #112: Performance improvements to convolve_2d
Convolve 2D can be slow for large rasters. One slow, and inefficient step is filling the target raster with 0s on a first pass. This issue will address such inefficiencies with an eye for measurable performance on large raster data.

richpsharp:
This PR is the result of experimenting with a variety of parallel processing techniques as well as pre-calculation of kernel FFTs or signal FFTs and more. By experimenting on a 75x75 kernel and a 12600x64000 signal raster, I found the following changes were helpful:
* Removed a "pre-fill" step for the target raster and mask if needed by the `ignore_nodata...` flag. Instead I added a set that determines if a particular block has been loaded before, and if not, defaults it to 0s. This saves having to fill an entire raster with zeros that just get overwritten later. In my example this saved a couple minutes of runtime off of a total runtime of around an hour.
* I found that the `_fft_cache...` functions did not significantly affect runtime performance in any way that I could measure, so I took them out.
* I added an asyncronous step that loads work into the work queue rather than stopping the main thread to do so. This allows the `_convolve_2d_worker` to start working right away rather than waiting for a full queue. This has a noticeable kickstart to the convolve function when comparing to large workloads and will have a small impact in total runtime since work can continue while work is loaded.
* I found that the addition of any more than 1 worker thread did not increase performance, so I removed the `n_threads` option. Since this is no longer an option, I removed the test that tested this functionality.
* I pushed the test for setting small absolute values to 0 to the worker rather than the main thread, this didn't make a performance difference one way or the other, but it does push all the "work" to a worker and leaves the main thread only to process data.

In all this PR also slightly reduces the footprint of this function and the complexity of the codebase.

Fixes #112

### DIFF
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -2494,26 +2494,35 @@ def convolve_2d(
         kernel_sum += numpy.sum(kernel_block)
 
     work_queue = multiprocessing.Queue()
-    LOGGER.info(f'fill the work queue')
-    n_blocks = 0
-    for signal_offset in iterblocks(s_path_band, offset_only=True):
-        for kernel_offset in iterblocks(k_path_band, offset_only=True):
-            work_queue.put((signal_offset, kernel_offset))
-            n_blocks += 1
-    for _ in range(max(1, n_threads-1)):
-        # signal end to worker
-        work_queue.put(None)
+    signal_offset_list = list(iterblocks(s_path_band, offset_only=True))
+    kernel_offset_list = list(iterblocks(k_path_band, offset_only=True))
+    n_blocks = len(signal_offset_list) * len(kernel_offset_list)
+
+    def _fill_work_queue():
+        LOGGER.debug('fill work queue')
+        for signal_offset in signal_offset_list:
+            for kernel_offset in kernel_offset_list:
+                work_queue.put((signal_offset, kernel_offset))
+        for _ in range(max(1, n_threads-1)):
+            # signal end to worker
+            work_queue.put(None)
+        LOGGER.debug('work queue full')
+
+    fill_work_queue_worker = threading.Thread(
+        target=_fill_work_queue)
+    fill_work_queue_worker.daemon = True
+    fill_work_queue_worker.start()
 
     # limit the size of the write queue so we don't accidentally load a whole
     # array into memory, work queue is okay because it's only passing block
     # indexes
     write_queue = multiprocessing.Queue(n_threads * 2)
-    LOGGER.info(f'start {n_threads} worker threads')
+    LOGGER.debug(f'start {n_threads} worker threads')
     worker_list = []
     # workers is n_threads - 1 because we count the current thread
     for worker_id in range(max(1, n_threads-1)):
-        worker = threading.Thread(
-        # worker = multiprocessing.Process(
+        #worker = threading.Thread(
+        worker = multiprocessing.Process(
             target=_convolve_2d_worker,
             args=(
                 signal_path_band, kernel_path_band,
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -11,6 +11,7 @@ import shutil
 import tempfile
 import threading
 import time
+import multiprocessing
 
 from . import geoprocessing_core
 from .geoprocessing_core import DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS
@@ -2449,7 +2450,6 @@ def convolve_2d(
         target_nodata = numpy.finfo(numpy.float32).min
     new_raster_from_base(
         signal_path_band[0], target_path, target_datatype, [target_nodata],
-        fill_value_list=[0],
         raster_driver_creation_tuple=raster_driver_creation_tuple)
 
     signal_raster_info = get_raster_info(signal_path_band[0])
@@ -2477,8 +2477,7 @@ def convolve_2d(
         mask_raster_path = os.path.join(mask_dir, 'convolved_mask.tif')
         new_raster_from_base(
             signal_path_band[0], mask_raster_path, gdal.GDT_Float32,
-            [-1.0], fill_value_list=[0],
-            raster_driver_creation_tuple=raster_driver_creation_tuple)
+            [-1.0], raster_driver_creation_tuple=raster_driver_creation_tuple)
         mask_raster = gdal.OpenEx(
             mask_raster_path, gdal.GA_Update | gdal.OF_RASTER)
         mask_band = mask_raster.GetRasterBand(1)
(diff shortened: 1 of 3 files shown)


richpsharp:
Hi @dcdenu4, I'm putting you as a reviewer here since you were just doing some `convolve_2d` stuff a few days ago, so hopefully it's fresh on your mind?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2494,37 +2486,49 @@ def convolve_2d(
...
+    worker = threading.Thread(
+        target=_convolve_2d_worker,
+        args=(
+            signal_path_band, kernel_path_band,
+            ignore_nodata_and_edges, normalize_kernel,
+            set_tol_to_zero, work_queue, write_queue))
+    worker.daemon = True
+    worker.start()
 
     # used to count how many workers are still running
```
dcdenu4: is this comment needed anymore since `n_active_workers` is gone?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2591,16 +2604,15 @@ def convolve_2d(
             _LOGGING_PERIOD)
 
     LOGGER.info(
-        "convolution worker 100.0%% complete on %s",
-        os.path.basename(target_path))
+        f"convolution worker 100.0%% complete on "
```
dcdenu4: Does this double percent signify something or a typo?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2631,17 +2643,13 @@ def convolve_2d(
                     os.path.basename(target_path)),
                 _LOGGING_PERIOD)
         # delete the mask raster
```
dcdenu4: Should this comment make more sense now as: dereference mask_raster

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2631,17 +2643,13 @@ def convolve_2d(
...
+        mask_band = None
         os.remove(mask_raster_path)
         LOGGER.info(
             "convolution nodata normalize 100.0%% complete on %s",
             os.path.basename(target_path))
 
-    for worker in worker_list:
-        worker.join(_MAX_TIMEOUT)
-    target_band.FlushCache()
-    target_raster.FlushCache()
```
dcdenu4, line 2644: I think I remember a discussion based around these not being needed anymore, but I can't remember the reasoning?

src/pygeoprocessing/geoprocessing.py
richpsharp: Removed in 34802e5be006a4dbd7a9e0bb18330478cec6154a

src/pygeoprocessing/geoprocessing.py
richpsharp: I learned something here... There are a lot of old strings that use `printf` style formatting, like `('percent complete %f%%' % count/100)`. You need that `%%` there, otherwise a % is interpreted as a replacement of something. I used this so much I assumed % always had to be escaped with another %, but that's only the case if you do the % replacement i.e. `'this is fine %'` `'but this crashes % %f' % 0.1`.  

Anyway, long way to say I removed that % and a couple others in afa1524663cac498b954923f33ed02b61d92389d

src/pygeoprocessing/geoprocessing.py
richpsharp: It actually is deleted though, see `os.remove(mask_raster_path)`. I suppose that doesn't need a comment though? Removed in 05572f475c42081e9242b4b01e534cf7cc067399

src/pygeoprocessing/geoprocessing.py
dcdenu4: Oh, also don't need the leading `f` there since nothing is being string replaced.

src/pygeoprocessing/geoprocessing.py
richpsharp: I think we were talking about the `__swig_destroy__` calls we used to do, but in this case, flushing the cache is not necessary since the very next lines dereference the object anyway and GDAL 3+ (and earlier?) say this is the *only* way you can guarantee data are written to disk. So `FlushCache` is not necessary and might not even work anyway.

dcdenu4 (changes requested):
Hey @richpsharp This looks okay to me for the most part. I had some very minor comments and questions. Great to see a little boost to this core algorithm.

richpsharp:
Thanks @dcdenu4, I think I addressed all your comments and re-requested your review.

dcdenu4:
> Thanks @dcdenu4, I think I addressed all your comments and re-requested your review.

Woah! Super fast! Thanks, I'll take a look.

src/pygeoprocessing/geoprocessing.py
dcdenu4: Ha! Sorry, I was so focused on the removal of the flush cache I didn't even see the `os.remove` down there.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2495,37 +2487,48 @@ def convolve_2d(
             kernel_block[numpy.isclose(kernel_block, kernel_nodata)] = 0.0
         kernel_sum += numpy.sum(kernel_block)
 
+    work_queue = queue.Queue()
+    signal_offset_list = list(iterblocks(s_path_band, offset_only=True))
+    kernel_offset_list = list(iterblocks(k_path_band, offset_only=True))
+    n_blocks = len(signal_offset_list) * len(kernel_offset_list)
+
+    LOGGER.debug(f'start fill work queue thread')
```
dcdenu4: no leading `f` necessary here.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2495,37 +2487,48 @@ def convolve_2d(
...
+    fill_work_queue_worker.start()
+
     # limit the size of the write queue so we don't accidentally load a whole
     # array into memory, work queue is okay because it's only passing block
     # indexes
-    work_queue = queue.Queue()
-    write_queue = queue.Queue(n_threads * 2)
-
-    # workers is n_threads - 1 because we count the current thread
+    LOGGER.debug(f'start worker thread')
```
dcdenu4: another `f` here with no string replacement.

dcdenu4 (changes requested):
Hey, sorry about this but I spotted a couple more strings with unnecessary leading `f` with no string replacement being done.

src/pygeoprocessing/geoprocessing.py
richpsharp: Removed in 79409027d7d280b665fd2812aa4437350acfa59b

src/pygeoprocessing/geoprocessing.py
richpsharp: Removed in 79409027d7d280b665fd2812aa4437350acfa59b

richpsharp:
> Hey, sorry about this but I spotted a couple more strings with unnecessary leading `f` with no string replacement being done.

No problem, I removed those and re-requested.

dcdenu4 (approved):
Thanks!

PR #114, opened 2020-08-25
2.1.0 release -- only changed 2.1.0 in HISTORY.rst header

richpsharp:


phargogh:
Content-wise the changes look good to me, but @richpsharp  is that one failing test cause for concern?

richpsharp:
> Content-wise the changes look good to me, but @richpsharp is that one failing test cause for concern?

Oh boy I didn't expect that! Lemme figure that out. There may be a 2.1.1 release I guess?

richpsharp:
> > Content-wise the changes look good to me, but @richpsharp is that one failing test cause for concern?
> 
> Oh boy I didn't expect that! Lemme figure that out. There may be a 2.1.1 release I guess?

So I ran the tests again and they passed? It was a test failing in `multiprocessing.raster_calculator` that I didn't understand, almost looked like the wrong version of numpy was compiled but that doesn't make sense if the tests pass later so I'm guessing a concurrency bug. Lets keep an eye out for it and be careful about putting `multiprocessing.raster_calculator` into default production -- an option to turn that on is always nice.

phargogh (approved):
Approving and we'll watch out for that possible issue in the `multiprocessing.raster_calculator` tests!

PR #116, opened 2020-08-26
Bugfix/convolve 2d overlap 115

richpsharp:
Fixing an issue introduced by optimization steps in 2.1.0. Also added a test that would have caught this bug in the 2.1.0 (tested on 2.1.0 and it fails, tested on this branch and it passes).

### DIFF
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -2551,9 +2551,6 @@ def convolve_2d(
     worker.start()
 
     n_blocks_processed = 0
-
-    # this set is indexed by (x,y) tuple from iterblock offsets to check if the
-    # target block is on its first pass and set to zero if so.
     LOGGER.info(f'{n_blocks} sent to workers, wait for worker results')
     while True:
         write_payload = write_queue.get()
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -2467,10 +2467,11 @@ def convolve_2d(
             "Expected raster path band sequences for the following arguments "
             f"but instead got: {bad_raster_path_list}")
 
-    if target_nodata is None:
-        target_nodata = numpy.finfo(numpy.float32).min
+    # The nodata value is reset to a different value at the end of this
+    # function. Here 0 is chosen as a default value since data are
+    # incrementally added to the raster
     new_raster_from_base(
-        signal_path_band[0], target_path, target_datatype, [target_nodata],
+        signal_path_band[0], target_path, target_datatype, [0],
         raster_driver_creation_tuple=raster_driver_creation_tuple)
 
     signal_raster_info = get_raster_info(signal_path_band[0])
@@ -2497,8 +2498,8 @@ def convolve_2d(
         mask_dir = tempfile.mkdtemp(dir=working_dir)
         mask_raster_path = os.path.join(mask_dir, 'convolved_mask.tif')
         new_raster_from_base(
-            signal_path_band[0], mask_raster_path, gdal.GDT_Float32,
-            [-1.0], raster_driver_creation_tuple=raster_driver_creation_tuple)
+            signal_path_band[0], mask_raster_path, gdal.GDT_Float64,
+            [0.0], raster_driver_creation_tuple=raster_driver_creation_tuple)
         mask_raster = gdal.OpenEx(
             mask_raster_path, gdal.GA_Update | gdal.OF_RASTER)
         mask_band = mask_raster.GetRasterBand(1)
@@ -2540,7 +2541,6 @@ def convolve_2d(
     # indexes
     LOGGER.debug('start worker thread')
     write_queue = queue.Queue(10)
-    worker_list = []
     worker = threading.Thread(
         target=_convolve_2d_worker,
         args=(
@@ -2554,7 +2554,6 @@ def convolve_2d(
 
     # this set is indexed by (x,y) tuple from iterblock offsets to check if the
     # target block is on its first pass and set to zero if so.
-    target_block_initalized = set()
     LOGGER.info(f'{n_blocks} sent to workers, wait for worker results')
     while True:
         write_payload = write_queue.get()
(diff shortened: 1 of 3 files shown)


tests/test_geoprocessing.py
```diff
@@ -4186,3 +4186,40 @@ def test_convolve_2d_nodata(self):
...
+        """PGP.geoprocessing: test convolve 2d with large gaussian kernel."""
+        # choosing twice the max memory block
+        n_pixels = 256*2
+        # this is a fun seed random seed
+        numpy.random.seed(n_pixels)
+        signal_array = numpy.random.random((n_pixels, n_pixels))
+
+        workspace_dir = 'local'
+        if not os.path.exists(workspace_dir):
+            os.makedirs(workspace_dir)
```
dcdenu4: Should this be `self.workspace_dir`? so that it gets deleted on cleanup?

tests/test_geoprocessing.py
```diff
@@ -4186,3 +4186,40 @@ def test_convolve_2d_nodata(self):
...
+            (signal_path, 1), (kernel_path, 1), target_path,
+            ignore_nodata_and_edges=False, mask_nodata=True)
+        target_array = pygeoprocessing.raster_to_numpy_array(target_path)
+
+        # gaussian filter with constant is the same as bleeding off the edges
+        expected_output = scipy.ndimage.gaussian_filter(
+            signal_array, 1.0, mode='constant')
+
+        _array_to_raster(
+            expected_output, None, os.path.join(workspace_dir, 'extected.tif'))
```
dcdenu4: Not a big deal, but was this supposed to be `expected.tif`?

dcdenu4: Also, is there any reason to save this array to a raster?

tests/test_geoprocessing.py
```diff
@@ -4186,3 +4186,40 @@ def test_convolve_2d_nodata(self):
...
+
+        target_path = os.path.join(workspace_dir, 'target.tif')
+        pygeoprocessing.convolve_2d(
+            (signal_path, 1), (kernel_path, 1), target_path,
+            ignore_nodata_and_edges=False, mask_nodata=True)
+        target_array = pygeoprocessing.raster_to_numpy_array(target_path)
+
+        # gaussian filter with constant is the same as bleeding off the edges
+        expected_output = scipy.ndimage.gaussian_filter(
+            signal_array, 1.0, mode='constant')
```
dcdenu4, line 4219: This is pretty slick, basically created a convolution test case that mimics a scipy guassian filter process and test against it.

tests/test_geoprocessing.py
```diff
@@ -4186,3 +4186,40 @@ def test_convolve_2d_nodata(self):
...
         expected_output[-1, -1] = 1
         numpy.testing.assert_allclose(target_array, expected_output)
+
+    def test_convolve_2d_gaussian(self):
+        """PGP.geoprocessing: test convolve 2d with large gaussian kernel."""
+        # choosing twice the max memory block
+        n_pixels = 256*2
+        # this is a fun seed random seed
+        numpy.random.seed(n_pixels)
+        signal_array = numpy.random.random((n_pixels, n_pixels))
```
dcdenu4: Nice. It does appear that Numpy frowns upon reseeding the BitGenerator (https://numpy.org/doc/stable/reference/random/generated/numpy.random.seed.html) but, it's probably fine. Also, I'm not sure what numpy version we're sticking to at the moment. It looks like they've recently reworked their random API.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2540,7 +2541,6 @@ def _fill_work_queue():
     # indexes
     LOGGER.debug('start worker thread')
     write_queue = queue.Queue(10)
-    worker_list = []
```
dcdenu4, line 2543: This was left over from the last changes and not removed I take it?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2554,7 +2554,6 @@ def _fill_work_queue():
 
     # this set is indexed by (x,y) tuple from iterblock offsets to check if the
     # target block is on its first pass and set to zero if so.
```
dcdenu4: Can this comment go as well?

dcdenu4 (changes requested):
Hey @richpsharp thanks for this fix. I tested it on my data and it works. I just have some minor questions, suggestions, and comments!

tests/test_geoprocessing.py
richpsharp: Oh gosh of course it should, I do this a lot when I'm debugging tests and then I also forget to change it back. Changed in 3349d4b9eb4ba5d09fc8ba576e8fb13e00c8e657

tests/test_geoprocessing.py
richpsharp: Nope! Same as above, left over from testing/debugging. Removed in 3a5f174093f133daeb8f60de269d8220a2341971

tests/test_geoprocessing.py
richpsharp: Thanks! Trying to get some confidence!

tests/test_geoprocessing.py
richpsharp: I did not know about this! I guess I get why they want to do that -- no global state? I changed it in ad2bf39ea328fe9d71249e1f86f8aefd4ff7a0ed

src/pygeoprocessing/geoprocessing.py
richpsharp: Yes!

src/pygeoprocessing/geoprocessing.py
richpsharp: Yes, removed in 8afbb2e387c996f4905c3edf02a6802596e2c334

richpsharp:
> Hey @richpsharp thanks for this fix. I tested it on my data and it works. I just have some minor questions, suggestions, and comments!

Thanks @dcdenu4 Dough, those were good comments. I addressed all of them and pushed and re-requesting your review.

dcdenu4 (approved):
Thanks Rich! Looks good to me.

PR #120, opened 2020-09-11
Allow a user-defined layer name in watershed delineation

### LINKED ISSUE #119: Primary vector layer(s) of watershed delineation should include the suffix
Watershed delineation currently produces a geopackage with a couple of output layers, the details of which depend on the input parameters to the function.  All non-diagnostic output layers should include the file suffix, if one is provided.  Alternatively, if there is only 1 standard output layer, we could also just have the layer name match the filename.

Related to https://github.com/natcap/invest/issues/56

phargogh:
This PR allows layer names to be defined in watershed delineation according to a user-defined parameter.  This is mostly useful as an addition to https://github.com/natcap/invest/issues/56 for use in DelineateIt.  Once this PR is merged, and the feature released, we'll also need to update DelineateIt to use this new feature.

Fixes #119

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -5,6 +5,9 @@ Unreleased Changes
 ------------------
 * Fixed a critical bug introduced in 2.1.0 that generated invalid results in
   `convolve_2d` for any raster larger than 256x256.
+* Added an optional parameter, ``target_layer_name`` to
+  ``pygeoprocessing.routing.delineate_watersheds_d8`` for cases where a caller
+  would like to define the output layer name.
 
 2.1.0 (2020-08-25)
 ------------------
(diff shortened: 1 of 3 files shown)


richpsharp (approved):
Hey @phargogh this looks good to me, but I'm slightly confused by the PR title of "suffix in watershed delineation". This just changes the layer name, not the file suffix right? I've approved because I think they're good changes but mentioning just in case there's an issue or wording mixed up somewhere.

richpsharp:
Oh so if it looks okay to you @phargogh, feel free to merge it.

phargogh:
@richpsharp yes, you're right that the PR name is a bit of a misnomer.  This PR exposes the layer name to the user for customization when that is desired.  The use case for InVEST does happen to be related to the suffix (Stacie asked if the watersheds vector could have its suffix added to the layer name), but this PR allows for the layer to be defined as whatever the user would like, and DelineateIt, when updated to accommodate this PR, will include the suffix in the layer name.

I'll update the PR title to reflect this.

PR #127, opened 2020-11-06
Fixes #124, assessing a _safe_rmtree and updating history

### LINKED ISSUE #124: convolve_2d race condition in removing file
A user working with the habitat quality model posting an interesting error from `convolve_2d` that looks like a race condition between removing `mask_raster_path`. The user post with the full output log is here: https://community.naturalcapitalproject.org/t/filenotfounderror-winerror-2/1408

```python
2020-10-10 00:13:36,506 geoprocessing.convolve_2d(2568) INFO convolution nodata normalize 100.0% complete on filtered_ag_per_c.tif
2020-10-10 00:13:37,047 model._logged_target(1632) ERROR Exception while executing <function execute at 0x1C221540>
  File "site-packages\natcap\invest\ui\model.py", line 1629, in _logged_target
  File "site-packages\natcap\invest\habitat_quality.py", line 531, in execute
  File "site-packages\pygeoprocessing\geoprocessing.py", line 2580, in convolve_2d
  File "shutil.py", line 516, in rmtree
  File "shutil.py", line 400, in _rmtree_unsafe
  File "shutil.py", line 398, in _rmtree_unsafe
FileNotFoundError: [WinError 2] El sistema no puede encontrar el archivo especificado: 'C:\\Users\\RRomo\\AppData\\Local\\Temp\\tmpuwwh4_ng\\convolved_mask.tif'
2020-10-10 00:13:37,085 model._logged_target(1635) INFO Execution finished
2020-10-10 00:13:38,742 utils.prepare_workspace(117) INFO Elapsed time: 14h 7m 16.290000000000873s
```

The error looks related to this Python issue here: https://bugs.python.org/issue29699.

The user looks to have been using quite a large dataset as the convolution was running for **10+ hours** before failing.

richpsharp:
This PR wraps all calls to `shutil.rmtree` in a "safe" version that guards against the Python race condition related to `shutil.rmtree` described in https://bugs.python.org/issue29699 and updates `HISTORY.rst` indicating that.

On my own I can't recreate that issue easily in a test suite so I think it's safe to leave it without a test... that function is called by many other regression covered tests in PyGeoprocessing, but please ensure to inspect the code flow carefully since there's no easy way to test that it covers that issue.

### DIFF
--- a/src/pygeoprocessing/geoprocessing_core.pyx
+++ b/src/pygeoprocessing/geoprocessing_core.pyx
@@ -25,6 +25,7 @@ from libc.stdio cimport fwrite
 from libcpp.vector cimport vector
 from osgeo import gdal
 import numpy
+
 import pygeoprocessing
 
 DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS = ('GTIFF', (
@@ -800,7 +801,7 @@ def _raster_band_percentile_int(
             # you never know if this might fail!
             LOGGER.warning('unable to remove %s', file_path)
     if rm_dir_when_done:
-        shutil.rmtree(working_sort_directory)
+        _safe_rmtree(working_sort_directory)
     return result_list
 
 
@@ -941,6 +942,38 @@ def _raster_band_percentile_double(
             # you never know if this might fail!
             LOGGER.warning('unable to remove %s', file_path)
     if rm_dir_when_done:
-        shutil.rmtree(working_sort_directory)
+        _safe_rmtree(working_sort_directory)
     LOGGER.debug('here is percentile_list: %s', str(result_list))
     return result_list
+
+
+def _safe_rmtree(path):
+    """Attempts to remove a file path and logs a warning if it fails.
+
+    There is a possible bug in shutil.rmtree that can raise a FileNotFound
+    error even if the removal of a directory is possible and it removed it
+    see: https://bugs.python.org/issue29699. This function is effectively a
+    guard against that bug and logs a warning if the file path could not
+    be removed.
+
+    Args:
+        path (str): represents a file path to remove
+
+    Returns:
+        None
+    """
+    try:
+        # check if the path exists first before removing so we know if we
+        # get a FileNotFoundError we can be confident it is related to
+        # Python issue29699.
+        if os.path.exists(path):
+            shutil.rmtree(path)
+    except FileNotFoundError:
+        # check if the dir path is not there to verify if there was an
+        # incorrect exception raised due to Python issue29699. If the path
+        # does not exist assume the directory was removed correctly.
+        if os.path.exists(path):
+            LOGGER.warn(f'unable to remove {path}')
+    except OSError:
+        LOGGER.exception(
+            f'Ignoring unexpected OSError when removing {path}.')
--- a/src/pygeoprocessing/geoprocessing_core.pyx
+++ b/src/pygeoprocessing/geoprocessing_core.pyx
@@ -947,33 +947,33 @@ def _raster_band_percentile_double(
     return result_list
 
 
-def _safe_rmtree(path):
-    """Attempts to remove a file path and logs a warning if it fails.
+def _safe_rmtree(dir_path):
+    """Attempt remove directory via shutil.rmtree and log warning on failure.
 
     There is a possible bug in shutil.rmtree that can raise a FileNotFound
     error even if the removal of a directory is possible and it removed it
     see: https://bugs.python.org/issue29699. This function is effectively a
-    guard against that bug and logs a warning if the file path could not
+    guard against that bug and logs a warning if dir_path could not
     be removed.
 
     Args:
-        path (str): represents a file path to remove
+        dir_path (str): path to directory to remove
 
     Returns:
         None
     """
     try:
-        # check if the path exists first before removing so we know if we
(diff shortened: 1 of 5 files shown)


src/pygeoprocessing/geoprocessing_core.pyx
```diff
@@ -941,6 +942,38 @@ def _raster_band_percentile_double(
...
+    """Attempts to remove a file path and logs a warning if it fails.
+
+    There is a possible bug in shutil.rmtree that can raise a FileNotFound
+    error even if the removal of a directory is possible and it removed it
+    see: https://bugs.python.org/issue29699. This function is effectively a
+    guard against that bug and logs a warning if the file path could not
+    be removed.
+
+    Args:
+        path (str): represents a file path to remove
```
dcdenu4: `shutil.rmtree` expects `path` to be a directory. The argument description should reflect that OR if we want this to be a _general_ remove, we should probably call it something other than `rmtree`.

src/pygeoprocessing/geoprocessing_core.pyx
```diff
@@ -941,6 +942,38 @@ def _raster_band_percentile_double(
...
+
+    Returns:
+        None
+    """
+    try:
+        # check if the path exists first before removing so we know if we
+        # get a FileNotFoundError we can be confident it is related to
+        # Python issue29699.
+        if os.path.exists(path):
```
dcdenu4: I wonder if this should be `os.path.isdir` since we're really trying to delete a directory structure.

dcdenu4: Although, I guess `shutil.rmtree` will raise an error or something if it's not... So I take back what I said above.

src/pygeoprocessing/geoprocessing_core.pyx
```diff
@@ -941,6 +942,38 @@ def _raster_band_percentile_double(
...
+            shutil.rmtree(path)
+    except FileNotFoundError:
+        # check if the dir path is not there to verify if there was an
+        # incorrect exception raised due to Python issue29699. If the path
+        # does not exist assume the directory was removed correctly.
+        if os.path.exists(path):
+            LOGGER.warn(f'unable to remove {path}')
```
dcdenu4: `warn` is deprecated and should now be `warning`!

src/pygeoprocessing/geoprocessing_core.pyx
```diff
@@ -941,6 +942,38 @@ def _raster_band_percentile_double(
...
+    try:
+        # check if the path exists first before removing so we know if we
+        # get a FileNotFoundError we can be confident it is related to
+        # Python issue29699.
+        if os.path.exists(path):
+            shutil.rmtree(path)
+    except FileNotFoundError:
+        # check if the dir path is not there to verify if there was an
+        # incorrect exception raised due to Python issue29699. If the path
+        # does not exist assume the directory was removed correctly.
```
dcdenu4: Thinking about the case where I saw this from a user the problem went like this in `convolve_2d`:

1) Call for removal of `mask_dir/mask_raster_path` with `os.remove`
2) Call `shutil.rmtree` on `mask_dir`
3) `FileNotFoundError` on `mask_dir/mask_raster_path`, NOT the `mask_dir` directory path itself.

So, with this implementation we're not trying to "resolve" the python bug by then trying to "re-delete" the directory `path`, but just logging that "hey, some rare Python bug occurred so that directory `path` might still be around". And continue on our merry way.

I'm just jotting this down to make sure I'm on the same page.

dcdenu4 (changes requested):
Hey @richpsharp, I have some small suggested changes and a few comments.

I also have a question of whether we should make an issue to remove this patch once Python has provided a fix or we get past supporting a version of Python with this fix. When looking at all the Python issues related to this problem it seemed pretty clear that they would fix this at some point and so I'm wondering if it makes sense keeping code around for a problem that doesn't exist in the future?

Here's the PR that has been Open for a year and a half: https://github.com/python/cpython/pull/14064

So, maybe we shouldn't hold our breath!

src/pygeoprocessing/geoprocessing_core.pyx
richpsharp: I changed the docstring to reflect directory paths in d309f81cad51f8f152ca4e80ec8a6d79161c32b0

src/pygeoprocessing/geoprocessing_core.pyx
richpsharp: Sure, I changed it to `os.path.isdir` in d309f81cad51f8f152ca4e80ec8a6d79161c32b0

src/pygeoprocessing/geoprocessing_core.pyx
richpsharp: Fixed in d309f81cad51f8f152ca4e80ec8a6d79161c32b0

src/pygeoprocessing/geoprocessing_core.pyx
richpsharp: OH! I get what this is now. I thought this was some sort of internal `rmtree` race condition. I should just skip that #1 and dump this whole thing.

richpsharp:
Closing this PR in lieu of just not doing `rmtree` in. Please review #131 instead.

PR #139, opened 2020-12-08
fixes #138 allowing for true/false setting, added a logger debug for …

### LINKED ISSUE #138: Allow for `convolve_2d` to process in nodata areas and not mask the result
It is useful to process convolve_2d in areas with nodata, but not include those areas in the raw convolve calculation, in cases that involve weighted distance interpolation such as dilation or noise filling. This would require `mask_nodata=True` and `normalize_kernel=True` to be set in a call, but currently this raises an exception claiming an invalid use of arguments.

richpsharp:
This PR allows the user to set `ignore_nodata_and_edges=True` and `mask_nodata=False`, describes in the docstring why this is useful, adds a "debug" logger message to help during development in case this setting is used accidentally, and added a test case to cover the functionality in the way it would be expected to be used.

### DIFF
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -2030,7 +2030,7 @@ def warp_raster(
         callback_data=[target_raster_path])
 
     if vector_mask_options:
-        # Make sure the raster creation options passed to ``mask_raster`` 
+        # Make sure the raster creation options passed to ``mask_raster``
         # reflect any metadata updates
         updated_raster_driver_creation_tuple = (
             raster_driver_creation_tuple[0], tuple(raster_creation_options))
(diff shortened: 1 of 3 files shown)


src/pygeoprocessing/geoprocessing.py
```diff
@@ -2455,19 +2455,31 @@ def convolve_2d(
...
             to ``True`` to avoid distortions signal values near edges for
             large integrating kernels.
+                It can be useful to set this value to ``True`` to fill
+            nodata holes through distance weighted averaging. In this case
+            ``mask_nodata`` must be set to ``False`` so the result does not
+            mask out these areas which are filled in. When using this
+            technique be careful of cases where the kernel does not extend
+            over any areas except nodata holes, in this case the resulting
+            values in these areas will be nonsensical numbers, perhaps
+            numerical infinity or NaNs.
```
dcdenu4, line 2470: This is great and helpful detail.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2455,19 +2455,31 @@ def convolve_2d(
...
             sum of the kernel.
         mask_nodata (boolean): If true, ``target_path`` raster's output is
             nodata where ``signal_path_band``'s pixels were nodata. Note that
             setting ``ignore_nodata_and_edges`` to ``True`` while setting
-            ``mask_nodata`` to False would be a nonsensical result and would
-            result in exposing the numerical noise where the nodata values were
-            ignored. An exception is thrown in this case.
+            ``mask_nodata`` to ``False`` can allow for a technique involving
+            distance weighted averaging to define areas that would otherwise
+            be nodata, however be careful in cases where the kernel does not
```
dcdenu4: Maybe change to: "... areas that would otherwise be nodata. However, be careful in cases where ..."

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2455,19 +2455,31 @@ def convolve_2d(
...
+            extend over any valid non-nodata area since the result can be
+            numerical infinity or NaNs. would be a nonsensical result and
+            would result in exposing the numerical noise where the nodata
+            values were ignored. An exception is thrown in this case.
```
dcdenu4: Is this last part left over?
>would be a nonsensical result and would result in exposing the numerical noise where the nodata values were ignored. An exception is thrown in this case.

dcdenu4 (changes requested):
Hey @richpsharp ! This looks good to me. Love the detail in the doc string and just had a comment on maybe some left over notes there.

src/pygeoprocessing/geoprocessing.py
richpsharp: Thanks!

src/pygeoprocessing/geoprocessing.py
richpsharp: Yeah, run on. Fixed in 6a727e32a4866d5a83ecdb6e1c78f25034bfb9c1

src/pygeoprocessing/geoprocessing.py
richpsharp: Oops, it sure was! Fixed in 6a727e32a4866d5a83ecdb6e1c78f25034bfb9c1

dcdenu4 (approved):
Looks good!

dcdenu4:
Assuming the rest of the checks pass and don't fail on a docstring update.

PR #140, opened 2021-01-05
Feature/geometric stream extraction #128

### LINKED ISSUE #128: Generate geometric representations of streams
This issue comes from prototyping a flood risk model that requires backfill from each of the junctions in a [Strahler stream](https://en.wikipedia.org/wiki/Strahler_number#River_networks) system. It will also be useful for automatic derivation of flow accumulation parameters by studying the statistical difference in drop depths from Strahler stream order 1s to 2s (this comes form https://www.researchgate.net/publication/2329568_Advances_in_the_Mapping_of_Flow_Networks_From_Digital_Elevation_Data)

richpsharp:
This PR adds new new functions to `pygeoprocessing.routing` and closes issue #128 

* `extract_strahler_streams_d8` - creates geometric line segments corresponding to streams in a D8 flow raster and automatically determines a reasonable flow accumulation value per river with a statistical test for expected drop distance between streams.
 * `calculate_subwatershed_boundary` - uses the stream layer from above to calculate subwatersheds using a perimeter tracing technique.

Typical output from `extract_strahler_streams_d8` may look like this
image

And the subwatershed calculation would look like this: 
image

This PR includes new code for both functions, supporting helper functions, a large full integration test, and a small history update.

Runtime seems reasonable with performance of the stream extraction taking approximately 2 minutes for a 6k+ X 5k+ raster and 10ish seconds for the subwatershed calculation.


HISTORY.rst
```diff
@@ -25,6 +25,13 @@ Unreleased Changes
 * Fixed issue in ``convolve_2d`` that could lead to a file removal race
   condition and raise a ``FileNotFoundException`` when ignoring nodata
   regions.
+* Added a D8 Strahler order stream extraction function at
+  ``pygeoprocessing.extract_strahler_streams_d8``. Creates segmented
+  geometric lines which correspond to streams on the landscape.
+* Added a D8 subwatershed calculation function at
+  ``pygeoprocessing.calculate_watershed_boundary``. Creates subwatersheds
```
phargogh: Wouldn't this be `calculate_subwatershed_boundary` rather than `calculate_watershed_boundary`?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -77,6 +88,9 @@ cdef int* NEIGHBOR_OFFSET_ARRAY = [
     1, 1  # 7
     ]
 
+cdef int *D8_XOFFSET = [1, 1, 0, -1, -1, -1, 0, 1]
+cdef int *D8_YOFFSET = [0, -1, -1, -1, 0, +1, +1, +1]
```
phargogh, line 81: I realize these and `NEIGHBOR_OFFSET_ARRAY` are basically the same thing just in ever-so-slightly different presentations, but would it be worth consolidating them or deriving one from the other to reduce duplication?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2960,3 +3006,1287 @@ def _is_raster_path_band_formatted(raster_path_band):
...
+        * "order" (int): an integer representing the stream order
+        * "river_id" (int): unique ID used by all stream segements that
+            connect to the same outlet.
+
+    Args:
+        flow_dir_d8_raster_path_band (tuple): a path/band representing the D8
+            flow direction raster.
+        flow_accum_raster_path_band (tuple): a path/band representing the D8
+            flow accumulation raster represented by
+            `flow_dir_d8_raster_path_band`.
```
phargogh: Double-backticks for RST?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2960,3 +3006,1287 @@ def _is_raster_path_band_formatted(raster_path_band):
...
+        float min_p_val=0.05,
+        osr_axis_mapping_strategy=DEFAULT_OSR_AXIS_MAPPING_STRATEGY):
+    """Extract Strahler order stream geometry from flow accumulation.
+
+    Creates a Strahler ordered stream vector containing line segments
+    representing each separate stream fragment. The final vector contains
+    at least the fields:
+        * "order" (int): an integer representing the stream order
+        * "river_id" (int): unique ID used by all stream segements that
+            connect to the same outlet.
```
phargogh: I can't find the RTD build for this PR to verify, I think we might need a blank line above the bulleted list for it to be interpreted as a list.

phargogh: There are a bunch of other fields that are also written to the vector, would it be worth documenting them here as well?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2960,3 +3006,1287 @@ def _is_raster_path_band_formatted(raster_path_band):
...
+        flow_dir_srs.ImportFromWkt(flow_dir_info['projection_wkt'])
+        flow_dir_srs.SetAxisMappingStrategy(osr_axis_mapping_strategy)
+    else:
+        flow_dir_srs = None
+    gpkg_driver = gdal.GetDriverByName('GPKG')
+
+    stream_vector = gpkg_driver.Create(
+        target_stream_vector_path, 0, 0, 0, gdal.GDT_Unknown)
+    stream_layer = stream_vector.CreateLayer(
+        'stream', flow_dir_srs, ogr.wkbLineString)
```
phargogh: With watershed delineation, Stacie ended up requesting that we add the file suffix to the layer name as well, since it's the layer name that is actually shown in the Arc and QGIS.  With watershed delineation, the solution we settled on was to [allow a user-defined layer name](https://github.com/natcap/pygeoprocessing/blob/main/src/pygeoprocessing/routing/watershed.pyx#L628).  Maybe this would be useful here as well?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2960,3 +3006,1287 @@ def _is_raster_path_band_formatted(raster_path_band):
...
+    cdef int x_n, y_n  # the _n is for "neighbor"
+    cdef int *d8_backflow = [
+        0, 0, 0, 0, 1, 0, 0, 0,
+        0, 0, 0, 0, 0, 1, 0, 0,
+        0, 0, 0, 0, 0, 0, 1, 0,
+        0, 0, 0, 0, 0, 0, 0, 1,
+        1, 0, 0, 0, 0, 0, 0, 0,
+        0, 1, 0, 0, 0, 0, 0, 0,
+        0, 0, 1, 0, 0, 0, 0, 0,
+        0, 0, 0, 1, 0, 0, 0, 0]
```
phargogh: Neat adjacency matrix!  I see this same matrix used in 3 places throughout the file ... would it be worth abstracting out to a module-level definition?

And on a related note, it also looks like we have a similar backflow array represented in `D8_REVERSE_DIRECTION ` at the module level.  I realize that their representations are slightly different, but would it be worth the effort to consolidate the approaches?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -409,9 +439,21 @@ cdef class _ManagedRaster:
...
+                    if n_attempts == 0:
+                        raise RuntimeError(
+                            f'could not open {self.raster_path} for writing')
+                    LOGGER.warning(
+                        f'opening {self.raster_path} resulted in null, '
+                        f'trying {n_attempts} more times.')
+                    n_attempts -= 1
+                    time.sleep(0.5)
+                raster_band = raster.GetRasterBand(self.band_id)
+                break
```
phargogh, line 445: Hey, great idea!  Was this something that same up strictly in the context of debugging this new routing functionality, or should it be replicated elsewhere where we have `ManagedRaster` classes defined?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2960,3 +3006,1298 @@ def _is_raster_path_band_formatted(raster_path_band):
...
+    cdef int y_n = y_l + D8_YOFFSET[direction_to_test]
+    if x_n < 0 or y_n < 0 or x_n >= n_cols or y_n >= n_rows:
+        return 0
+    cdef long point_discovery = <long>discovery_managed_raster.get(x_n, y_n)
+    return (point_discovery != discovery_nodata and
+            point_discovery >= discovery and
+            point_discovery <= finish)
+
+
+def _calculate_stream_geometry(
```
phargogh: Would it be worth `cdef`ing this function?  Most of the parameter types are already defined, so I wonder if that was the original intent here.  Based on where it's being used in this file, I'd guess it wouldn't be a critical speedup.

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2960,3 +3006,1298 @@ def _is_raster_path_band_formatted(raster_path_band):
...
+    """
+    cdef int *d8_backflow = [
+        0, 0, 0, 0, 1, 0, 0, 0,
+        0, 0, 0, 0, 0, 1, 0, 0,
+        0, 0, 0, 0, 0, 0, 1, 0,
+        0, 0, 0, 0, 0, 0, 0, 1,
+        1, 0, 0, 0, 0, 0, 0, 0,
+        0, 1, 0, 0, 0, 0, 0, 0,
+        0, 0, 1, 0, 0, 0, 0, 0,
+        0, 0, 0, 1, 0, 0, 0, 0]
```
phargogh: Here's another case of the D8 backflow adjacency matrix that could be consolidated to the top of the module.

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2960,3 +3006,1298 @@ def _is_raster_path_band_formatted(raster_path_band):
...
+        flow_accum_threshold (int): minimum flow accumulation value to define
+            string.
+        coord_to_stream_ids (dict): map raster space coordinate tuple to
+            a list of stream ids
+
+    Returns:
+        Georeferenced linestring connecting x/y to upper point where upper
+            point's threshold is the last point where its flow accum value
+            is >= `flow_accum_threshold`. Returns None if x/y is below
+            flow accum threshold.
```
phargogh: Looks like the return value has changed since this returns item was written!  Could be worth updating.

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2960,3 +3006,1298 @@ def _is_raster_path_band_formatted(raster_path_band):
...
+        0, 0, 1, 0, 0, 0, 0, 0,
+        0, 0, 0, 1, 0, 0, 0, 0]
+    cdef int x_n, y_n, d, d_n, stream_end=0
+
+    if flow_accum_managed_raster.get(x_l, y_l) < flow_accum_threshold:
+        return None
+    upstream_id_list = []
+    # anchor the line at the downstream end
+    stream_line = ogr.Geometry(ogr.wkbLineString)
+    x_p, y_p = gdal.ApplyGeoTransform(geotransform, x_l+0.5, y_l+0.5)
```
phargogh, line 4220: TIL about `gdal.ApplyGeoTransform`.  Neat!  I bet there are more than a few places where we could be using this function in InVEST and pygeoprocessing.

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2960,3 +3006,1298 @@ def _is_raster_path_band_formatted(raster_path_band):
...
+    """
+    cdef int *d8_backflow = [
+        0, 0, 0, 0, 1, 0, 0, 0,
+        0, 0, 0, 0, 0, 1, 0, 0,
+        0, 0, 0, 0, 0, 0, 1, 0,
+        0, 0, 0, 0, 0, 0, 0, 1,
+        1, 0, 0, 0, 0, 0, 0, 0,
+        0, 1, 0, 0, 0, 0, 0, 0,
+        0, 0, 1, 0, 0, 0, 0, 0,
+        0, 0, 0, 1, 0, 0, 0, 0]
```
phargogh: And here's another backflow adjacency matrix.

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2960,3 +3006,1298 @@ def _is_raster_path_band_formatted(raster_path_band):
...
+            search walk, used to determine watershed containment.
+        d8_flow_dir_raster_path_band (tuple): raster/path band for d8 flow dir
+            raster
+        strahler_stream_vector_path (str): path to stream segment vector
+        target_watershed_boundary_vector_path (str): path to created vector
+            of stringline for watershed boundaries.
+
+    Returns:
+        linestring of projected points indicating watershed boundary.
+    """
```
phargogh: This docstring looks to be a bit out of date relative to the function signature and what the function appears to be returning.

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2960,3 +3006,1298 @@ def _is_raster_path_band_formatted(raster_path_band):
...
+    workspace_dir = tempfile.mkdtemp(
+        dir=os.path.join(
+            os.path.dirname(target_watershed_boundary_vector_path)))
```
phargogh: Could I recommend using one of the prefix/suffix naming arguments to clarify what the tmpdir is being used for?  I've personally found it useful when there are a bunch of temporary directories lying around.

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2960,3 +3006,1298 @@ def _is_raster_path_band_formatted(raster_path_band):
...
+                    discovery_managed_raster, discovery_nodata):
+                edge_side = (edge_side-2) % 8
+                edge_dir = (edge_dir-2) % 8
+                x_l += D8_XOFFSET[edge_dir]
+                y_l += D8_YOFFSET[edge_dir]
+                # note the pixel moved
+                boundary_list.append((x_l, y_l))
+
+        # use to keep track of how many turns around a pixel, used if
+        # a degerate
```
phargogh: "Degenerate"?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2960,3 +3006,1298 @@ def _is_raster_path_band_formatted(raster_path_band):
...
+        n_left_turns = 0
+        while True:
+            # step the edge then determine the projected coordinates
+            x_f += D8_XOFFSET[edge_dir]
+            y_f += D8_YOFFSET[edge_dir]
+            # equivalent to gdal.ApplyGeoTransform(geotransform, x_f, y_f):
```
phargogh: Are the `gdal.ApplyGeoTransform` equivalents below just to eliminate the python function call overhead?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2960,3 +3006,1298 @@ def _is_raster_path_band_formatted(raster_path_band):
...
+            x_p = g0 + g1*x_f + g2*y_f
+            y_p = g3 + g4*x_f + g5*y_f
+            watershed_boundary.AddPoint(x_p, y_p)
+
+            if n_left_turns == 4:
+                LOGGER.info(
+                    '4 left turns around the same pixel so terminating')
```
phargogh: Would it be useful to log some other information about this feature, such as the FID of the watershed?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2960,3 +3006,1298 @@ def _is_raster_path_band_formatted(raster_path_band):
...
+        discovery/finish (long): the discovery and finish time that defines
+            whether a pixel discovery time is inside a watershed or not.
+        discovery_managed_raster (_ManagedRaster): discovery time raster
+            x/y gives the discovery time for that pixel.
+        discovery_nodata (long): nodata value for discovery raster
+        boundary_list (list): this list is appended to for new pixels that
+            should be neighbors in the fill.
+
+    Return:
+        None.
```
phargogh, line 4114: There's no return value, but there is a side effect.  This is a CDEF'd function so probably no one will actually see this unless they're reading the source, but should the side effect be recorded here?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2960,3 +3006,1298 @@ def _is_raster_path_band_formatted(raster_path_band):
...
+        # doesn't interfere with the loop return to think the cells are no
+        # longer in the watershed
+        for boundary_x, boundary_y in boundary_list:
+            discovery_managed_raster.set(boundary_x, boundary_y, -1)
+    watershed_layer.CommitTransaction()
+    watershed_layer = None
+    watershed_vector = None
+    discovery_managed_raster.close()
+    finish_managed_raster.close()
+    #shutil.rmtree(workspace_dir)
```
phargogh: Lingering deprecated code?

phargogh (changes requested):
Thanks, Rich!  I had a few comments, mostly pretty minor, and the algorithms themselves look really interesting.  Let me know what you think!

HISTORY.rst
richpsharp: It should be! And I just noticed i had it in the wrong part of the history file -- 2.1.2 release, not 'unreleased'. Fixed both in eaedf452dbec0be690ca17d88e81af7928429339

src/pygeoprocessing/routing/routing.pyx
richpsharp: Yep I like it. Done in f1812913d3416a5b8269d5e2cdabeec841342b47

src/pygeoprocessing/routing/routing.pyx
richpsharp: Fixed in 0d03242574f2801b124cab5a526dec4897aa8c2f

src/pygeoprocessing/routing/routing.pyx
richpsharp: Fixed in 0d0081f

richpsharp: Weeeelll... I did say "at least the following fields" because in a way they're useful for something else, but I guess that should be documented if it really is useful huh?

richpsharp: Okay i documented all the features in 0903796a7b389b1123f4176f62774c1ca9f82f30 except "d8_flow_dir" which was just an intermediate field anyway that I delete now before the function is over.

src/pygeoprocessing/routing/routing.pyx
richpsharp: hilarious and also unnecessary since that bit of code wasn't needed at all. i removed the command and all the  "n_turns" bits too in 28fbda7

src/pygeoprocessing/routing/routing.pyx
richpsharp: Yes and i realize there are other python calls in there, but they're so few i figure if i can cut one out of 4 that's 1/4 speedup? i added a comment in there about that too in 361d9e1

src/pygeoprocessing/routing/routing.pyx
richpsharp: That line is no longer in there because i no longer explicitly check for 4 turns as in 28fbda7 above

src/pygeoprocessing/routing/routing.pyx
richpsharp: I think I do document that side effect in the 'args' as "boundary_list (list): this list is appended to for new pixels that should be neighbors in the fill." Is that okay?

src/pygeoprocessing/routing/routing.pyx
richpsharp: So weird, that line is not in my working file, but otherwise yes, I would have removed it.

richpsharp:
Hi @phargogh ! I just addressed all your comments which were all great. My favorite was consolidating those two different ways to determine x/y offsets for neighbors.

src/pygeoprocessing/routing/routing.pyx
richpsharp: oh sure, in fact i think I've always just used the base filename which is what i'd prefer

richpsharp: done in c46e5741c1948a29fa0da521f6dafb35471ad4fe

src/pygeoprocessing/routing/routing.pyx
richpsharp: It came up when testing the subwatershed routing... that's not to say i wouldn't want it everywhere else but it was specifically happening during this development. What I'd *really* like is to avoid duplicating this class in every Cython module, but every time I try to abstract this into a Cython class things start to break... I think bevcause of the linkage to the LRUCache.h file. I think we've both attempted it over the years. Can I punt on duplicating this code everywhere else and if it's an issue in the future we try really hard to get a ManagedRaster abstract component we can use in all the pygeoprocessing cython modules?

src/pygeoprocessing/routing/routing.pyx
richpsharp: Sure! changed in f97cfcd. I think I left it as a def because it returns a geometry object so there's overhead on that but it doesn't hurt anything. Mostly this is in a function because it's a complicated bit of code I needed to use a few different places and since we were already in Cython I static typed everything that made sense to.

src/pygeoprocessing/routing/routing.pyx
richpsharp: Fixed in previous commit

src/pygeoprocessing/routing/routing.pyx
richpsharp: Update the return docstring in 8d6eea1

src/pygeoprocessing/routing/routing.pyx
richpsharp: I use it all the time. Also useful in combination with `gdal.InvGeoTransform` if you want to turn projected coordinates into raster coordinates.

src/pygeoprocessing/routing/routing.pyx
richpsharp: fixed fixed!

src/pygeoprocessing/routing/routing.pyx
richpsharp: Fixed in e057a8b3d95b96091a3189c4b41d1c8fae8399ba

src/pygeoprocessing/routing/routing.pyx
richpsharp: Sure, added in e1afc69

richpsharp:
Okay @phargogh , I know earlier I said I addressed all your comments but github had compressed the conversation and I didn't notice at first that 10 more of your comments were hidden. So I did one more pass and you'll notice one other change I made was to ADD a `__swig_destroy__` back into pygeoprocessing.rasterize. For some reason a test case started to fail where I cause the callback handler to trigger a divide by zero exception, but then the test case couldn't clean up the working directory because the raster was still in use. I tried a LOT of things and the only thing that worked in the end was a specific `__swig_destroy__` call on the raster when that particular exception is triggered. So sorry that's tossed in here too.

src/pygeoprocessing/routing/routing.pyx
richpsharp: Ugh you're right, `D8_REVERSE_DIRECTION` is superior and I dumped d8_backflow in lieu of that in fb837a1.

src/pygeoprocessing/routing/routing.pyx
phargogh: Oh yeah, that's even better.  Thanks!

One thing I noticed, though is that c46e5741c1948a29fa0da521f6dafb35471ad4fe also includes the new function from #148, which doesn't look intentional!

src/pygeoprocessing/routing/routing.pyx
phargogh: Oh for sure!  We've definitely both tried to abstract this out, but it became clear that it's going to take a C++ extension to really do it right with templating.  So yes, let's punt.

src/pygeoprocessing/routing/routing.pyx
phargogh: Nice!  I like the RST formatting, as well :)

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2948,6 +2983,114 @@ def extract_streams_mfd(
     LOGGER.info('100.0% complete')
 
 
+def detect_outlets(d8_flow_dir_raster_path_band, target_outlet_vector_path):
```
phargogh: I think this somehow made its way into this PR, though it's also under review in #148 ... I think that wasn't the intent?

phargogh (changes requested):
Thanks, Rich!  I just had one last comment about a function being reviewed elsewhere that seems to have made its way into this file.  But other than that, I think this is looking good.

src/pygeoprocessing/routing/routing.pyx
richpsharp: Sure wasn't the intent and honestly not sure how it got in there. And double honestly I'm still not comfortable with git's branching, I can't seem to unroll it/detect what happened as easily as mercurial. Anyway, I've removed `detect_outlets` in 593996ac11670eb709a1327fe98bf961e4369f7f.

richpsharp:
> Thanks, Rich! I just had one last comment about a function being reviewed elsewhere that seems to have made its way into this file. But other than that, I think this is looking good.

thanks @phargogh, I removed the function, pushed, tests pass etc.

phargogh (approved):
Merging!

PR #142, opened 2021-01-05
Fixes issue #141 

### LINKED ISSUE #141: Test failure/race condition when live logging tests
When live logging tests with this command:

```
pytest -xv --log-cli-level=DEBUG tests/test_geoprocessing.py::PyGeoprocessing10::test_zonal_statistics
```

The test fails because it appears the aggregate test vector has not been dereferenced. Likely this is because of a missing `=None` in the call chain execution.

richpsharp:
This PR fixes a race issue when releasing `zonal_statistics`'s aggregate vector object that occured when `pytest` ran with live logging -- essentially a race condition when deleting the test workspace while the aggregate vector still had an open reference somewhere.

This PR addresses all the possible places that a leaked `ogr` object could exist that was not explicitly set to `None` before the function terminates.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -12,7 +12,7 @@ Unreleased Changes
 * Fixed issue in zonal statistics that could cause the aggregate vector to be
   locked due to a dereferencing race condition. This was present in some
   cases with a flaky unit test but could have been seen in practice if the
-  vector was deleted immediately after the call ton ``zonal_statistics``.
+  vector was deleted immediately after the call to ``zonal_statistics``.
 
 2.1.2 (2020-12-03)
 ------------------
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -9,6 +9,10 @@ Unreleased Changes
   in missing datasets. Additionally added a logger ``debug`` message to note
   this "unusual" setting of these parameters in case of accidental usage
   which could be noted during development.
+* Fixed issue in zonal statistics that could cause the aggregate vector to be
+  locked due to a dereferencing race condition. This was present in some
+  cases with a flaky unit test but could have been seen in practice if the
+  vector was deleted immediately after the call ton ``zonal_statistics``.
 
 2.1.2 (2020-12-03)
 ------------------
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -1181,7 +1181,6 @@ def zonal_statistics(
     if aggregate_vector is None:
         raise RuntimeError(
             "Could not open aggregate vector at %s" % aggregate_vector_path)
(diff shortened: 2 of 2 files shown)


HISTORY.rst
```diff
@@ -9,6 +9,10 @@ Unreleased Changes
   in missing datasets. Additionally added a logger ``debug`` message to note
   this "unusual" setting of these parameters in case of accidental usage
   which could be noted during development.
+* Fixed issue in zonal statistics that could cause the aggregate vector to be
+  locked due to a dereferencing race condition. This was present in some
+  cases with a flaky unit test but could have been seen in practice if the
+  vector was deleted immediately after the call ton ``zonal_statistics``.
```
dcdenu4: Little typo: `call ton`

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1205,10 +1204,10 @@ def zonal_statistics(
...
     # -1 here because bands are 1 indexed
     raster_nodata = raster_info['nodata'][base_raster_path_band[1]-1]
-    with tempfile.NamedTemporaryFile(
-            prefix='clipped_raster', suffix='.tif', delete=False,
-            dir=working_dir) as clipped_raster_file:
-        clipped_raster_path = clipped_raster_file.name
+    temp_working_dir = tempfile.mkdtemp(dir=working_dir)
+    clipped_raster_path = os.path.join(
+        temp_working_dir, 'clipped_raster.tif')
+
```
dcdenu4, line 1210: This is certainly easier to read!

dcdenu4 (commented):
Just a real minor comment on a typo in History. Interesting that a geometry or feature reference would keep a "file object open", even when setting the layer and vector objects to None at the end of the function.

HISTORY.rst
richpsharp: fixed in https://github.com/natcap/pygeoprocessing/pull/142/commits/4e46f37a52b3e591dd7389aa7cbede096c1cce8f

dcdenu4 (approved):

richpsharp:
> Just a real minor comment on a typo in History. Interesting that a geometry or feature reference would keep a "file object open", even when setting the layer and vector objects to None at the end of the function.

I removed the `n`! And I think what's happening there is that it's a chain of references that needs to get garbage collected, which it really does do, but under certain circumstances one thing might get ahead by 1ms. What's really interesting to me is that in Python it allows that, but also if you set an object to `None` it immediately garbage collects it so you don't have to sweat it.

So you're right, I think the reference to the geometry was what was keeping the feature open which was keeping the layer then the vector open. Maybe that chain was a little too long for it to beat. Lesson learned and I like making that explicit anyway.

PR #143, opened 2021-01-07
Bugfix/pitfill small delta error 130

### LINKED ISSUE #130: Fix bug in pitfilling when DEM has small increments
I experienced this issue when running tests on a DEM with delta height < 1e-5 units between pixels. I'm guessing from the results that some pixels were determined to be at the same height because of their floating point approximations were very close. For example "346.2117919921875" vs "346.2773132324219". I'm also guessing these values came out of a bilinear interpolation in an attempt to make a DEM higher resolution. If that's the case this could be an expected value and should be handled properly.

richpsharp:
Fixes an issue that was causing DEMs with very small delta changes between pixels to not fill a pit when one existed. This was caused by an aggressive use of `is_close` to determine if two pixels were at the same height. Instead this is a rare case when a floating point value shoudl be considered equal only when it is exactly equal.

I also changed the notation of `is_close` to `_is_close` to show it's an internal function to this module.

Also added a test that would have failed before this fix and an update to history.

Closes #130


richpsharp:
oops wrong branch

PR #144, opened 2021-01-07
fixes #130 -- small delta in DEM causes pit fill to not fill pit

### LINKED ISSUE #130: Fix bug in pitfilling when DEM has small increments

richpsharp:
Fixes an issue that was causing DEMs with very small delta changes between pixels to not fill a pit when one existed. This was caused by an aggressive use of `is_close` to determine if two pixels were at the same height. Instead this is a rare case when a floating point value shoudl be considered equal only when it is exactly equal.

I also changed the notation of `is_close` to `_is_close` to show it's an internal function to this module.

Also added a test that would have failed before this fix and an update to history.

Closes #130

### DIFF
--- a/tests/test_routing.py
+++ b/tests/test_routing.py
@@ -57,7 +57,6 @@ class TestRouting(unittest.TestCase):
         pygeoprocessing.routing.fill_pits(
             (base_path, 1), fill_path, working_dir=self.workspace_dir)
         result_array = pygeoprocessing.raster_to_numpy_array(fill_path)
-        numpy.set_printoptions(suppress=True)
         self.assertTrue(
             (result_array == expected_result).all(),
             result_array == expected_result)
--- a/tests/test_routing.py
+++ b/tests/test_routing.py
@@ -40,6 +40,28 @@ class TestRouting(unittest.TestCase):
         dem_array[3:8, 3:8] = 0.0
         numpy.testing.assert_almost_equal(result_array, dem_array)
 
+    def test_pit_filling_small_delta(self):
+        """PGP.routing: test pitfilling on small delta."""
+        base_path = os.path.join(self.workspace_dir, 'base.tif')
+        dem_array = numpy.empty((4, 4), dtype=numpy.float32)
+        # these values came from a real world dem that failed
+        lower_val = 272.53228759765625
+        higher_val = 272.5325012207031
+        dem_array[:] = higher_val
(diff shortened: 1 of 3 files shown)


tests/test_routing.py
```diff
@@ -40,6 +40,28 @@ def test_pit_filling(self):
...
+        dem_array[2, 2] = lower_val
+
+        expected_result = numpy.empty((4, 4), numpy.float32)
+        expected_result[:] = higher_val
+        _array_to_raster(dem_array, None, base_path)
+        fill_path = os.path.join(self.workspace_dir, 'filled.tif')
+        pygeoprocessing.routing.fill_pits(
+            (base_path, 1), fill_path, working_dir=self.workspace_dir)
+        result_array = pygeoprocessing.raster_to_numpy_array(fill_path)
+        numpy.set_printoptions(suppress=True)
```
davemfish: I wonder if this was just meant for debugging? If it's useful to keep these print options, maybe it's better to use as a context manager to limit their effects?

tests/test_routing.py
```diff
@@ -40,6 +40,28 @@ def test_pit_filling(self):
...
+        self.assertTrue(
+            (result_array == expected_result).all(),
+            result_array == expected_result)
+
```
davemfish, line 63: This looked a little strange at first with two args inside `assertTrue`. Then I tried it out and see how it makes for a very useful failure report!

davemfish (commented):
Hey @richpsharp , changes look good to me, there's just one line left in here that I thought may have just been for debugging.

tests/test_routing.py
richpsharp: it was meant for debugging and it didn't even work! removed in 1650bc6b7977c7ef1d48d1a239cb17d7946f6229

tests/test_routing.py
richpsharp: ha glad you like it! as you might guess from your previous comment i'd tried a few things.

davemfish (approved):

PR #147, opened 2021-01-14
Bugfix/stats worker core dump 146

### LINKED ISSUE #146: Fix race condition/core dump on Python 3.8 shared memory raster stats
On many parallel process runs of raster calculator i have occasionally seen a core dump that's originating somewhere in `stats_worker`. Suspiciously that's the only function in pygeoprocessing that uses shared memory and I suspect there's something happening that's causing memory to be incorrectly deallocated. This is tricky to debug and reproduce but it happens when running 64 parallel processes driven by taskgraph on tens of thousands of tasks. After a while a process will core dump and digging through the core file shows corrupted memory when inside this worker.... Fix that.

richpsharp:
This PR fixes a bug I discovered when examining the results of a ``core`` file that dumped when running pygeoprocessing in a multiprocessing enviornment with 64 processes and thousands of tasks. Occasionally some of these processes would terminate and the core file suggested it was a memory corruption inside of ``stats_worker``. That function uses shared memory objects and also is one of the remaining ``__swig_destroy__`` users in pygeoprocessing. I've made shared memory an optional flag and removed all the ``__swig...`` calls throughout PyGeoprocessing.

It's difficult to reproduce this so I don't have a way to add a test that wouldn't take significant cpu resources and runtime. Instead I've modified some tests to inclucde the ``use_shared_memory`` flag.

Closes #146

### DIFF
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -86,7 +86,8 @@ DEFAULT_OSR_AXIS_MAPPING_STRATEGY = osr.OAMS_TRADITIONAL_GIS_ORDER
 def raster_calculator(
         base_raster_path_band_const_list, local_op, target_raster_path,
         datatype_target, nodata_target,
-        calc_raster_stats=True, largest_block=_LARGEST_ITERBLOCK,
+        calc_raster_stats=True, use_shared_memory=False,
+        largest_block=_LARGEST_ITERBLOCK, max_timeout=_MAX_TIMEOUT,
         raster_driver_creation_tuple=DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS):
     """Apply local a raster operation on a stack of rasters.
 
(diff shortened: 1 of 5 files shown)


src/pygeoprocessing/geoprocessing.py
```diff
@@ -455,7 +460,7 @@ def raster_calculator(
                     target_block = target_block[target_block != nodata_target]
                 target_block = target_block.astype(numpy.float64).flatten()
 
-                if sys.version_info >= (3, 8):
+                if sys.version_info >= (3, 8) and use_shared_memory:
```
dcdenu4, line 465: Ah, yea I thought this was a Python 3.8 thing. Should it be mentioned in the API docs or HISTORY that this is for Python 3.8+ users? Speaking of which, Pyinstaller just released 4.1 which supports Python 3.8 and 3.9. So InVEST should officially support at least Python 3.8 very soon.

dcdenu4 (commented):
Hey @richpsharp looks good to me. I had one comment about whether the use of `shared_memory` should mention that it's only Python 3.8+ compatible? Should I as a user be aware of that?

src/pygeoprocessing/geoprocessing.py
richpsharp: Sure, added in both in 107cbfa0bbb0cb3ed4dabd7ffda6e611a2887f35

richpsharp:
> Hey @richpsharp looks good to me. I had one comment about whether the use of `shared_memory` should mention that it's only Python 3.8+ compatible? Should I as a user be aware of that?

I don't think the end user would have been affected in a way that was measurable but I think mentioning it in the API docs and history is a good idea. I did both.

dcdenu4 (approved):
Looks good!

PR #151, opened 2021-01-21
Feature/outlet maker 148

### LINKED ISSUE #148: Determine outlets of raster
It's often useful to see where the "outlets" are in a DEM. These are defined as internal sinks, drains to nodata, or drains to the edge of a raster. This test requires per-pixel testing and is only efficient without Pythonic loop overhead. Make a function that takes a D8 raster and creates a point vector output showing the outlets with additional annotations of the i/j coordinates in the dem.

richpsharp:
Creates a `pygeoprocessing.routing.detect_outlets` function that creates a vector with points placed on the center of outlet pixel on 
 a D8 flow direction raster. The vector has additional metadata about the (i, j) locations of the pixel in raster coordinate space which can be useful when turning the vector into a raster or using as some other indexing mechanism at the raster level.

This PR adds that function, has a unit test to test all the boundary conditions of an outlet detection, and makes an update to `HISTORY.rst`.

Closes #148 

p.s. there's one other change where i moved the call to open a vector for rasterization to after the section where the code first checks to see if there's valid input. That's the `vector = gdal.OpenEx(vector_path, gdal.OF_VECTOR)` line you'll see modified in `geoprocessing.py`. It functionally doesn't change anything but conceptually it's a nice idea to not allocate resources before checking to see if the state of the system is valid.


src/pygeoprocessing/routing/routing.pyx
```diff
@@ -4076,6 +4076,111 @@ def calculate_subwatershed_boundary(
...
+                    outlet_point = ogr.Geometry(ogr.wkbPoint)
+                    proj_x, proj_y = gdal.ApplyGeoTransform(
+                        raster_info['geotransform'],
+                        xi_root+0.5, yi_root+0.5)
+                    outlet_point.AddPoint(proj_x, proj_y)
+
+                    outlet_feature = ogr.Feature(outlet_layer.GetLayerDefn())
+                    outlet_feature.SetGeometry(outlet_point)
+                    outlet_feature.SetField('i', xi_root)
+                    outlet_feature.SetField('j', yi_root)
```
emlys, line 4186: I'm wondering about including `i` and `j` as the original raster coordinates, what would be the use case for that? It seems potentially confusing to give each point an extra pair of coordinates. And without additional metadata like the pixel size and original spatial reference, would `i` and `j` lose their meaning?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -4076,6 +4076,111 @@ def calculate_subwatershed_boundary(
...
+                    outlet_layer.CreateFeature(outlet_feature)
```
emlys, line 4189: should there also be a unique id field? Remembering a comment in delineateit that I didn't write: _"It's important to have a user-facing unique ID field for post-processing (e.g. table-joins) that is not the FID. FIDs are not stable across file conversions that users might do."_

emlys (commented):
Looks good! Had a couple of questions. My only other suggestion would be to add a few comments pointing out the key parts of the algorithm.

src/pygeoprocessing/routing/routing.pyx
richpsharp: I mentioned it in the docstring and added a comment in c134127e48167e25e946523b4604cd9e05a50319 explaining this.. these aren't the projected coordinates but the raster i,j coordinates. It's useful for when you're using this datastructure to seed an upstream search from a pixel coordinate rather than taking the step of having the projected coordinate and then inverse geotransforming it back into raster space. All the hydrological algorithms we have operate at the raster/pixel scale so that's just a handy way of getting it back down there.

A very concrete example of this is where I just used it where I needed to calculate the number of beneficiaries "downstream" from a point. You can do this by using a "distance to channel" function with a weighted raster that is the beneficiary. However you need to also provide where your channels are. Since I actually wanted it to go to the outlet I use this data structure to make a raster that only has the pixels set to 1 where these points are.

I see what you mean that it looks extra and could be derived anyway, but the use case for this is in the raster space not the projected space. In that context the extra information is actually the projected coordinates, but it's nice to have those because then you can just plink it down in qgis and everything matches up.

How does that sound?

src/pygeoprocessing/routing/routing.pyx
richpsharp: Sure, added in c134127e48167e25e946523b4604cd9e05a50319 and added a test for it in 7c6d48beb86f5530211cf739b82866fea8baa37b.

richpsharp:
> Looks good! Had a couple of questions. My only other suggestion would be to add a few comments pointing out the key parts of the algorithm.

Sure, I added some more comments in 7c6d48beb86f5530211cf739b82866fea8baa37b. Do you think it's okay now?

richpsharp:
> Looks good! Had a couple of questions. My only other suggestion would be to add a few comments pointing out the key parts of the algorithm.

Thanks @emlys, I addressed all your comments and requested a re-review.

src/pygeoprocessing/routing/routing.pyx
emlys: Makes sense! thanks for the explanation

emlys:
Thanks @richpsharp !

emlys (approved):

src/pygeoprocessing/routing/routing.pyx
davemfish: > should there also be a unique id field? Remembering a comment in delineateit that I didn't write: _"It's important to have a user-facing unique ID field for post-processing (e.g. table-joins) that is not the FID. FIDs are not stable across file conversions that users might do."_

Haha, I wrote that comment in Coastal Vulnerability after Jess converted a GPKG to a SHP because Arc didn't have full-fledged GPKG support yet. And then her table-joins were off by one.

PR #153, opened 2021-01-25
Feature/speed up outlet detection #152

### LINKED ISSUE #152: Integrate InVEST's pour point detection with pygeoprocessing's?
With PR #151 merged, we now have two implementations of pour point detection:

* One in [`pygeoprocessing.routing`](https://github.com/natcap/pygeoprocessing/blob/main/src/pygeoprocessing/routing/routing.pyx#L4079)
* And one in [`natcap.invest.delineateit`](https://github.com/natcap/invest/blob/846fef97762ed23b1c3053d8b96b390c8dbb962f/src/natcap/invest/delineateit/delineateit.py#L634)

When comparing the two implementations, I noticed that @emlys 's implementation in DelineateIt runs in 1/3 the time of the implementation in pygeoprocessing that was recently merged.  [Here's a gist](https://gist.github.com/phargogh/d8ee8935ac1a5ead507cee601720d45b) comparing the two DEMs (and let me know if you'd like the Colombia DEM to try this out on).  Here are the mean runtimes (from n=5 runs) for two rasters running each of these  implementations:

```
Colombia DEM (15030 x 20631 pixels)
delineateit mean: 10.326407289505005s
pygeoprocessing mean: 31.769051218032835s

Gura DEM (1939 x 603 pixels)
delineateit mean: 0.031762552261352536s
pygeoprocessing mean: 0.09338397979736328s
```

I see that #151 has already been merged, but this difference in timing makes me wonder where the speedup is and if there's a key part of the InVEST implementation that can be integrated to grant the same speedup.

richpsharp:
This PR fixes issue #152 which noted the previous implementation of `detect_outlets` had a 3-fold slower runtime than the one in `routedem`. This PR improves runtime of `detect_outlets` by at least 10-fold. Timings below:

Runtimes for a 28525x36560 raster (DEM_filled_LdM.tif) are below:

Previous implementation:
`[87.3490002155304, 83.47498416900635, 87.20200085639954, 90.5391161441803, 94.48638939857483]`

This implementation:
`[7.877000093460083, 7.831982612609863, 7.840015411376953, 7.8379998207092285, 7.906984806060791]`

Worst case speedup:
83.4 / 7.9 = 10.5

Closes #152

### DIFF
--- a/src/pygeoprocessing/routing/routing.pyx
+++ b/src/pygeoprocessing/routing/routing.pyx
@@ -4081,7 +4081,8 @@ def detect_outlets(d8_flow_dir_raster_path_band, target_outlet_vector_path):
 
     Args:
         d8_flow_dir_raster_path_band (tuple): raster path/band tuple
-            indicating D8 flow direction.
+            of type uint8  indicating D8 flow direction created by
+            `routing.flow_dir_d8`.
         target_outlet_vector_path (str): path to a vector that is created
             by this call that will be in the same projection units as the
             raster and have a point feature in the center of each pixel that
@@ -4094,9 +4095,9 @@ def detect_outlets(d8_flow_dir_raster_path_band, target_outlet_vector_path):
     Return:
         None.
     """
-    cdef int flow_dir, flow_dir_n
     cdef int xoff, yoff, win_xsize, win_ysize, xi, yi
     cdef int xi_root, yi_root, raster_x_size, raster_y_size
+    cdef int flow_dir, flow_dir_n
     cdef int next_id=0
     cdef char x_off_border, y_off_border, win_xsize_border, win_ysize_border
 
@@ -4134,7 +4135,7 @@ def detect_outlets(d8_flow_dir_raster_path_band, target_outlet_vector_path):
         os.path.splitext(target_outlet_vector_path)[0])
     outlet_layer = outlet_vector.CreateLayer(
         outet_basename, raster_srs, ogr.wkbPoint)
-    # i and j indicate the coordinates of the point in raster space wereas
+    # i and j indicate the coordinates of the point in raster space whereas
     # the geometry is in projected space
     outlet_layer.CreateField(ogr.FieldDefn('i', ogr.OFTInteger))
     outlet_layer.CreateField(ogr.FieldDefn('j', ogr.OFTInteger))
@@ -4142,7 +4143,7 @@ def detect_outlets(d8_flow_dir_raster_path_band, target_outlet_vector_path):
     outlet_layer.StartTransaction()
 
     cdef time_t last_log_time = ctime(NULL)
-    # iterate by iterblocks so ManagedRaster can efficiently cache reads
+    # iterate by iterblocks so ReadAsArray can efficiently cache reads
     # and writes
     LOGGER.info('outlet detection: 0% complete')
     for block_offsets in pygeoprocessing.iterblocks(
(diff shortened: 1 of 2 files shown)


src/pygeoprocessing/routing/routing.pyx
```diff
@@ -4190,7 +4241,9 @@ def detect_outlets(d8_flow_dir_raster_path_band, target_outlet_vector_path):
                     outlet_feature = None
                     outlet_point = None
 
-    LOGGER.info('outlet detection: committing transaction')
+    d8_flow_dir_raster = None
+    d8_flow_dir_band = None
+    LOGGER.info('outlet detection: 100% complete -- committing transaction')
```
emlys, line 4246: I like that you added the logging at 0 and 100% completion - it can be confusing having the 5 second logging intervals if it never indicates it actually finished.

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -4130,49 +4135,95 @@ def detect_outlets(d8_flow_dir_raster_path_band, target_outlet_vector_path):
...
+
+        # Test for bottom border and if so stripe nodata on the bottom margin
+        win_ysize_border = 0
+        if yoff+win_ysize < raster_y_size-1:
+            win_ysize_border += 1
+        else:
+            flow_dir_block[-1, :] = flow_dir_nodata
+
+        # Read iterblock plus a possible margin on top/bottom/left/right side
+        d8_flow_dir_band.ReadAsArray(
```
emlys, line 4191: Do you lose the efficiency of reading by `iterblocks` blocks when you add a 1 pixel border, since the dimensions now cover multiple memory blocks? I did the same thing in delineateit and I don't know if there's any way around it.

tests/test_routing.py
```diff
@@ -150,7 +152,8 @@ def test_detect_outlets(self):
             (0, 0), (1, 0), (2, 0), (3, 0),
             (3, 1),
             (0, 2),
-            (1, 3), (2, 3)}
+            (1, 3), (2, 3),
+            (511, 511)}
```
emlys, line 156: It would be kinda nice to test cases on the edges of blocks because there's potential for weird stuff to happen there. I tried to do that in delineateit by mocking `iterblocks`: https://github.com/natcap/invest/blob/00b6c6e77f5f335e96d326d0c91d20dfcdc132f7/tests/test_delineateit.py#L411-L456
Up to you whether it's worth it to do that here.

emlys (commented):
Looks good! Just had a couple of comments. Thanks for doing this.

emlys (approved):

src/pygeoprocessing/routing/routing.pyx
richpsharp: No it's still pretty fine. GDAL can cache many memory blocks so when you hit the border it'll have those cached and the next window you slide over it won't need to re-load them. Also if you make `iterblocks` operate over many blocks with generous `largest_block` you can minimize that cost even more. In practice I haven't seen it be anything significant enough to measure or worry about.

emlys: neat!

src/pygeoprocessing/routing/routing.pyx
richpsharp: Ahh that's nice you noticed / bad that you noticed when it's not there. I agree that's annoying and I try to fix it whenever I get the chance.

tests/test_routing.py
richpsharp: I agree, I was thinking of that and I couldn't think of an easy way to do it and decided it was probably fine. `mock`ing `iterblocks` is a great way to force that. I copied most of your test case you linked and finagled it to work with `detect_outlets` in 017339a9c59aae9e2884685b8466d141d8b35161.

richpsharp:
Thanks @emlys, I see you already approved the PR but I still added that test case you suggested for testing against memory aligned rows/columns.

PR #156, opened 2021-01-29
Feature/raster stitch 150

### LINKED ISSUE #150: Add raster stitching functionality
A common use case is for a GIS pipeline to make many small rasters in different projections (say watersheds or countries) and then need a final result in a single raster. Non-trivial things:

* how should one handle different projections between base and target raster?
    * area invariant values should only need a pixel size change and a common interpolation rule
    * area variant values are difficult to handle in general, but perhaps lat/lng is an okay one to base everything off of?
* how should one handle base rasters which intersect with non-nodata values?
    * options to replace old, ignore new, or add
* how to handle batch processing of multi images

richpsharp:
Creates a new function called `stitch_rasters` that replaces `merge_rasters` that has the advantage of dealing with different projections, accounting for area variant values when stitching from projected to wgs84 units, allowing incremental updates to stitching, and 3 replacement functions 'etch', 'replace', and 'add' when stitching on top of existing data.

This PR makes all those changes, plus test cases to cover the new functionality, updates to history, and a minor update to the test suite class name to simply `TestGeoprocessing`.

Closes #150

### DIFF
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -3799,6 +3799,7 @@ def stitch_rasters(
         base_raster_path_band_list,
         resample_method_list,
         target_stitch_raster_path_band,
+        overlap_algorithm='etch',
         area_weight_m2_to_wgs84=False,
         osr_axis_mapping_strategy=DEFAULT_OSR_AXIS_MAPPING_STRATEGY):
     """Stitch the raster in the base list into the existing target.
@@ -3819,6 +3820,17 @@ def stitch_rasters(
             If the pixel size or projection are different between base and
             target the base is warped to the target's cell size and target
             with the interpolation method provided.
+        overlap_algorithm (str): this value indicates which algorithm to use
+            when a raster is stitched on non-nodata values in the target
+            stitch raster. It can be one of the following:
+                'etch': write a value to the target raster only if the target
+                    raster pixel is nodata. If the target pixel is non-nodata
+                    ignore any additional values to write on that pixel.
+                'replace': write a value to the target raster irrespective
+                    of the value of the target raster
+                'add': add the value to be written to the target raster to
+                    any existing value that is there. If the existing value
+                    is nodata, treat it as 0.0.
         area_weight_m2_to_wgs84 (bool): If `True` the stitched raster will be
             converted to a per-area value before reprojection to wgs84, then
             multiplied by the m^2 area per pixel in the wgs84 coordinate
@@ -3833,6 +3845,12 @@ def stitch_rasters(
     Return:
         None.
     """
+    valid_overlap_algorithms = ['etch', 'replace', 'add']
+    if overlap_algorithm not in valid_overlap_algorithms:
+        raise ValueError(
+            f'overlap algorithm {overlap_algorithm} is not one of '
+            f'{valid_overlap_algorithms}')
+
     if not _is_raster_path_band_formatted(target_stitch_raster_path_band):
         raise ValueError(
             'Expected raster path/band tuple for '
@@ -4023,8 +4041,30 @@ def stitch_rasters(
                 base_nodata_mask = numpy.ones(
                     base_array.shape, dtype=numpy.bool)
 
-            valid_mask = ~base_nodata_mask & target_nodata_mask
-            target_array[valid_mask] = base_array[valid_mask]
+            if overlap_algorithm == 'etch':
+                # place values only where target is nodata
+                valid_mask = ~base_nodata_mask & target_nodata_mask
+                target_array[valid_mask] = base_array[valid_mask]
+            elif overlap_algorithm == 'replace':
+                # write valid values into the target -- disregard any
+                # existing values in the target
+                valid_mask = ~base_nodata_mask
+                target_array[valid_mask] = base_array[valid_mask]
+            elif overlap_algorithm == 'add':
+                # add values to the target and treat target nodata as 0.
+                valid_mask = ~base_nodata_mask
+                masked_target_array = target_array[valid_mask]
+                target_array[valid_mask] = (
+                    base_array[valid_mask] +
+                    numpy.where(numpy.isclose(
+                        masked_target_array, target_nodata),
+                        0, masked_target_array))
+            else:
+                raise RuntimeError(
+                    f'overlap_algorithm {overlap_algorithm} was not defined '
+                    'but also not detected earlier -- this should never '
+                    'happen')
+
             target_band.WriteArray(
                 target_array,
                 xoff=_offset_vars['target_xoff']+_offset_vars['xoff_clip'],
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -3610,7 +3610,10 @@ def stitch_rasters(
             values in the base rasters will not be written into the target.
             If the pixel size or projection are different between base and
             target the base is warped to the target's cell size and target
-            with the interpolation method provided.
+            with the interpolation method provided. If any part of the
+            base raster lies outside of the target, that part of the base
+            is ignored. A warning is logged if the entire base raster is
+            outside of the target bounds.
         overlap_algorithm (str): this value indicates which algorithm to use
             when a raster is stitched on non-nodata values in the target
             stitch raster. It can be one of the following:
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -3783,3 +3783,10 @@ def raster_to_numpy_array(raster_path, band_id=1):
     band = None
     raster = None
     return array
+
+
+def stitch_rasters(
+        base_raster_path_band_list,
+        target_stitch_raster_path,
+        ):
+    pass
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -76,6 +76,7 @@ _GDAL_TYPE_TO_NUMPY_LOOKUP = {
     gdal.GDT_CFloat64: numpy.complex64,
 }
 
+
 def raster_calculator(
         base_raster_path_band_const_list, local_op, target_raster_path,
         datatype_target, nodata_target,
@@ -3798,6 +3799,7 @@ def stitch_rasters(
         base_raster_path_band_list,
         resample_method_list,
         target_stitch_raster_path_band,
+        area_weight_m2_to_wgs84=False,
         osr_axis_mapping_strategy=DEFAULT_OSR_AXIS_MAPPING_STRATEGY):
     """Stitch the raster in the base list into the existing target.
 
@@ -3817,6 +3819,11 @@ def stitch_rasters(
             If the pixel size or projection are different between base and
             target the base is warped to the target's cell size and target
             with the interpolation method provided.
+        area_weight_m2_to_wgs84 (bool): If `True` the stitched raster will be
+            converted to a per-area value before reprojection to wgs84, then
+            multiplied by the m^2 area per pixel in the wgs84 coordinate
+            space. This is useful when the quantity being stitched is a total
+            quantity per pixel rather than a per unit area density.
         osr_axis_mapping_strategy (int): OSR axis mapping strategy for
             ``SpatialReference`` objects. Defaults to
             ``geoprocessing.DEFAULT_OSR_AXIS_MAPPING_STRATEGY``. This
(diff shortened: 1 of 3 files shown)


src/pygeoprocessing/geoprocessing.py
```diff
@@ -3788,3 +3584,352 @@ def raster_to_numpy_array(raster_path, band_id=1):
...
+            stitch raster. It can be one of the following:
+                'etch': write a value to the target raster only if the target
+                    raster pixel is nodata. If the target pixel is non-nodata
+                    ignore any additional values to write on that pixel.
+                'replace': write a value to the target raster irrespective
+                    of the value of the target raster
+                'add': add the value to be written to the target raster to
+                    any existing value that is there. If the existing value
+                    is nodata, treat it as 0.0.
+        area_weight_m2_to_wgs84 (bool): If `True` the stitched raster will be
```
phargogh: Double-backticks for RST?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -3788,3 +3584,352 @@ def raster_to_numpy_array(raster_path, band_id=1):
...
+            f'Location: "{target_stitch_raster_path_band[0]}"" '
+            f'GIS type: {gis_type}')
+    target_raster_info = get_raster_info(target_stitch_raster_path_band[0])
+    if target_stitch_raster_path_band[1] > len(target_raster_info['nodata']):
+        raise ValueError(
+            'target_stitch_raster_path_band refers to a band that exceeds '
+            'the number of bands in the raster:\n'
+            f'''target_stitch_raster_path_band[1]: {
+                target_stitch_raster_path_band[1]} '''
+            f'n bands: {len(target_raster_info["nodata"])}')
```
phargogh: It took me a second to realize that this will join a single-quote string with a triple-quote f-string with a single-quote f-string.  Plus, the triple-quote string will include the indentation from the contained line.  Since it switches from having a new line defined and the triple-quote will include the bonus tabs, could I suggest having a consistent quoting scheme, either all within a `textwrap.dedent`ed triple-quote or in a series of regular-quotes?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -3788,3 +3584,352 @@ def raster_to_numpy_array(raster_path, band_id=1):
...
+
+    target_raster = gdal.OpenEx(
+        target_stitch_raster_path_band[0], gdal.OF_RASTER | gdal.GA_Update)
+    target_band = target_raster.GetRasterBand(
+        target_stitch_raster_path_band[1])
+    target_inv_gt = gdal.InvGeoTransform(target_raster_info['geotransform'])
+    target_raster_x_size, target_raster_y_size = target_raster_info[
+        'raster_size']
+    for (raster_path, raster_band_id), resample_method in zip(
+            base_raster_path_band_list, resample_method_list):
```
phargogh, line 3698: Since `zip` will stop iterating when the end of the shortest iterable is reached, should there also be a check for equal length of the iterables?

Imagine a case where `len(resample_method_list) < len(base_raster_path_band_list)` and the function would silently not stitch in the last few rasters(s).

src/pygeoprocessing/geoprocessing.py
```diff
@@ -3788,3 +3584,352 @@ def raster_to_numpy_array(raster_path, band_id=1):
...
+        try:
+            # merge the bounding boxes only to see if they don't intersect
+            _ = merge_bounding_box_list(
+                [projected_raster_bounding_box,
+                 target_raster_info['bounding_box']], 'intersection')
+        except ValueError:
+            LOGGER.warning(
+                f'the raster at "{raster_path}"" does not intersect the '
+                f'stitch raster at "{target_stitch_raster_path_band[0]}", '
+                'skipping...')
```
phargogh: This is an interesting case and interesting behavior ... could you add a note in the docstring mentioning what happens in this case?  It doesn't look to be specifically described.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -3788,3 +3584,352 @@ def raster_to_numpy_array(raster_path, band_id=1):
...
+
+        if warped_raster and area_weight_m2_to_wgs84:
+            # determine base area per pixel currently and area per pixel
+            # once it is projected to wgs84 pixel sizes
+            base_pixel_area_m2 = abs(numpy.prod(raster_info['pixel_size']))
+            base_stitch_raster_info = get_raster_info(
+                base_stitch_raster_path)
+            lat_min, lat_max = [
+                base_stitch_raster_info['bounding_box'][_]
+                for _ in [1, 3]]
```
phargogh: I get the intent of using `_` as a single-character variable, but it is a little inconsistent with how python typically uses `_`.  And as an aside, the same statement can be expressed without the comprehension: `_, lat_min, _, lat_max = base_stitch_raster_info['bounding_box']`.  Not an important runtime improvement, but maybe slightly simpler?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -3788,3 +3584,352 @@ def raster_to_numpy_array(raster_path, band_id=1):
...
+            n_rows = base_stitch_raster_info['raster_size'][1]
+            # this column is a longitude invariant latitude variant pixel
+            # area for scaling area dependent values
+            m2_area_per_lat = _create_latitude_m2_area_column(
+                lat_min, lat_max, n_rows)
```
phargogh, line 3754: Neat.  I was about to comment about the possibility of a `MemoryError` here (like if we had super high-res outputs), but I suppose we can handle that case when it happens.  Was that your thinking as well?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -3788,3 +3584,352 @@ def raster_to_numpy_array(raster_path, band_id=1):
...
+
+            def _mult_op(base_array, base_nodata, scale):
+                """Scale non-nodata by scale."""
+                result = numpy.empty(base_array.shape)
```
phargogh: I see the default `dtype` here from numpy is `numpy.float64`.  For the sake of explicitness, would it be worth defining?  As another consideration, would it be worth using the `base_arra.dtype`, in case there's a memory savings to be had?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -3788,3 +3584,352 @@ def raster_to_numpy_array(raster_path, band_id=1):
...
+                result[:] = base_array
+                if base_nodata is not None:
+                    valid_mask = ~numpy.isclose(base_array, base_nodata)
```
phargogh, line 3760: We learned [fairly recently](https://github.com/natcap/invest/pull/269#discussion_r473254105) that integer pixel comparison (`array != nodata`) is an order of magnitude faster than `numpy.isclose` comparison.  I appreciate the simplicity of a single, general-purpose function call, but would it be worth taking advantage of this speedup here?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -3788,3 +3584,352 @@ def raster_to_numpy_array(raster_path, band_id=1):
...
+                else:
+                    valid_mask = numpy.ones(
+                        base_array.shape, dtype=numpy.bool)
```
phargogh, line 3763: We [also learned fairly recently](https://github.com/natcap/invest/pull/269#issuecomment-679471260) that `slice(None)` is the equivalent of `[:]`, and offers order-of-magnitude faster indexing.  Would it be worth using?

phargogh (changes requested):
Neat function!  Just a few comments, nothing major.  Let me know what you think!

src/pygeoprocessing/geoprocessing.py
richpsharp: hah sure, i did it this way because one time doug did a review where i was using an f string but didn't need to. i'd prefer to make it consistent

src/pygeoprocessing/geoprocessing.py
richpsharp: Memory error is not a problem-- that's the beauty of this. This function returns a 1 dimensional column of areas because the areas are not variant with longitude. This is the motivation for why I changed `raster_calculator` a couple years ago to allows brodcastable arrays rather than just arrays that are the correct dimensions. `raster_calculator` will slice and broadcast as needed to fit whatever raster block its working on.

src/pygeoprocessing/geoprocessing.py
richpsharp: Where should i use `slice(None)`? Instead of `numpy.ones`?

src/pygeoprocessing/geoprocessing.py
richpsharp: That timing is not surprising since it's a float with a range test vs equality of an int. But here there's no guarantee `base_nodata` is an int. It's why we have `numpy.isclose` all throughout the place because of large nodata values that are not quite equal. Am I understanding your comment correctly?

src/pygeoprocessing/geoprocessing.py
richpsharp: I think you meant on line 3752 right? Changed in 073dfdc

src/pygeoprocessing/geoprocessing.py
richpsharp: Changed in 073dfdc.

src/pygeoprocessing/geoprocessing.py
richpsharp: Changed in 951af67

src/pygeoprocessing/geoprocessing.py
richpsharp: I added a couple of sentences to the `target_stitch_raster_band` description in 7c84b11

src/pygeoprocessing/geoprocessing.py
richpsharp: Changed in 4548a2d

src/pygeoprocessing/geoprocessing.py
richpsharp: Added a test for this in 4548a2d and then added a test to test for that in 5c78a86.

src/pygeoprocessing/geoprocessing.py
richpsharp: Fixed in 97e9dad

src/pygeoprocessing/geoprocessing.py
richpsharp: Actually there was a reason I did that as I found out when the tests failed. I'm not trying to match the base here but the target, and by doing a mult we know it needs to be a float... Anyway, I changed the whole thing in 53fd6ae2015e3bb399c4154b3c1800be4f81d413 so that it uses the type of the target raster.

richpsharp:
> Neat function! Just a few comments, nothing major. Let me know what you think!

Hi James, I changed all the stuff you suggested except the check for `~isclose` w/r/t a nodata value. I wanna keep that in there since there's the chance for it to be a float. If runtime performance were an issue we could if/else it to have the fast integer test or the slow float test depending on the type but I'm not worried that testing for nodata is a big issue there.

The other thing I just want to note, because I'm proud of it, is the lat/lng area array being 1D not 2D so there's no memory error issue there. It's a key part of this function so just wanna make sure you didn't miss it and thought it was actually the worst part of the function.

src/pygeoprocessing/geoprocessing.py
phargogh: > This is the motivation for why I changed raster_calculator a couple years ago to allows brodcastable arrays rather than just arrays that are the correct dimensions. raster_calculator will slice and broadcast as needed to fit whatever raster block its working on.

Right, I totally get that! so we know that `raster_calculator` won't `MemoryError`, which is great.  The `MemoryError` I was referring to was about:

> This function returns a 1 dimensional column of areas because the areas are not variant with longitude.

Imagine if we had a base raster with millions of rows.  `__create_latitude_m2_area_column` will happily create a 1 x n column of pixel areas.  Are we at risk of `MemoryError` on that?  Feels like a bit of an edge case (and even less likely now that InVEST binaries use 64-bit builds), but what do you think?

src/pygeoprocessing/geoprocessing.py
phargogh: Yep!

src/pygeoprocessing/geoprocessing.py
phargogh: Actually, I was thinking of cases where we construct a boolean array where we select all pixels (`valid_mask = numpy.ones(base_array.shape, dtype=numpy.bool)`).  The point of `slice(None)` is that you can assign it to a variable, which is great when we're conditionally constructing a mask.

```
a = slice(None)  # Works great, equivalent of array[:], so you could now say array[a] == array[:]
b = :  # This is a syntax error, of course
```

phargogh (changes requested):
Thanks, @richpsharp !

No problem at all about the `isclose` mask here, and I left a comment about the specific situation that I was wondering about the `MemoryError` on.  I also left a clarification about what I meant with `slice(None)` and when it could be useful.  Otherwise, looks good to me!

src/pygeoprocessing/geoprocessing.py
richpsharp: I really think that issue is okay. If you have millions of rows, that's still only a few megabytes (estimating 4 bytes * millions numbers of bytes). I'd start to worry about a billion rows but at that point you're talking about a resolution on Earth of about .02m/pixel (estimating 180/1000000000*111km/degree at equator).  But even there you'd still only use 8gb to store the row but you'd be dealing with a 1billion X 2billion element raster on disk that would be on the order of petabytes of hardrive space... 

Am I doing my math right? I'm trying to make the case that this an order of magnitude problem that cannot be exceeded even with crazy numbers, and by the time we got to the technology level that makes that crazy the norm, we'd still only have 8gb array in memory. Which I'm sure would be fine with the many terabytes of local ram we'd have.

Does that sound right?

src/pygeoprocessing/geoprocessing.py
richpsharp: Okay I thought I understood, but when I tried what I thought you were saying I'm not seeing a consistent speedup. Can you check? Or maybe this is already taking too long and isn't really about this PR... Here's a gist of what I tried, I was trying to make a real world scenario with rando nodata values and some kind of oepration etc. I only time the parts where the functions are called, not the setup: https://gist.github.com/richpsharp/94a65b0b01c6fffcb47ec00c999f7938

Running a few times I see this:

```
(py38) Rich@DESKTOP-ETOEL2F D:\repositories
> python timeit.py
D: 256
op_time: 0.1299595832824707
noneslice_op_time: 0.13809561729431152
noneslice performance increase: 94.11

(py38) Rich@DESKTOP-ETOEL2F D:\repositories
> python timeit.py
D: 256
op_time: 0.1300041675567627
noneslice_op_time: 0.14000415802001953
noneslice performance increase: 92.86

(py38) Rich@DESKTOP-ETOEL2F D:\repositories
> python timeit.py
D: 256
op_time: 1.1344609260559082
noneslice_op_time: 1.1809608936309814
noneslice performance increase: 96.06

(py38) Rich@DESKTOP-ETOEL2F D:\repositories
> python timeit.py
D: 256
op_time: 1.1216988563537598
noneslice_op_time: 1.1957592964172363
noneslice performance increase: 93.81

(py38) Rich@DESKTOP-ETOEL2F D:\repositories
> python timeit.py
D: 256
op_time: 1.1153075695037842
noneslice_op_time: 1.1903018951416016
noneslice performance increase: 93.70

(py38) Rich@DESKTOP-ETOEL2F D:\repositories
> python timeit.py
D: 256
op_time: 1.1111719608306885
noneslice_op_time: 1.1871743202209473
noneslice performance increase: 93.60
```

Basically I see the `slice(None)` is always a liiiitle bit slower. That's what I'd kinda expect since I'm turning one function call into two and I'd sort of hope the intializers that numpy provides are actually tuned to be used for initialization. Can you take a look at that? And/or maybe we can move this to slack instead of this PR assuming you don't need me to do anything else here?

richpsharp:
> Thanks, @richpsharp !
> 
> No problem at all about the `isclose` mask here, and I left a comment about the specific situation that I was wondering about the `MemoryError` on. I also left a clarification about what I meant with `slice(None)` and when it could be useful. Otherwise, looks good to me!

Okay! You said it looks good but the review had a "requesting changes". I commented on those things above and re-requested a review. Also if that timing for the `slice(None)` looks interesting maybe we can move it over to the slack software team channel?

src/pygeoprocessing/geoprocessing.py
phargogh: Ach, sorry ... The `slice(None)` advantage here is only useful in a _very_ specific case, and I agree, no need to pursue it further here.  I'll follow up about this in slack a little later.

src/pygeoprocessing/geoprocessing.py
phargogh: Yep, yep, that sounds right to me!  And if we're working with petabyte-size rasters, we'll run into other more practical problems (like that we'll be pushing the upper limit of number of pixels supported by GeoTiff) before this ends up being a memory issue.

phargogh (approved):
Looks good!  I'll follow up separately with what I was thinking about the `slice(None)` thing and the specific case where it could be useful.

PR #157, opened 2021-02-01
add guard against filling very large pits

### LINKED ISSUE #117: Threshold for max pitfill
There are some degenerate pitfill cases such as death valley or the dead sea that don't make sense to fill. To prevent this, expose a "max pitfill size" parameter in `pit_fill` that stops searching for pits once the search space gets to be that many pixels.

richpsharp:
This fixes #117 which deals with a problem in some DEMs where there are very large natural depressions but the fill algorithm attempts to fill them to the top. You wouldn't want to fill culverins in volcanoes or death valley. I defaulted the number of pixels to maximum fill to 500 in both the search and the fill just in case either runs away. 500 would be a pretty big natural depression, but an order of magnitude smaller than the crazy degenerate cases. In practice for a 30m dem I think this value would probably be something like 10-20.

This PR adds the functionality for this feature, a test to cover that functionality, and a note in history.

### DIFF
--- a/src/pygeoprocessing/routing/routing.pyx
+++ b/src/pygeoprocessing/routing/routing.pyx
@@ -900,9 +900,7 @@ def fill_pits(
                                 xi_n, yi_n) == 1:
                             # been set before on a previous iteration, skip
                             continue
-                        if n_height == center_val and (
-                                flat_region_mask_managed_raster.get(
-                                    xi_n, yi_n) == mask_nodata):
+                        if n_height == center_val:
                             # only grow if it's at the same level and not
                             # previously visited
                             search_queue.push(
--- a/src/pygeoprocessing/routing/routing.pyx
+++ b/src/pygeoprocessing/routing/routing.pyx
@@ -84,6 +84,10 @@ cdef int *D8_YOFFSET = [0, -1, -1, -1, 0, +1, +1, +1]
 # as a D8 direction
 cdef int* D8_REVERSE_DIRECTION = [4, 5, 6, 7, 0, 1, 2, 3]
 
+# default of number of pixels to naturally drain. This number was derived off
+# of reasonable expectations from a 30m DEM
+cdef int _MAX_PIXEL_FILL_COUNT = 500
+
 # exposing stl::priority_queue so we can have all 3 template arguments so
 # we can pass a different Compare functor
 cdef extern from "<queue>" namespace "std" nogil:
@@ -607,7 +611,7 @@ def _generate_read_bounds(offset_dict, raster_x_size, raster_y_size):
 
 def fill_pits(
         dem_raster_path_band, target_filled_dem_raster_path,
-        working_dir=None,
+        working_dir=None, max_pixel_fill_count=_MAX_PIXEL_FILL_COUNT,
         raster_driver_creation_tuple=DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS):
     """Fill the pits in a DEM.
 
@@ -631,6 +635,9 @@ def fill_pits(
             it is created by this call. If None, a temporary directory is
             created by tempdir.mkdtemp which is removed after the function
             call completes successfully.
+        max_pixel_fill_count (int): maximum number of pixels to fill a pit
+            before leaving as a depression. Useful if there are natural
+            large depressions.
         raster_driver_creation_tuple (tuple): a tuple containing a GDAL driver
             name string as the first element and a GDAL creation options
             tuple/list as the second. Defaults to a GTiff driver tuple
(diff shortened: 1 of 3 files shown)


src/pygeoprocessing/routing/routing.pyx
```diff
@@ -850,34 +860,45 @@ def fill_pits(
...
                             xi_n, yi_n)
                         if _is_close(n_height, dem_nodata, 1e-8, 1e-5):
-                            nodata_drain = 1
+                            natural_drain_exists = 1
                             continue
                         if n_height < center_val:
-                            downhill_drain = 1
+                            natural_drain_exists = 1
+                            continue
+                        if flat_region_mask_managed_raster.get(
```
dcdenu4, line 899: Just a clarifying question. Can `flat_region_mask_managed_raster.get(xi_n, yi_n)` be anything but 1 or `nodata`? Here we skip if it's been set and identified before, but on 903,904 it's checked again against `mask_nodata`. Just seeing if that redundancy is necessary?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -850,34 +860,45 @@ def fill_pits(
...
+                search_steps = 0
                 while not search_queue.empty():
                     xi_q = search_queue.front().xi
                     yi_q = search_queue.front().yi
                     search_queue.pop()
+                    if search_steps > max_pixel_fill_count:
+                        # clear the search queue and quit
+                        while not search_queue.empty():
+                            search_queue.pop()
+                        natural_drain_exists = 1
```
dcdenu4, line 880: I get the use of `natural_drain_exists` over multiple variable indicators. But in this case, if we exit because of a "natural depression", then `natural_drain_exists` set to `true` is not really the case, it doesn't naturally drain! I'm not suggesting an edit or adding more variables, but this is a thought that jumped out to me.

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -923,7 +955,7 @@ def fill_pits(
                             break
 
                         if pit_mask_managed_raster.get(
-                                xi_n, yi_n) == feature_id:
+                                xi_n, yi_n) != mask_nodata:
```
dcdenu4, line 956: `mask_nodata` is better here because it might be possible `feature_id` for `xi_n`, `yi_n` is not `feature_id` but has still been visited and has a different `feature_id`?

dcdenu4 (commented):
Hey @richpsharp, this looks good to me. I worked my way through the function and the addition seems to fit nicely. I had a couple general comments and questions. I was curious if there should be a test for plateaus as well as pits?

src/pygeoprocessing/routing/routing.pyx
richpsharp: It's not necessary! I missed that I'd checked it below. Removed in 485c3f6f19bb031b60912f741543db62d80c6c0d

src/pygeoprocessing/routing/routing.pyx
richpsharp: I was thinking that if there's a natural depression then that is the natural drain for the region, such as a lake or culverin. That's what I'm trying to capture here - not everything drains to the ocean. I get how it sounds confusing though. I've been trying to think of some alternatives... At first I was thinking "sink"? But that's really not a drain for the places that do drain. And this bit of code comes out of a case where I *did* have two variables checking the same thing `drains_to_nodata` `drains_to_edge` and it didn't matter, hence the change to `natural_drain_exists`. To me this seems fine, but if you've got an idea that's clearer I'm happy to use it.

src/pygeoprocessing/routing/routing.pyx
richpsharp: Correct. That used to not be the case when you couldn't stop on the iteration of a fill region part way through.

richpsharp:
> Hey @richpsharp, this looks good to me. 

Okay I'll merge but -- 

>I worked my way through the function and the addition seems to fit nicely. I had a couple general comments and questions. I was curious if there should be a test for plateaus as well as pits?

Plateaus are fine since they drain and the function already terminates if it's iterating across a plateau (hence one of the `natural_drain_exists` cases you pointed out above. What's different now is it will stop searching a giant plateau for a drain which doesn't matter anyway because it already drains and wouldn't need to be filled... Is that what you mean by a test for a plateau?

richpsharp:
> > Hey @richpsharp, this looks good to me.
> 
> Okay I'll merge but --
> 
> > I worked my way through the function and the addition seems to fit nicely. I had a couple general comments and questions. I was curious if there should be a test for plateaus as well as pits?
> 
> Plateaus are fine since they drain and the function already terminates if it's iterating across a plateau (hence one of the `natural_drain_exists` cases you pointed out above. What's different now is it will stop searching a giant plateau for a drain which doesn't matter anyway because it already drains and wouldn't need to be filled... Is that what you mean by a test for a plateau?

Oh wait I can't merge, you didn't approve changes. Requesting your review again @dcdenu4

src/pygeoprocessing/routing/routing.pyx
dcdenu4: Nope, that sounds good!

dcdenu4 (approved):
Thanks for the explanations Rich!

dcdenu4:
@richpsharp wait. I don't see the 485c3f6f19bb031b60912f741543db62d80c6c0d commit anywhere? Did that get pushed here?

richpsharp:
> @richpsharp wait. I don't see the 485c3f6f19bb031b60912f741543db62d80c6c0d commit anywhere? Did that get pushed here?

hah no, i just pushed

PR #160, opened 2021-02-08
Replace deprecated numpy.bool

emlys:
> DeprecationWarning: `np.bool` is a deprecated alias for the builtin `bool`. To silence this warning, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.

Just replacing these, goes with https://github.com/natcap/invest/issues/255


richpsharp (approved):
Great!

PR #164, opened 2021-02-09
Feature/allow fixed flow accumulation mahler 162

richpsharp:
Adds a `autotune_flow_accumulation` parameter that when True will attempt to get reasonable values of threshold flow accumulation parameters based on the drop distance of order 1 and order 2 streams.

Didn't update HISTORY because this feature hasn't been released yet, otherwise updated docstring and added a test to cover this behavior.

### DIFF
--- a/src/pygeoprocessing/routing/routing.pyx
+++ b/src/pygeoprocessing/routing/routing.pyx
@@ -3009,6 +3009,7 @@ def extract_strahler_streams_d8(
         long min_flow_accum_threshold=100,
         int river_order=5,
         float min_p_val=0.05,
+        autotune_flow_accumulation=False,
         osr_axis_mapping_strategy=DEFAULT_OSR_AXIS_MAPPING_STRATEGY):
     """Extract Strahler order stream geometry from flow accumulation.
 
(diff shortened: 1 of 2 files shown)


src/pygeoprocessing/routing/routing.pyx
```diff
@@ -3056,12 +3057,17 @@ def extract_strahler_streams_d8(
             the above arguments. Contains the fields "order" and "parent" as
             described above.
         min_flow_accum_threshold (int): minimum number of upstream pixels
-            required to create a stream, the final value may be adjusted based
-            on significant differences in 1st and 2nd order streams.
+            required to create a stream. If `autotune_flow_accumulation`
```
phargogh: Double-backticks?

phargogh (changes requested):
Just one minor RST issue and this otherwise looks good to me!  Sorry for the trivial change request.

src/pygeoprocessing/routing/routing.pyx
richpsharp: okay just added in cb2c8e25b7d104d0a221b16bebea9dc54a96506b (used the web editor actually)

richpsharp:
Thanks for checking @phargogh, tossed on an extra set of ``s

phargogh (approved):

phargogh:
Thanks @richpsharp !

PR #166, opened 2021-02-10
Feature/outlet at confluence 165

### LINKED ISSUE #165: Add "outlet at confluence" flag to subwatershed creation
Currently subwatersheds are created at the confluence point, but it is often desirable to make them one step up. But then again someone might want it the other way around so I've been told. This issue is to add an `outlet_at_confluence` flag that is defaulted to False that will provide this behavior.

richpsharp:
This feature adds an `outlet_at_confluence` flag which adjusts whether subwatersheds seed from the junction of upstream streams or one pixel up from that. Apparently both are useful. Doing this is a little tricky because you need to know in raster space what one pixel upstream on your arbitrary line is. I did this by adding `ds_x_1`, `ds_y_1` fields to the stream generator (the 1 means 1 pixel upstream) then I modified the subwatershed generator to use the upstream points on a stream as a segment starter if we're doing a confluence root or the `ds_x_1`, `ds_y_1` points if it's not.

I added a test to cover this functionality which is interesting by using non confluences you almost double the number of watersheds you have if you only ever join two streams. 

I didn't update history because this feature isn't released yet and the history note doesn't talk about the subfeatures in subwatershed generation.

I also added ONE BONUS bit of data which is an `outlet_x`/`outlet_y` field on the subwatersheds that give the raster x/y coordinates of the outlet. This is useful for sampling any raster that's aligned with these data at their outlets such as flow accumulation, DEM height, and maybe something else. It makes the class of flood filling algorithms a lot easier to implement because now we've got little discrete geometries that can be used to offset local hydrological fills.

Closes #165

### DIFF
--- a/src/pygeoprocessing/routing/routing.pyx
+++ b/src/pygeoprocessing/routing/routing.pyx
@@ -3824,18 +3824,20 @@ def calculate_subwatershed_boundary(
         strahler_stream_vector_path (str): path to stream segment vector
         target_watershed_boundary_vector_path (str): path to created vector
             of linestring for watershed boundaries. Contains the fields:
-            "stream_id": this is the stream ID from the
-                ``strahler_stream_vector_path`` that corresponds to this
-                subwatershed.
-            "terminated_early": if set to 1 this watershed generation was
-                terminated before it could be complete. This value should
-                always be 0 unless something is wrong as a software bug
-                or some degenerate case of data.
-            "outlet_x", "outlet_y": this is the x/y coordinate in raster
-                space of the outlet of the watershed. It can be useful when
-                determining other properties about the watershed when indexed
-                with underlying raster data that created the streams in
-                ``strahler_stream_vector_path``.
+
+            * "stream_id": this is the stream ID from the
+              ``strahler_stream_vector_path`` that corresponds to this
+              subwatershed.
+            * "terminated_early": if set to 1 this watershed generation was
+              terminated before it could be complete. This value should
+              always be 0 unless something is wrong as a software bug
+              or some degenerate case of data.
+            * "outlet_x", "outlet_y": this is the x/y coordinate in raster
+              space of the outlet of the watershed. It can be useful when
+              determining other properties about the watershed when indexed
+              with underlying raster data that created the streams in
+              ``strahler_stream_vector_path``.
+
         max_steps_per_watershed (int): maximum number of steps to take when
             defining a watershed boundary. Useful if the DEM is large and
             degenerate or some other user known condition to limit long large
--- a/src/pygeoprocessing/routing/routing.pyx
+++ b/src/pygeoprocessing/routing/routing.pyx
@@ -3821,7 +3821,19 @@ def calculate_subwatershed_boundary(
             raster
         strahler_stream_vector_path (str): path to stream segment vector
         target_watershed_boundary_vector_path (str): path to created vector
-            of stringline for watershed boundaries.
+            of stringline for watershed boundaries. Contains the fields:
+            "stream_id": this is the stream ID from the
+                ``strahler_stream_vector_path`` that corresponds to this
+                subwatershed.
+            "terminated_early": if set to 1 this watershed generation was
+                terminated before it could be complete. This value should
+                always be 0 unless something is wrong as a software bug
+                or some degenerate case of data.
+            "outlet_x", "outlet_y": this is the x/y coordinate in raster
+                space of the outlet of the watershed. It can be useful when
+                determining other properties about the watershed when indexed
+                with underlying raster data that created the streams in
+                ``strahler_stream_vector_path``.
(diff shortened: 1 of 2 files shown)


src/pygeoprocessing/routing/routing.pyx
```diff
@@ -3926,13 +3961,28 @@ def calculate_subwatershed_boundary(
                 working_stack.extend(upstream_fids)
             else:
                 working_stack.pop()
-                if working_feature.GetField('order') > 1:
-                    visit_order_stack.append((working_fid, us_x, us_y))
+                # the `not outlet_at_confluence` bit allows us to seed
+                # even if the order is 1, otherwise confluences fill fill
```
phargogh: Duplicate "fill"?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -3813,11 +3827,26 @@ def calculate_subwatershed_boundary(
             raster
         strahler_stream_vector_path (str): path to stream segment vector
         target_watershed_boundary_vector_path (str): path to created vector
-            of stringline for watershed boundaries.
+            of stringline for watershed boundaries. Contains the fields:
```
phargogh: This is trivial, but I think "stringline" is intended to be "linestring"

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -3039,6 +3039,10 @@ def extract_strahler_streams_d8(
             stream segment outlet.
         * "ds_y" (int): the downstream y coordinate in raster space for the
             stream segment outlet.
+        * "ds_x_1" (int): the downstream x coordinate for one pixel upstream
+            from the stream segment outlet raster space coordinates.
+        * "ds_y_1" (int): the downstream y coordinate for one pixel upstream
+            from the stream segment outlet raster space coordinates.
```
phargogh: Ah, sorry, this reads a little ambiguously to me.  How about something like "the {x,y} coordinate (in raster space) that is 1 pixel upstream from the stream segment outlet"?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -3813,11 +3827,26 @@ def calculate_subwatershed_boundary(
...
+                subwatershed.
+            "terminated_early": if set to 1 this watershed generation was
+                terminated before it could be complete. This value should
+                always be 0 unless something is wrong as a software bug
+                or some degenerate case of data.
+            "outlet_x", "outlet_y": this is the x/y coordinate in raster
+                space of the outlet of the watershed. It can be useful when
+                determining other properties about the watershed when indexed
+                with underlying raster data that created the streams in
+                ``strahler_stream_vector_path``.
```
phargogh: I think this will be a bit malformed in the rendered RST.  Maybe a bulleted list (like in the docstring for `extract_strahler_streams_d8`) would work better?

phargogh (changes requested):
Just had a few comments, all minor, and almost exclusively in the docstrings.  Looks great otherwise!

src/pygeoprocessing/routing/routing.pyx
richpsharp: Fixed in 42efeec

src/pygeoprocessing/routing/routing.pyx
richpsharp: Fixed in 0477682

src/pygeoprocessing/routing/routing.pyx
richpsharp: Thanks! I changed all the wording of {ds,us}_{x,y}{,_1} so they sound better in 766bc8f

src/pygeoprocessing/routing/routing.pyx
richpsharp: Good idea, changed to bulleted list in e704e04

richpsharp: p.s. in schools these days they say "jot dots" instead of "bullet list" for the obvious tragic implications. I wonder when this will roll around to our world. so... just a heads up that'll happen someday.

phargogh:
@richpsharp I see that you've already taken a look at my comments, but maybe at least one revision (e704e04) hasn't been pushed yet?  Or would you like me to take another look?

richpsharp:
> @richpsharp I see that you've already taken a look at my comments, but maybe at least one revision (e704e04) hasn't been pushed yet? Or would you like me to take another look?

Oop, I don't know what that revision was referring to. I did the jot dots in d89c423e8ed8ec803760b59b7c434c50c8597e35

src/pygeoprocessing/routing/routing.pyx
phargogh: Well that is interesting ... I'll try to use "jot dots" instead.  Thanks for the heads up!

phargogh (approved):
🎉

PR #170, opened 2021-03-02
fixes #149 issue with empty vector on create raster from extents

### LINKED ISSUE #149: Trying to create a raster from an empty vector's extents crashes with cryptic error
This came up [on the forums recently](https://community.naturalcapitalproject.org/t/hra-model-typeerror-nonetype-object-is-not-subscriptable/1607/3?u=jdouglass), where a user was trying to create a raster from vector extents when the vector had no features whatsoever.

Minimally reproducible example: 

```python
import pygeoprocessing

from osgeo import gdal, ogr, osr

SRS = osr.SpatialReference()
SRS.ImportFromEPSG(4326)

# Make a vector with no features.
SHP_DRIVER = ogr.GetDriverByName('ESRI Shapefile')
VECTOR_PATH = 'testtest.shp'
VECTOR = SHP_DRIVER.CreateDataSource(VECTOR_PATH)
LAYER = VECTOR.CreateLayer('testtest', SRS)
LAYER = None
VECTOR = None

pygeoprocessing.create_raster_from_vector_extents(
    VECTOR_PATH, 'out_raster.tif', (20, -20), gdal.GDT_Byte, 255)
```

Which results in 

```python
  File "test.py", line 16, in <module>
    VECTOR_PATH, 'out_raster.tif', (20, -20), gdal.GDT_Byte, 255)
  File "/Users/jdouglass/Downloads/HRA_lab1/env37-conda/lib/python3.7/site-packages/pygeoprocessing/geoprocessing.py", line 1004, in create_raster_from_vector_extents
    abs((shp_extent[1] - shp_extent[0]) / target_pixel_size[0])))
TypeError: 'NoneType' object is not subscriptable
```

This appears to still be an issue for determining both [n_rows](https://github.com/natcap/pygeoprocessing/blob/main/src/pygeoprocessing/geoprocessing.py#L1009) and [n_cols](https://github.com/natcap/pygeoprocessing/blob/main/src/pygeoprocessing/geoprocessing.py#L1013) and is present in `main`.

This is obviously a case of a broken contract, but it would still be nice to raise a human-readable error message.

richpsharp:
This raises an issue if a vector is used to create a raster from its extents but it has no extents. Since I was in there I also simplified the extent checking to just the layers as well as a few other lines, moved the check for a valid raster type to the top, and simplified some raster reference management.

Updated history and added a test that @phargogh proposed in #149.

Fixes #149

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,9 @@ Release History
 
 Unreleased Changes
 ------------------
+* Fixed an issue in ``create_raster_from_vector_extents`` that would cause a
+  confusing exception to be raised if there was no geometry in the vector.
+  Now raises a ``ValueError`` with a helpful error message.
 * Changed parameters in ``convolve_2d`` to allow API to set
   ``ignore_nodata_and_edges=True`` while ``mask_nodata=False`` and updated
   docstring to indicate this is useful in cases such as filling nodata holes
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -976,6 +976,10 @@ def create_raster_from_vector_extents(
     Return:
         None
     """
+    if target_pixel_type not in _VALID_GDAL_TYPES:
+        raise ValueError(
+            f'Invalid target type, should be a gdal.GDT_* type, received '
+            f'"{target_pixel_type}"')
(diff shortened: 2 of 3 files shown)


phargogh (approved):
👍

PR #173, opened 2021-03-20
Bugfix/ignore where clause 172

richpsharp:
Fixes an issue that would cause a bounding box error when an "align and resize" was invoked on a global extents lat/lng raster but focused on a narrow bounding box and a local utm target projection. If a mask vector was also applied with a where filter to limit to a particular feature, align and resize would test the extents of the ENTIRE raster rather than just the where subset which would cause a bounding box alignment error. 

This PR detects if a where clause is used and builds a local bounding box that is only the bounding box of the features selected by the where clause. There is also a test added to cover this issue and an update to history.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -57,8 +57,9 @@ Unreleased Changes
 * Added a ``max_pixel_fill_count`` parameter to ``routing.fill_pits`` to
   guard against pitfilling large natural depression. Defaults to 500.
 * Fixed an issue in ``align_and_resize_raster_stack`` that would ignore
-  the bounds of a feature in a mask vector if the ``"where"`` clause was
-  invoked and instead only considered the entire bounds of the vector.
+  the bounds of a feature in a mask vector if the
+  ``"mask_vector_where_filter"`` clause was invoked and instead only
+  considered the entire bounds of the vector.
 
 2.1.2 (2020-12-03)
 ------------------
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -56,6 +56,9 @@ Unreleased Changes
   rather than "units per area".
 * Added a ``max_pixel_fill_count`` parameter to ``routing.fill_pits`` to
   guard against pitfilling large natural depression. Defaults to 500.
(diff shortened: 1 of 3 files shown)


HISTORY.rst
```diff
@@ -56,6 +56,9 @@ Unreleased Changes
   rather than "units per area".
 * Added a ``max_pixel_fill_count`` parameter to ``routing.fill_pits`` to
   guard against pitfilling large natural depression. Defaults to 500.
+* Fixed an issue in ``align_and_resize_raster_stack`` that would ignore
+  the bounds of a feature in a mask vector if the ``"where"`` clause was
```
dcdenu4: Are the quotes necessary inside the ticks?

dcdenu4: Ah maybe it does if it's a SQL WHERE

dcdenu4 (approved):
Hey @richpsharp I think this looks okay. My understanding is:

When using `vector_mask_options` we only want to consider the bounding box for the `WHERE` clause features, not the entire vector. Doing the entire vector causes bounding box errors because we're trying to illogically move masked values somewhere that doesn't make sense.

HISTORY.rst
richpsharp: Right, I think it makes sense but it should really be the keyname which is `"mask_vector_where_filter"` which I changed in 70d61d2

richpsharp:
> Hey @richpsharp I think this looks okay. My understanding is:
> 
> When using `vector_mask_options` we only want to consider the bounding box for the `WHERE` clause features, not the entire vector. Doing the entire vector causes bounding box errors because we're trying to illogically move masked values somewhere that doesn't make sense.

Yep, the biggest issue is if you're accounting for features that should be masked out but they lie outside of the reasonable range of the local projection. My use case was projecting local watersheds to a local UTM zone and I just want one local watershed and all the data associated with that. But passing in this vector causes the bounds to be the full -180 to 180 -90 to 90 lat/lng coordinates then that bounding box doesn't correctly project to some local utm zone and the user experiences a confusing out of bounds error on the bounding box. The correct thing to do is only consider the bounding box of the selected features which is what this thing does.

PR #175, opened 2021-04-12
fixes #174 adding MFD support for detect outlets, a test case to cove…

### LINKED ISSUE #174: detect outlets for d8 and mfd
The detect outlets in pygeoprocessing is incorrectly named `detect_outlets` even though it is for d8 only, change to `d8` and provide a function for mfd.

richpsharp:
This PR adds  MFD support to outlet detection and modifies the function signature by adding a 3rd parameter to identify the mode as either 'd8' or 'mfd'. Added a test case to support MFD mode which was based off the D8 test but D8 manually converted to MFD.

I edited the `HISTORY.rst` file rather than adding a new entry since this feature hasn't been released yet.

I also removed several unneeded `fill_value_list` entries from routing that only filled the raster with nodata which is the default value anyway, but would take extra computation time.

I also removed a very unhelpful `LOGGER.debug('target_array_nodata_mask')` line that would get dumped when `stitch_rasters` was being invoked. I think it was debugging left over from development.

### DIFF
--- a/src/pygeoprocessing/routing/routing.pyx
+++ b/src/pygeoprocessing/routing/routing.pyx
@@ -4137,7 +4137,7 @@ def calculate_subwatershed_boundary(
 
 def detect_outlets(
         flow_dir_raster_path_band, flow_dir_type, target_outlet_vector_path):
-    """Create point vector indicating D8 outlets.
+    """Create point vector indicating flow raster outlets.
 
     Args:
         flow_dir_raster_path_band (tuple): raster path/band tuple
--- a/src/pygeoprocessing/routing/routing.pyx
+++ b/src/pygeoprocessing/routing/routing.pyx
@@ -4181,9 +4181,9 @@ def detect_outlets(
 
     raster_x_size, raster_y_size = raster_info['raster_size']
 
-    d8_flow_dir_raster = gdal.OpenEx(
+    flow_dir_raster = gdal.OpenEx(
         flow_dir_raster_path_band[0], gdal.OF_RASTER)
-    d8_flow_dir_band = d8_flow_dir_raster.GetRasterBand(
+    flow_dir_band = flow_dir_raster.GetRasterBand(
         flow_dir_raster_path_band[1])
 
     if raster_info['projection_wkt']:
@@ -4258,10 +4258,11 @@ def detect_outlets(
             flow_dir_block[-1, :] = flow_dir_nodata
 
         # Read iterblock plus a possible margin on top/bottom/left/right side
+        # and read as type int32 to handle both d8 or mfd formats
         flow_dir_block[
             1-y_off_border:win_ysize+1+win_ysize_border,
             1-x_off_border:win_xsize+1+win_xsize_border] = \
-            d8_flow_dir_band.ReadAsArray(
+            flow_dir_band.ReadAsArray(
                 xoff=xoff-x_off_border,
                 yoff=yoff-y_off_border,
                 win_xsize=win_xsize+win_xsize_border+x_off_border,
@@ -4332,13 +4333,12 @@ def detect_outlets(
                     outlet_feature = None
                     outlet_point = None
 
-
-    d8_flow_dir_raster = None
-    d8_flow_dir_band = None
+    flow_dir_raster = None
+    flow_dir_band = None
     LOGGER.info('outlet detection: 100% complete -- committing transaction')
     outlet_layer.CommitTransaction()
     outlet_layer = None
-    outlet_vector =  None
+    outlet_vector = None
     LOGGER.info('outlet detection: done')
 
 
--- a/src/pygeoprocessing/routing/routing.pyx
+++ b/src/pygeoprocessing/routing/routing.pyx
@@ -4141,9 +4141,11 @@ def detect_outlets(
 
     Args:
         flow_dir_raster_path_band (tuple): raster path/band tuple
-            of type uint8  indicating D8 flow direction created by
-            `routing.flow_dir_d8`.
-        flow_dir_type (str): one of 'd8' or 'mfd'
+            indicating D8 or MFD flow direction created by
+            `routing.flow_dir_d8` or `routing.flow_dir_mfd`.
+        flow_dir_type (str): one of 'd8' or 'mfd' to indicate the
+            ``flow_dir_raster_path_band`` is either a D8 or MFD flow
+            direction raster.
         target_outlet_vector_path (str): path to a vector that is created
             by this call that will be in the same projection units as the
             raster and have a point feature in the center of each pixel that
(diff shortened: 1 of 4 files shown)


richpsharp:
Added a test to ensure that a non d8 or mfd mode would be detected and exception raised.

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -4251,14 +4258,16 @@ def detect_outlets(d8_flow_dir_raster_path_band, target_outlet_vector_path):
...
+        # and read as type int32 to handle both d8 or mfd formats
+        flow_dir_block[
+            1-y_off_border:win_ysize+1+win_ysize_border,
+            1-x_off_border:win_xsize+1+win_xsize_border] = \
+            flow_dir_band.ReadAsArray(
                 xoff=xoff-x_off_border,
                 yoff=yoff-y_off_border,
                 win_xsize=win_xsize+win_xsize_border+x_off_border,
-                win_ysize=win_ysize+win_ysize_border+y_off_border,
-                buf_obj=flow_dir_block[
```
emlys, line 4259: did you mean to take out the `buf_obj` as well?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -4273,15 +4282,35 @@ def detect_outlets(d8_flow_dir_raster_path_band, target_outlet_vector_path):
...
+                        if flow_dir&(0xF<<(n_dir*4)) == 0:
+                            continue
+                        flow_dir_n = flow_dir_block[
+                            yi+D8_YOFFSET[n_dir],
+                            xi+D8_XOFFSET[n_dir]]
+                        if flow_dir_n != flow_dir_nodata:
+                            is_outlet = 0
+                            break
 
                 # if the outflow pixel is outside the raster boundaries or
```
emlys, line 4323: maybe don't need this comment here anymore?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -4139,13 +4135,17 @@ def calculate_subwatershed_boundary(
...
     Args:
-        d8_flow_dir_raster_path_band (tuple): raster path/band tuple
-            of type uint8  indicating D8 flow direction created by
-            `routing.flow_dir_d8`.
+        flow_dir_raster_path_band (tuple): raster path/band tuple
+            indicating D8 or MFD flow direction created by
+            `routing.flow_dir_d8` or `routing.flow_dir_mfd`.
+        flow_dir_type (str): one of 'd8' or 'mfd' to indicate the
+            ``flow_dir_raster_path_band`` is either a D8 or MFD flow
+            direction raster.
```
emlys, line 4152: maybe expand this to say something like, "if MFD, a pixel is an outlet if any proportion > 0 flows off the raster"? That's probably obvious to hydrology folks but I wasn't initially sure if there should be a proportion threshold or something.

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -4273,15 +4282,35 @@ def detect_outlets(d8_flow_dir_raster_path_band, target_outlet_vector_path):
...
+                    # is a nodata pixel it must mean xi,yi is an outlet
+                    if flow_dir_n != flow_dir_nodata:
+                        is_outlet = 0
+                else:
+                    # inspect all the outflow pixel neighbors in MFD mode
+                    for n_dir in range(8):
+                        # shift the 0xF mask to the outflow direction and
+                        # test if there's any outflow or not. 0 means nothing
+                        # flows out of that pixel in the ``n_dir`` direction.
+                        if flow_dir&(0xF<<(n_dir*4)) == 0:
```
emlys, line 4314: Perhaps this is worth another comment elaborating on how this works? it took me a while to understand:
`0xF<<(n_dir*4)` makes a binary number that has `1111` in the `n_dir`th position of the 32-bit `flow_dir` value. So if all of the 4 flow dir bits are `0`, the bitwise `&` will return `0`.

Question for my own understanding: each of the 8 directions gets its proportion represented in 4 bits. Does that mean we can only represent flow direction in 1/16th = 6.25% increments? Does `1111` mean 100% goes in that direction?

emlys (commented):
looks good! Just had a question and a couple of suggestions for comments.

src/pygeoprocessing/routing/routing.pyx
richpsharp: I think I did? Since we have to do a cast now I don't date pass a raw reference to flow dir block, so that's why I set it directly up on line 4262.

src/pygeoprocessing/routing/routing.pyx
richpsharp: I thought it was still relevant since we're about to enter the conditional of we're an outlet, but maybe we want to know the conditions of why it was an outlet? Is it okay if I leave it in? Or can you say more about why I should take it out?

src/pygeoprocessing/routing/routing.pyx
richpsharp: I added more description in 70b009e7cded905960f645b32c1ecf496bf32d29 but the only way MFD will have a flow direction to a nodata pixel or off the edge of the raster is if there is no other pixel to drain to. It'd be a bug if it proportionally drained to one pixel one way and off the edge of the raster.

src/pygeoprocessing/routing/routing.pyx
richpsharp: Sure, I expanded on that comment from your comment above in 59d8f2715dde0f66789dcc8d97984cb5918897c0

And re your question about max precision, you're close -- the value in the n-th 0xF place is the proportional flow in that direction so you don't know how much really flows in that without knowing how much flows in all other directions. Here's some examples, and assume pixel directions are laid out like this:

321
4x0
567

Then here's some select MFD pixel values and their proportional flow directions
* 0x00000000 -- nodata
* 0x00000001 - 1 proportional unit of flow in the 0 direction with a total sum of 1 means 1/1 -- 100% flows in direction 0
* 0x10000001 - 1 proportional unit of flow in the 0 direction and 1 in in the 7th, so direction 0 gets 1/2 of the flow and direction 7 gets 1/2
* 0xF0000001 - 1 proportional unit of flow in the 0 direction and 15 in in the 7th, so direction 0 gets 1/16 of the flow and direction 7 gets 15/16
* 0xFFFFFFF1 - 1 proportional unit of flow in the 0 direction and 15 in in ALL other directions. So direction 0 gets a proportion of 1/106 (0.94% increment) and all other directions get 15/106 (14.15% increment). This also shows that the smallest increment possible with this encoding scheme is 1/106. 

We thought that wasn't too bad of a precision with also being able to otherwise encode a complicated flow direction in a 32 bit int.

emlys: Thanks for the explanation! That's really cool that you can store 8 numbers in one.

src/pygeoprocessing/routing/routing.pyx
emlys: Ah I think I said that because there's the same comment above on lines 4296-4297. But it does kinda make sense both places.

emlys (approved):

PR #178, opened 2021-04-20
add to convolve_2d docstring

emlys:
Expanding a docstring to clarify a feature that @phargogh and I were discussing yesterday. It wasn't apparent to me until I tried it out that it would ignore nodata and edges in the average, rather than dividing by the constant `kernel_sum`.

### DIFF
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -2506,7 +2506,11 @@ def convolve_2d(
             values in these areas will be nonsensical numbers, perhaps
             numerical infinity or NaNs.
         normalize_kernel (boolean): If true, the result is divided by the
-            sum of the kernel.
+            sum of the kernel. If ``ignore_nodata_and_edges`` is also true,
+            this divides by the sum of the area of the kernel that is valid 
+            (not nodata or edge). For instance, a 3x3 kernel centered on a 
+            corner pixel would sum the 4 pixels that are within the raster 
+            and divide by 4.This combination is useful for averaging a raster.
         mask_nodata (boolean): If true, ``target_path`` raster's output is
             nodata where ``signal_path_band``'s pixels were nodata. Note that
             setting ``ignore_nodata_and_edges`` to ``True`` while setting
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -2477,6 +2477,18 @@ def convolve_2d(
     Nodata values are treated as 0.0 during the convolution and masked to
     nodata for the output result where ``signal_path`` has nodata.
 
+    Note with default values, boundary effects can be seen in the result where
+    the kernel would hang off the edge of the raster or in regions with 
+    nodata pixels. The function would treat these areas as values with "0.0"
+    by default thus pulling the total convolution down in these areas. This
+    is similar to setting ``mode='same'`` in Numpy's ``convolve`` function:
+    https://numpy.org/doc/stable/reference/generated/numpy.convolve.html
+
+    This boundary effect can be avoided by setting 
+    ``ignore_nodata_and_edges=True`` which normalizes the target result by 
+    dynamically accounting for the number of valid signal pixels the kernel 
+    overlapped during the convolution step.
+
     Args:
         signal_path_band (tuple): a 2 tuple of the form
             (filepath to signal raster, band index).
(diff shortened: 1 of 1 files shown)


src/pygeoprocessing/geoprocessing.py
```diff
@@ -2506,7 +2506,11 @@ def convolve_2d(
             values in these areas will be nonsensical numbers, perhaps
             numerical infinity or NaNs.
         normalize_kernel (boolean): If true, the result is divided by the
-            sum of the kernel.
+            sum of the kernel. If ``ignore_nodata_and_edges`` is also true,
+            this divides by the sum of the area of the kernel that is valid 
+            (not nodata or edge). For instance, a 3x3 kernel centered on a 
+            corner pixel would sum the 4 pixels that are within the raster 
+            and divide by 4.This combination is useful for averaging a raster.
```
richpsharp: That behavior description is correct, but it's not related to whether  `normalize_kernel` is True or False. `normalize_kernel` is only about "does your kernel sum to 1.0?" Otherwise what you describe is the functionality of `ignore_nodata_and_edges` which makes me think it should go there, but I feel like the first sentence says that it does this already:

``` 
ignore_nodata_and_edges (boolean): If true, any pixels that are equal
            to ``signal_path_band``'s nodata value or signal pixels where the
            kernel extends beyond the edge of the raster are not included when
            averaging the convolution filter.
```

But I like this idea of "how should I use these confusing parameters in this confusing function?" Would you be willing to expand this PR into making a more helpful descriptive paragraph above the args? The function does so much more than this:

```
Convolves the raster in ``kernel_path_band`` over ``signal_path_band``.
    Nodata values are treated as 0.0 during the convolution and masked to
    nodata for the output result where ``signal_path`` has nodata.
```

I think describing when you'd want to use any combination of normalize_kernel, ignore_nodata_and_edges, and mask_nodata would be nice to get up front rather than having to read the arguments.

richpsharp (changes requested):
I like the idea of this, but I think this kind of description of how parameters interact would be better in the discussion above the `args` section, would you be willing to expand that section instead?

src/pygeoprocessing/geoprocessing.py
emlys: Sure, I guess I put it with `normalize_kernel` because that's the feature that seemed relevant for my use case of averaging the raster. Agree it would be better in the descriptive paragraph above the args. 
> Otherwise what you describe is the functionality of ignore_nodata_and_edges which makes me think it should go there, but I feel like the first sentence says that it does this already

When I read the docstring originally it wasn't clear to me if `ignore_nodata_and_edges` would apply to the normalizing behavior as well as the convolution. I looked at the code and saw that `normalize_kernel` [divides everything by `kernel_sum`](https://github.com/natcap/pygeoprocessing/blob/d8c7cab583d08d26f0d9230e7d0a12f944779526/src/pygeoprocessing/geoprocessing.py#L3372) and that [`kernel_sum` is a constant](https://github.com/natcap/pygeoprocessing/blob/d8c7cab583d08d26f0d9230e7d0a12f944779526/src/pygeoprocessing/geoprocessing.py#L3323) so I concluded it wouldn't do what I wanted. James pointed out that it actually does work but the way that happens is pretty cryptic.

> I think describing when you'd want to use any combination of normalize_kernel, ignore_nodata_and_edges, and mask_nodata would be nice to get up front rather than having to read the arguments.

That would be super helpful to have in the docstring! I don't have a good understanding of what you'd use the other combinations of options for, though.

richpsharp: Hi @emlys, just getting back to this, I'm sorry if you were going to do this directly, but it sounded like you might have wanted a suggestion so I modified your PR to include that narrative about what `ignore_nodata_and_edges` does, also included another sentence in the `ignore_nodata_and_edges` parameter description and reverted the description of `normalize_kernel`. If you're okay with this, please feel free to merge the PR directly (I've already "approved") or I'm very happy to do more review if you want to work on this more.

richpsharp (approved):

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2505,6 +2517,9 @@ def convolve_2d(
             over any areas except nodata holes, in this case the resulting
             values in these areas will be nonsensical numbers, perhaps
             numerical infinity or NaNs.
+                Setting this value to ``True`` is similar to the effect
+            created by setting ``mode`` to ``"valid"`` in Numpy's
+            convolve function: https://numpy.org/doc/stable/reference/generated/numpy.convolve.html
```
emlys: As I'm understanding it this conflicts with the description above... the docs for `"valid"` say _"The convolution product is only given for points where the signals overlap completely."_ So, the output of `numpy.convolve(..., mode="valid")` would be smaller than the input array, because edges don't overlap completely and so they're excluded.  But the output of `convolve_2d(..., ignore_nodata_and_edges=True)` is the same size as the input. I don't think there's an analogous option in numpy: `mode="same"` doesn't account for boundary effects.

emlys (commented):
Thank you @richpsharp for summarizing this so much better than I could! Just had a question about the analogous numpy function.

src/pygeoprocessing/geoprocessing.py
richpsharp: oh you're right! that's not a comparable functionality at all. okay i'm going to just delete this part.

richpsharp (approved):

PR #181, opened 2021-04-23
convolve_2d: assign a default nodata value as described in the docstring.

### LINKED ISSUE #179: convolve_2d does not set a nodata value
The docstring suggests:

>target_nodata (int/float): nodata value to set on output raster.
            If ``target_datatype`` is not gdal.GDT_Float64, this value must
            be set.  Otherwise defaults to the minimum value of a float32.

But that checking for Float64 and defaulting never actually happens. Instead there is:

`output_array[:] = target_nodata` 

~~which numpy handles by converting `None` to `nan`. And this is before the FFT. So those nans are part of the signal block and the result is an entire raster block of nan if there was a nodata pixel in the block.~~

~~This seems pretty straightforward to fix & test, but then there's some other conditions around datatype and nodata checking that seem very intentional, but that I don't fully understand:~~

```
if target_datatype is not gdal.GDT_Float64 and target_nodata is None:
        raise ValueError(
            "`target_datatype` is set, but `target_nodata` is None. "
            "`target_nodata` must be set if `target_datatype` is not "
            "`gdal.GDT_Float64`.  `target_nodata` is set to None.")
```
~~Why the exception for Float64? ~~

davemfish:
This fixes #179 

If `target_nodata` is left as `None` we default to the minimum float32. This seemed like the intention all along based on the docstring.

### DIFF
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -2553,6 +2553,8 @@ def convolve_2d(
             "`target_datatype` is set, but `target_nodata` is None. "
             "`target_nodata` must be set if `target_datatype` is not "
             "`gdal.GDT_Float64`.  `target_nodata` is set to None.")
+    if target_nodata is None:
+        target_nodata = float(numpy.finfo(numpy.float32).min)
 
     if ignore_nodata_and_edges and not mask_nodata:
         LOGGER.debug(
@@ -2773,10 +2775,7 @@ def convolve_2d(
             f"{os.path.basename(target_path)}")
 
     # set the nodata value from 0 to a reasonable value for the result
-    if target_nodata is None:
-        target_band.DeleteNoDataValue()
-    else:
-        target_band.SetNoDataValue(target_nodata)
+    target_band.SetNoDataValue(target_nodata)
 
     target_band = None
     target_raster = None
--- a/tests/test_geoprocessing.py
+++ b/tests/test_geoprocessing.py
@@ -4166,8 +4166,9 @@ class TestGeoprocessing(unittest.TestCase):
         target_array = pygeoprocessing.raster_to_numpy_array(target_path)
         target_nodata = pygeoprocessing.get_raster_info(
             target_path)['nodata'][0]
+        self.assertIsNotNone(target_nodata)
 
-        expected_output = numpy.empty(signal_array.shape, numpy.float32)
+        expected_output = numpy.empty(signal_array.shape, numpy.float64)
         expected_output[:] = target_nodata
         expected_output[0, 0] = 1
         expected_output[-1, -1] = 1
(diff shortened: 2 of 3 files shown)


src/pygeoprocessing/geoprocessing.py
```diff
@@ -2553,6 +2553,8 @@ def convolve_2d(
             "`target_datatype` is set, but `target_nodata` is None. "
             "`target_nodata` must be set if `target_datatype` is not "
             "`gdal.GDT_Float64`.  `target_nodata` is set to None.")
+    if target_nodata is None:
+        target_nodata = float(numpy.finfo(numpy.float32).min)
```
davemfish, line 2557: Any thoughts on the best way to get this value? Without casting to `float` gdal would raise a type error on `double` when it tried to use this value to. 

Also, I figured using the minimum float64 would be a good idea because we know the output raster will be `GDT_Float64`. And while gdal did _not_ complain about type `double` in that case, it _did_ read the data back in as `-Inf`.

richpsharp: Hi Dave! I haven't tested it but this looks like what I'd expect to get the min float 32 value. Are you saying it doesn't work or something? Otherwise it looks great to me.

tests/test_geoprocessing.py
```diff
@@ -4166,8 +4166,9 @@ def test_convolve_2d_nodata(self):
         target_array = pygeoprocessing.raster_to_numpy_array(target_path)
         target_nodata = pygeoprocessing.get_raster_info(
             target_path)['nodata'][0]
+        self.assertIsNotNone(target_nodata)
```
richpsharp: Does it make sense to assert that it is equal to float32 min since that's what the docstring says it should be?

richpsharp (changes requested):
Hi Dave! Here's my review:

* unless I misunderstand it, I think the way you set the float32 min is correct in geoprocessing. 
* can you test for the explicit nodata value in the test case?
* can you update the HISTORY file to indicate this was a user facing bug that was fixed?

src/pygeoprocessing/geoprocessing.py
davemfish: @richpsharp , no it works. I just thought there might be a less-verbose way that maybe avoided the need to cast to python's `float`.

And the 2nd question was, why not use the min float64 value since this is a `GDT_Float64` raster? But, I'm happy using the min float32 value, it's certainly small enough.

tests/test_geoprocessing.py
davemfish: Done!

davemfish:
> Hi Dave! Here's my review:
> 
> * unless I misunderstand it, I think the way you set the float32 min is correct in geoprocessing.
> * can you test for the explicit nodata value in the test case?
> * can you update the HISTORY file to indicate this was a user facing bug that was fixed?

Thanks, @richpsharp ! I addressed these items.

richpsharp (approved):
GOOD PR Dave!

PR #184, opened 2021-05-05
Support Python 3.9

phargogh:
This PR verifies compatibility with Python 3.9, which we appear to get for free ... no real changes needed.

The only nontrivial change in this PR is the removal of the natcap PyPI, which was used for a time for GEOS builds that were compatible with the conda-forge GDAL builds (for some reason the conda-forge packages weren't compatible on Windows).  So the build process is a bit simpler as a result.

The question of which python versions are really needed for InVEST is still outstanding.


richpsharp (approved):
GOOD PR

PR #185, opened 2021-05-05
Bugfix/dist to channel no channels 171

### LINKED ISSUE #171: distance_to_channel_mfd when there are no channels
`distance_to_channel_mfd`, should handle the case where the channel raster contains only zeros (no channel anywhere). `distance_to_channel_d8` appears to set all pixels to the nodata value in this case. I'm not sure how `distance_to_channel_mfd` gets distance values in this case, but they should probably be set to nodata:
```
import pygeoprocessing
import pygeoprocessing.routing
import numpy
from osgeo import osr

srs = osr.SpatialReference()
srs.ImportFromEPSG(3857)
projection_wkt = srs.ExportToWkt()

flow_dir = numpy.ones((10, 10))
streams = numpy.zeros((10, 10))

flow_dir_path = '/Users/emily/Documents/test_stream_distance_flow_dir.tif'
streams_path = '/Users/emily/Documents/test_stream_distance_streams.tif'
distance_path = '/Users/emily/Documents/test_stream_distance_output.tif'
pygeoprocessing.numpy_array_to_raster(flow_dir, -1, (10, -10), (1000, 1000), projection_wkt, flow_dir_path)
pygeoprocessing.numpy_array_to_raster(streams, -1, (10, -10), (1000, 1000), projection_wkt, streams_path)

pygeoprocessing.routing.distance_to_channel_mfd((flow_dir_path, 1), (streams_path, 1), distance_path)
print(pygeoprocessing.raster_to_numpy_array(distance_path))
pygeoprocessing.routing.distance_to_channel_d8((flow_dir_path, 1), (streams_path, 1), distance_path)
print(pygeoprocessing.raster_to_numpy_array(distance_path))
```
```
$ python test_distance.py 
[[9. 8. 7. 6. 5. 4. 3. 2. 1. 0.]
 [9. 8. 7. 6. 5. 4. 3. 2. 1. 0.]
 [9. 8. 7. 6. 5. 4. 3. 2. 1. 0.]
 [9. 8. 7. 6. 5. 4. 3. 2. 1. 0.]
 [9. 8. 7. 6. 5. 4. 3. 2. 1. 0.]
 [9. 8. 7. 6. 5. 4. 3. 2. 1. 0.]
 [9. 8. 7. 6. 5. 4. 3. 2. 1. 0.]
 [9. 8. 7. 6. 5. 4. 3. 2. 1. 0.]
 [9. 8. 7. 6. 5. 4. 3. 2. 1. 0.]
 [9. 8. 7. 6. 5. 4. 3. 2. 1. 0.]]
[[-1. -1. -1. -1. -1. -1. -1. -1. -1. -1.]
 [-1. -1. -1. -1. -1. -1. -1. -1. -1. -1.]
 [-1. -1. -1. -1. -1. -1. -1. -1. -1. -1.]
 [-1. -1. -1. -1. -1. -1. -1. -1. -1. -1.]
 [-1. -1. -1. -1. -1. -1. -1. -1. -1. -1.]
 [-1. -1. -1. -1. -1. -1. -1. -1. -1. -1.]
 [-1. -1. -1. -1. -1. -1. -1. -1. -1. -1.]
 [-1. -1. -1. -1. -1. -1. -1. -1. -1. -1.]
 [-1. -1. -1. -1. -1. -1. -1. -1. -1. -1.]
 [-1. -1. -1. -1. -1. -1. -1. -1. -1. -1.]]
```
When testing with a Gura flow direction raster and a raster filled with zeros covering the same area, I got distance values in the millions.

richpsharp:
Fixed a bug in distance to channel MFD so that if the flow plain never reaches a channel the resulting distance raster is all nodata in that area, similar to D8 MFD.

This issue was caused by MFD tracking flows toward the channel, rather than channel up, but then not having a method to test if it never reached a channel. This was fixed by defaulting the distance to be nodata then tracking which cells have been visited in a separate raster rather than inferring it by a nodata value in the distance raster. 

Added a test case from @emlys in issue #171 to ensure correctness of this fix.

### DIFF
--- a/src/pygeoprocessing/routing/routing.pyx
+++ b/src/pygeoprocessing/routing/routing.pyx
@@ -2611,7 +2611,7 @@ def distance_to_channel_mfd(
                 "%s is supposed to be a raster band tuple but it's not." % (
                     path))
 
-    distance_nodata = IMPROBABLE_FLOAT_NODATA
+    distance_nodata = -1
     pygeoprocessing.new_raster_from_base(
         flow_dir_mfd_raster_path_band[0],
         target_distance_to_channel_raster_path,
@@ -2619,9 +2619,19 @@ def distance_to_channel_mfd(
         raster_driver_creation_tuple=raster_driver_creation_tuple)
     distance_to_channel_managed_raster = _ManagedRaster(
         target_distance_to_channel_raster_path, 1, 1)
-
     channel_managed_raster = _ManagedRaster(
         channel_raster_path_band[0], channel_raster_path_band[1], 0)
+
+    tmp_work_dir = tempfile.mkdtemp(
+        suffix=None, prefix='dist_to_channel_mfd_work_dir',
+        dir=os.path.dirname(target_distance_to_channel_raster_path))
+    visited_raster_path = os.path.join(tmp_work_dir, 'visited.tif')
+    pygeoprocessing.new_raster_from_base(
+        flow_dir_mfd_raster_path_band[0],
+        visited_raster_path,
+        gdal.GDT_Byte, [0],
+        raster_driver_creation_tuple=raster_driver_creation_tuple)
+    visited_managed_raster = _ManagedRaster(visited_raster_path, 1, 1)
 
     flow_dir_mfd_managed_raster = _ManagedRaster(
         flow_dir_mfd_raster_path_band[0], flow_dir_mfd_raster_path_band[1], 0)
@@ -2702,10 +2712,10 @@ def distance_to_channel_mfd(
                     # nodata flow, so we skip
                     continue
 
-                if _is_close(distance_to_channel_managed_raster.get(
-                        xi_root, yi_root), distance_nodata, 1e-8, 1e-5):
+                if visited_managed_raster.get(xi_root, yi_root) == 0:
+                    visited_managed_raster.set(xi_root, yi_root, 1)
                     distance_to_channel_stack.push(
-                        FlowPixelType(xi_root, yi_root, 0, 0.0))
+                        FlowPixelType(xi_root, yi_root, 0, -1))
 
                 while not distance_to_channel_stack.empty():
                     pixel = distance_to_channel_stack.top()
@@ -2735,16 +2745,22 @@ def distance_to_channel_mfd(
                                 yi_n < 0 or yi_n >= raster_y_size):
                             continue
 
-                        n_distance = distance_to_channel_managed_raster.get(
-                            xi_n, yi_n)
 
-                        if _is_close(n_distance, distance_nodata, 1e-8, 1e-5):
+                        if visited_managed_raster.get(xi_n, yi_n) == 0:
+                            visited_managed_raster.set(xi_n, yi_n, 1)
                             preempted = 1
                             pixel.last_flow_dir = i_n
                             distance_to_channel_stack.push(pixel)
                             distance_to_channel_stack.push(
-                                FlowPixelType(xi_n, yi_n, 0, 0.0))
+                                FlowPixelType(xi_n, yi_n, 0, -1))
                             break
+
+                        n_distance = distance_to_channel_managed_raster.get(
+                            xi_n, yi_n)
+
+                        if n_distance == distance_nodata:
+                            # a channel was never found
+                            continue
 
                         # if a weight is passed we use it directly and do
                         # not consider that a diagonal pixel is further
@@ -2758,6 +2774,8 @@ def distance_to_channel_mfd(
                         else:
                             weight_val = (SQRT2 if i_n % 2 else 1)
 
+                        if pixel.value == -1:
+                            pixel.value = 0
                         pixel.value += flow_dir_weight * (
                             weight_val + n_distance)
 
@@ -2781,6 +2799,7 @@ def distance_to_channel_mfd(
     flow_dir_mfd_managed_raster.close()
     if weight_raster is not None:
         weight_raster.close()
+    shutil.rmtree(tmp_work_dir)
     LOGGER.info('%.1f%% complete', 100.0)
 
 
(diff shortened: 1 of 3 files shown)


tests/test_routing.py
```diff
@@ -1163,3 +1163,39 @@ def test_extract_straher_streams_watersheds_d8(self):
...
+    def test_channel_not_exist_distance(self):
+        """PGP.routing: test for nodata result if channel doesn't exist."""
+        from osgeo import osr
+        srs = osr.SpatialReference()
+        srs.ImportFromEPSG(3857)
+        projection_wkt = srs.ExportToWkt()
+
+        flow_dir = numpy.ones((10, 10))
+        streams = numpy.zeros((10, 10))
+        expected_result = numpy.full((10, 10), -1)
```
emlys: Super minor thing! since the nodata value `-1` is used 3 times, you could make it a variable `nodata = -1` and reference that in each place to make the intent more clear.

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2702,10 +2712,10 @@ def distance_to_channel_mfd(
                     # nodata flow, so we skip
                     continue
 
-                if _is_close(distance_to_channel_managed_raster.get(
-                        xi_root, yi_root), distance_nodata, 1e-8, 1e-5):
+                if visited_managed_raster.get(xi_root, yi_root) == 0:
+                    visited_managed_raster.set(xi_root, yi_root, 1)
                     distance_to_channel_stack.push(
-                        FlowPixelType(xi_root, yi_root, 0, 0.0))
+                        FlowPixelType(xi_root, yi_root, 0, distance_nodata))
```
emlys, line 2722: it could be helpful to have a comment here briefly explaining what each of the 4 arguments are for

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2758,6 +2774,8 @@ def distance_to_channel_mfd(
                         else:
                             weight_val = (SQRT2 if i_n % 2 else 1)
 
+                        if pixel.value == distance_nodata:
+                            pixel.value = 0
                         pixel.value += flow_dir_weight * (
```
emlys, line 2783: Curious why the `flow_dir_weight` factors in to the distance calculation here?

emlys (approved):
This looks good and I'm sure it works! but honestly I'm really confused what's going on. I see you added a raster to keep track of which pixels have been visited, instead of determining that by comparing to `distance_nodata`. I guess I just don't understand the algorithm despite looking at it a whole lot. So I don't feel like I can give a very thorough review of this, but with the test I'm sure it's working. 

If you want to discuss it more and help me understand what's going on, that would be neat, but I'm happy to approve it as is.

src/pygeoprocessing/routing/routing.pyx
richpsharp: Sure, commented in c27358d91174e5458c512531a4f0d15637efd661

tests/test_routing.py
richpsharp: fixed in e760ce905e31e5bb5ddfff8f6943514d6efb109a

src/pygeoprocessing/routing/routing.pyx
richpsharp: Sure, so at this point in the algorithm this pixel has determined it has a neighbor it flows into that has its distance to channel defined, that means it will contribute to the current pixel, but since this is MFD it's possible the current pixel flows into other pixels as well. So by multiplying by flow_dir_weight we proportionally weight that downstream distance to how much it should get on this pixel and if it flows in more than one direction it'll pick up partial downstream distance from its other directions. Note later on line 2794 is when we finally divide by the sum of the weights so we get this proportional weighting correct.

richpsharp:
> This looks good and I'm sure it works! but honestly I'm really confused what's going on. I see you added a raster to keep track of which pixels have been visited, instead of determining that by comparing to `distance_nodata`. I guess I just don't understand the algorithm despite looking at it a whole lot. So I don't feel like I can give a very thorough review of this, but with the test I'm sure it's working.
> 
> If you want to discuss it more and help me understand what's going on, that would be neat, but I'm happy to approve it as is.

Sure, happy to explain more in person too if you're interested. This algorithm is tricky because we're scheduling from the top down with the general algorithm "if the downstream neighbors are defined, then we can define our own flow distance". So say we identify one neighbor downstream pixel that is not defined yet, we schedule it to be computed by pushing the current pixel on the processing stack and then the downstream pixel to get it processed next. Fine, we'd keep doing this until we got down to an outlet, then we'd pop off our stack and we'd recursively construct values on the way back up...

But say it's the case that caused this bug -- there is no outlet. Now when we pop back up how do we know if the downstream pixel hasn't been calculated yet, or it couldn't have been calculated just by looking at that pixel? Say our nodata value was -1 so we set it to -2 to indicate it was going to be processed, so then great, we're like oh no, it couldn't find an outlet so we set it to nodata -1 again... BUT then you could follow another flowpath later that connected to this pixel which is now -1 again and so you set it to -2 again and recalculate that entire flowpath again just to find it didn't connect to a channel and set it back to -1... I'm sure we could work out some algorithm to do this but I think the following is straightforward ---> 

Rather than try use the target distance raster as a multipurpose target value and intermediate processing status structure (we'd still have to develop that latter algorithm anyway) we use this new byte raster that merely marks if the pixel has been sent to be processed. Now if we encounter a downstream neighbor pixel that *has* been visited (easy to check thanks to our new raster) but the distance is nodata, then we know there is no downstream connecting flowpath and we set our pixel to nodata and bail. There's no extra work to make a special case for a missing channel raster!

Does that help any?

emlys:
Thanks @richpsharp that's really helpful!! I think I get it now. So the bug fix was to make sure that pixels that never reach an outlet get set to nodata, rather than to their distance to the raster edge. And the byte raster is just a neater/more efficient way to keep track of that.

src/pygeoprocessing/routing/routing.pyx
emlys: Oh interesting. I assumed it would be the shortest distance (following the shortest path of pixels that have any MFD proportion > 0). But it's really more like the average distance that water travels from that pixel to the outlets.

Issue #168, opened 2021-02-24
ensure convolve_2d warns if a raster is not square blocks


richpsharp:
Just unrolled a very very very slow data pipeline that was occuring because the base raster had a blocksize of 126000x1. Convolve 2d moves little windows around and this was thrashing the gdal cache in a way that it was the only operation happening. I didn't spot it except to `gdalinfo` it, so convolve 2d should warn if this might happen.

dcdenu4:
Interesting!

PR #187, opened 2021-05-06
fixes #177 raises an exception if transform goes out of bounds

### LINKED ISSUE #177: Warnings/Exception if bounding box is out of bounds on reprojection
Some reprojections can cause non-finite bounding box bounds as a result.... Try to warn the user against this?

richpsharp:
This PR adds a check to `transform_bounding_box` to ensure that the result is finite. If not this raises a helpful exception. Added a test case to cover this feature and noting change in history.

### DIFF
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -2973,6 +2973,11 @@ def transform_bounding_box(
     minx, maxx = sorted([raw_bounding_box[0], raw_bounding_box[2]])
     miny, maxy = sorted([raw_bounding_box[1], raw_bounding_box[3]])
     transformed_bounding_box = [minx, miny, maxx, maxy]
+    if not all(numpy.isfinite(numpy.array(transformed_bounding_box))):
+        raise ValueError(
+            f'some transformed coordinates are not finite: '
+            f'{transformed_bounding_box}, base bounding box may not fit into '
+            f'target coordinate projection system.')
     return transformed_bounding_box
 
 
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -2922,6 +2922,11 @@ def transform_bounding_box(
         fitting bounding box around the original warped bounding box in
         ``new_epsg`` coordinate system.
 
+    Raises:
+        ``ValueError`` if resulting transform yields non-finite coordinates.
+        This would indicate an ill posed transform region that the user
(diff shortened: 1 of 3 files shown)


src/pygeoprocessing/geoprocessing.py
```diff
@@ -2973,6 +2978,11 @@ def _transform_point(point):
     minx, maxx = sorted([raw_bounding_box[0], raw_bounding_box[2]])
     miny, maxy = sorted([raw_bounding_box[1], raw_bounding_box[3]])
     transformed_bounding_box = [minx, miny, maxx, maxy]
+    if not all(numpy.isfinite(numpy.array(transformed_bounding_box))):
+        raise ValueError(
+            f'some transformed coordinates are not finite: '
+            f'{transformed_bounding_box}, base bounding box may not fit into '
+            f'target coordinate projection system.')
```
dcdenu4, line 2985: Love `f` strings, but not necessary on the first and third line! But if you like them there in general, it doesn't really bother me.

tests/test_geoprocessing.py
```diff
@@ -2401,6 +2401,23 @@ def test_transform_box_utm_to_utm(self):
...
+        bounding_box_lat_lng_oob = [-45, 89, -43, 91]
+
+        target_srs = osr.SpatialReference()
+        target_srs.ImportFromEPSG(32619)  # UTM19N EPSG
+
+        with self.assertRaises(ValueError) as cm:
+            result = pygeoprocessing.transform_bounding_box(
+                bounding_box_lat_lng_oob, osr.SRS_WKT_WGS84_LAT_LONG,
+                target_srs.ExportToWkt())
+            print(result)
```
dcdenu4: Left over print?

dcdenu4 (changes requested):
Look good beside a few minor comments.

src/pygeoprocessing/geoprocessing.py
richpsharp: Similar comment to what I left on #168. James and I both like the consistent f down the line for aesthetic reasons. If you don't agree with that then lets chat more about it.

tests/test_geoprocessing.py
richpsharp: Removed in 26a7b14

richpsharp:
> Look good beside a few minor comments.

I addressed your comments!

dcdenu4 (approved):
Looks good!

PR #189, opened 2021-05-06
fixes #161 extra long runtimes on large subwatersheds with large coordinates

### LINKED ISSUE #161: watershed boundary doesn't terminate when coordinates are large
This is basically an issue of using `gdal.ApplyGeotransform` and then manually doing it inside of `calculate_subwatershed_boundary`. The former has high precision (double) and the latter has lower precision (float) because I just set it to float. The problem is that when the watershed is tracing itself the start and end coordinate don't quite match up within a reasonable tolerance because of the difference in this precision. The result for the user is that watersheds terminate after many steps and have many loops run around them.

richpsharp:
This PR fixes #161 by tracking the integer delta of number of x/y steps taken when tracing a watershed's boundary rather than keeping track of the start and end point in projected coordinates. 

I didn't make a note in history since this feature wasn't released yet. I also haven't included a specific test case for this since it would requires a LARGE dataset and would require a runtime larger than all other tests combined in pygeoprocessing.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -64,11 +64,6 @@ Unreleased Changes
 * Fixed an issue with ``convolve_2d`` that allowed output rasters to be
   created without a defined nodata value.
 * Fixed a LOGGER message bug that occurred in ``zonal_statistics``.
-* Fixed an issue in ``calculate_subwatershed_boundary`` that would cause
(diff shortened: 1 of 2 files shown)


phargogh (approved):
I like it!  Using int pixel indices to store and check the delta will not only fix the stated bug, but would probably be mildly faster as well.

PR #191, opened 2021-05-09
Feature/force single drainage point 190

### LINKED ISSUE #190: Force a single drainage point for `fill_pits`
By default `fill_pits` treats pixels on the edge of a raster or that abut nodata that don't otherwise border a lower pixel to be a drain. This can lead to cases where a predetermined watershed polygon can result in a DEM that drains little tiny watersheds around the edge of the polygon. The biggest error with this is a single pixel in the Nile watershed which splits the nile so it is only 1/3 as long as it should be:

image

This is not an issue that can be fixed by burning streams or any other method except to force a single outlet to a watershed. This this issue proposes two enhancements:
* Add an optional argument to "fill pits" which takes an x/y raster coordinate to indicate that is the only drain in the watershed
* helper function in `pygeoprocessing.routing` to find either the lowest dem pixel that drains off the nodata/the edge of the raster OR the lowest pixel in the raster for cases that may be a large depression.

richpsharp:
This feature fixes #190 by adding a parameter to `fill_pits` called `single_outlet_tuple`. If present this forces the fill to treat that pixel as the only valid drain. Combined with another feature that `max_pixel_fill_count` can be set to -1 to fill non-stop this will force every non-nodata pixel in the raster to drain to the pixel targed in `single_outlet_tuple`. To help identify a candidate I've also added another function `detect_lowest_drain_and_sink` which returns the lowest edge drain and internal sink pixels that could be used in tandem with `single_outlet_tuple`. Here's all the changes:
* added new `single_outlet_tuple` parameter to `fill_pits`
* created new helper function `detect_lowest_drain_and_sink`
* fixed a bug in `fill_pits` that would have filled up the nodata region of a raster in a corner case
* added functionality that if `max_pixel_fill_count` is -1 it will not terminate early and set that to the default parameter for `fill_pits`
* added test cases to cover the above
* updated history describing the above

### DIFF
--- a/src/pygeoprocessing/routing/routing.pyx
+++ b/src/pygeoprocessing/routing/routing.pyx
@@ -4245,7 +4245,7 @@ def detect_lowest_drain_and_sink(dem_raster_path_band):
         (drain_pixel, drain_height, sink_pixel, sink_height) -
             two (x, y) tuples with corresponding heights, first
             list is for edge drains, the second is for pit sinks. The x/y
-            coordinate is in raster coordinate space and `height` is the
+            coordinate is in raster coordinate space and ``height`` is the
             height of the given pixels in edge and pit respectively.
     """
     # this outer loop drives the raster block search
--- a/src/pygeoprocessing/routing/routing.pyx
+++ b/src/pygeoprocessing/routing/routing.pyx
@@ -612,6 +612,7 @@ def _generate_read_bounds(offset_dict, raster_x_size, raster_y_size):
 def fill_pits(
         dem_raster_path_band, target_filled_dem_raster_path,
         working_dir=None, long max_pixel_fill_count=_MAX_PIXEL_FILL_COUNT,
+        single_outlet_tuple=None,
         raster_driver_creation_tuple=DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS):
     """Fill the pits in a DEM.
 
@@ -638,6 +639,11 @@ def fill_pits(
         max_pixel_fill_count (int): maximum number of pixels to fill a pit
             before leaving as a depression. Useful if there are natural
             large depressions.
+        single_outlet_tuple (tuple): If not None, this is an x/y tuple in
+            raster coordinates indicating the only pixel that can be
+            considered a drain. If None then any pixel that would drain to
+            the edge of the raster or a nodata hole will be considered a
+            drain.
         raster_driver_creation_tuple (tuple): a tuple containing a GDAL driver
             name string as the first element and a GDAL creation options
             tuple/list as the second. Defaults to a GTiff driver tuple
@@ -695,6 +701,13 @@ def fill_pits(
     # algorithm, it's written into the mask rasters to indicate which pixels
     # have already been processed
     cdef int feature_id
+
+    # used to handle the case for single outlet mode
+
+    cdef int single_outlet=0, outlet_x=-1, outlet_y=-1
+    if single_outlet_tuple is not None:
+        single_outlet = 1
+        outlet_x, outlet_y = single_outlet_tuple
 
     # used for time-delayed logging
     cdef time_t last_log_time
@@ -808,19 +821,35 @@ def fill_pits(
                         xi_root, yi_root) != mask_nodata:
                     continue
 
+                # a single outlet trivially drains
+                if (single_outlet and
+                        xi_root == outlet_x and
+                        yi_root == outlet_y):
+                    continue
+
                 # search neighbors for downhill or nodata
                 downhill_neighbor = 0
                 nodata_neighbor = 0
                 for i_n in range(8):
                     xi_n = xi_root+D8_XOFFSET[i_n]
                     yi_n = yi_root+D8_YOFFSET[i_n]
+
+                    if single_outlet:
+                        if (xi_n == outlet_x and yi_n == outlet_y):
+                            downhill_neighbor = 1
+                            break
                     if (xi_n < 0 or xi_n >= raster_x_size or
                             yi_n < 0 or yi_n >= raster_y_size):
-                        # it'll drain off the edge of the raster
-                        nodata_neighbor = 1
-                        break
+                        if not single_outlet:
+                            # it'll drain off the edge of the raster
+                            nodata_neighbor = 1
+                            break
+                        else:
+                            # continue so we don't access out of bounds
+                            continue
                     n_height = filled_dem_managed_raster.get(xi_n, yi_n)
-                    if _is_close(n_height, dem_nodata, 1e-8, 1e-5):
+                    if not single_outlet and _is_close(
+                            n_height, dem_nodata, 1e-8, 1e-5):
                         # it'll drain to nodata
                         nodata_neighbor = 1
                         break
@@ -861,18 +890,27 @@ def fill_pits(
                     for i_n in range(8):
                         xi_n = xi_q+D8_XOFFSET[i_n]
                         yi_n = yi_q+D8_YOFFSET[i_n]
+
                         if (xi_n < 0 or xi_n >= raster_x_size or
                                 yi_n < 0 or yi_n >= raster_y_size):
-                            natural_drain_exists = 1
+                            if not single_outlet:
+                                natural_drain_exists = 1
                             continue
+
+                        if single_outlet:
+                            if xi_n == outlet_x and yi_n == outlet_y:
+                                natural_drain_exists = 1
+                                continue
                         n_height = filled_dem_managed_raster.get(
                             xi_n, yi_n)
-                        if _is_close(n_height, dem_nodata, 1e-8, 1e-5):
+                        if not single_outlet and _is_close(
+                                n_height, dem_nodata, 1e-8, 1e-5):
                             natural_drain_exists = 1
                             continue
                         if n_height < center_val:
                             natural_drain_exists = 1
                             continue
+
                         if flat_region_mask_managed_raster.get(
                                 xi_n, yi_n) == 1:
                             # been set before on a previous iteration, skip
@@ -922,11 +960,20 @@ def fill_pits(
                     for i_n in range(8):
                         xi_n = xi_q+D8_XOFFSET[i_n]
                         yi_n = yi_q+D8_YOFFSET[i_n]
+
                         if (xi_n < 0 or xi_n >= raster_x_size or
                                 yi_n < 0 or yi_n >= raster_y_size):
                             # drain off the edge of the raster
-                            pour_point = 1
-                            break
+                            if not single_outlet:
+                                pour_point = 1
+                                break
+                            else:
+                                continue
+
+                        if single_outlet:
+                            if xi_n == outlet_x and yi_n == outlet_y:
+                                pour_point = 1
+                                break
 
                         if pit_mask_managed_raster.get(
                                 xi_n, yi_n) == feature_id:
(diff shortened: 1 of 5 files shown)


src/pygeoprocessing/routing/routing.pyx
```diff
@@ -696,6 +700,13 @@ def fill_pits(
     # have already been processed
     cdef int feature_id
 
+    # used to handle the case for single outlet mode
+
```
dcdenu4: remove empty line to better relate the comment to the code?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -793,10 +804,10 @@ def fill_pits(
...
 
         # search block for locally undrained pixels
-        for yi in range(1, win_ysize+1):
-            yi_root = yi-1+yoff
-            for xi in range(1, win_xsize+1):
-                xi_root = xi-1+xoff
+        for yi in range(win_ysize):
+            yi_root = yi+yoff
+            for xi in range(win_xsize):
+                xi_root = xi+xoff
```
dcdenu4, line 809: I'm assuming this is where you "fixed a prior bug". Could you just comment why it was iterating from `1` before and why we don't want to do that now? I'd love to know the context for this fix / change.

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -850,8 +876,12 @@ def fill_pits(
                     xi_q = search_queue.front().xi
                     yi_q = search_queue.front().yi
                     search_queue.pop()
-                    if search_steps > max_pixel_fill_count:
+                    if (max_pixel_fill_count > 0 and
```
dcdenu4, line 878: Okay, so if `max_pixel_fill_count == -1`, we don't want to come in here and quit the search early. We are riding it out until we find that single drain point. (just 'typing' out loud)

dcdenu4: Well I guess if `single_outlet` is `None`, then we're just searching until we find an actual drain, be it nodata, an edge, or an outlet.

dcdenu4: Or if it's determined it should be filled.

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -4159,6 +4216,131 @@ def calculate_subwatershed_boundary(
...
+
+    Args:
+        dem_raster_path_band (tuple): a raster/path band tuple to detect
+            sinks in.
+
+    Return:
+        (drain_pixel, drain_height, sink_pixel, sink_height) -
+            two (x, y) tuples with corresponding heights, first
+            list is for edge drains, the second is for pit sinks. The x/y
+            coordinate is in raster coordinate space and `height` is the
```
dcdenu4: Should `height` be double ticks here?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -611,7 +607,9 @@ def _generate_read_bounds(offset_dict, raster_x_size, raster_y_size):
 
 def fill_pits(
         dem_raster_path_band, target_filled_dem_raster_path,
-        working_dir=None, long max_pixel_fill_count=_MAX_PIXEL_FILL_COUNT,
+        working_dir=None,
+        long long max_pixel_fill_count=-1,
```
dcdenu4, line 611: This is probably going to have an impact on InVEST models if the default was `500`. Is there an issue in the InVEST repo mentioning this change and the need to update the models?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -4159,6 +4216,131 @@ def calculate_subwatershed_boundary(
...
+            LOGGER.info(
+                '(infer_sinks): '
+                f'{current_pixel} of {raster_x_size * raster_y_size} '
+                'pixels complete')
+
+        # search block for locally undrained pixels
+        for yi in range(1, win_ysize+1):
+            yi_root = yi-1+yoff
+            for xi in range(1, win_xsize+1):
+                xi_root = xi-1+xoff
```
dcdenu4: Here we are doing the iteration with the 1 offset. Is this still intended given the change above in `fill_dem`?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -4159,6 +4216,131 @@ def calculate_subwatershed_boundary(
...
+                # this value is set in case it turns out to be the root of a
+                # pit, we'll start the fill from this pixel in the last phase
+                # of the algorithm
```
dcdenu4: Update this comment? I don't think we're doing any filling here.

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -4159,6 +4216,131 @@ def calculate_subwatershed_boundary(
...
+
+        if ctime(NULL) - last_log_time > _LOGGING_PERIOD:
+            last_log_time = ctime(NULL)
+            current_pixel = xoff + yoff * raster_x_size
+            LOGGER.info(
+                '(infer_sinks): '
+                f'{current_pixel} of {raster_x_size * raster_y_size} '
+                'pixels complete')
+
+        # search block for locally undrained pixels
```
dcdenu4: and sinks

tests/test_routing.py
```diff
@@ -1164,6 +1162,65 @@ def test_extract_straher_streams_watersheds_d8(self):
         watershed_vector = None
         watershed_layer = None
 
+    def test_single_drain_point(self):
+        """PGP.routing: test single_drain_point pitfill."""
+        dem_array = numpy.zeros((11, 11), dtype=numpy.float32)
+        dem_array[0, 0] = -1.0
+        dem_array[1:8, 1:8] = -2.0
+        dem_array[10, 7] = -4
```
dcdenu4: should this be `-4.0` for consistency?

tests/test_routing.py
```diff
@@ -1164,6 +1162,65 @@ def test_extract_straher_streams_watersheds_d8(self):
...
+        dem_array[8:11, 8:11] = 2.0
+        dem_array[9:11, 9:11] = 1.0
+        dem_array[10, 10] = -7.0
+        dem_path = os.path.join(self.workspace_dir, 'dem.tif')
+        _array_to_raster(dem_array, None, dem_path)
+
+        # outlet tuple at 0,0, just drain one edge
+        expected_array_0_0 = numpy.copy(dem_array)
+        expected_array_0_0[1:8, 1:8] = -1
```
dcdenu4, line 1179: Just my own curiosity here. A plateau of all `-1` will drain to an outlet of also `-1`? I guess I'd expect these to be ever so slightly greater than `-1`.

dcdenu4: By plateau I guess I mean pit in this instance.

tests/test_routing.py
```diff
@@ -1164,6 +1162,65 @@ def test_extract_straher_streams_watersheds_d8(self):
...
+        expected_array_0_0[10, 7] = 0
+        expected_array_0_0[8:11, 8:11] = 2.0
+
+        # output tuple at 5,5, it's a massive pit so drain the edges
+        expected_array_5_5 = numpy.copy(dem_array)
+        expected_array_5_5[8:11, 8:11] = 2.0
+        expected_array_5_5[10, 7] = 0
```
dcdenu4, line 1186: We're just testing the algorithm works by draining everything to Sarlacc Pit?

tests/test_routing.py
```diff
@@ -1164,6 +1162,65 @@ def test_extract_straher_streams_watersheds_d8(self):
...
+
+        for output_tuple, expected_array in [
+                ((0, 0), expected_array_0_0),
+                ((5, 5), expected_array_5_5),
+                ]:
```
dcdenu4: Could we have another test in here or separate that also manipulates the `max_fill_counter` variable using the same fake dem you've used here?

tests/test_routing.py
```diff
@@ -1164,6 +1162,65 @@ def test_extract_straher_streams_watersheds_d8(self):
...
+            fill_path = os.path.join(self.workspace_dir, 'filled.tif')
+            pygeoprocessing.routing.fill_pits(
+                (dem_path, 1), fill_path,
+                single_outlet_tuple=output_tuple,
+                working_dir=self.workspace_dir)
+            result_array = pygeoprocessing.raster_to_numpy_array(fill_path)
+            numpy.testing.assert_almost_equal(result_array, expected_array)
+
+    def test_detect_lowest_drain_and_sink(self):
+        """PGP>routing: test detect_lowest_sink_and_drain."""
```
dcdenu4: >PGP.routing

dcdenu4 (changes requested):
Hey @richpsharp I just gave this a look through. I followed it pretty well I think and didn't find anything glaring. I have a few comments though that should be minor.

src/pygeoprocessing/routing/routing.pyx
richpsharp: Sure, I made an issue here: https://github.com/natcap/invest/issues/560 won't affect any test data and it was a stopgap to only affect runs that were run on degenerate DEMs.

src/pygeoprocessing/routing/routing.pyx
richpsharp: removed in 5aaefd9

src/pygeoprocessing/routing/routing.pyx
richpsharp: I just noticed this loop ran from 1 to n+1 but then inside the parameter was always i-1. so just changed it from 0 to n then just use i. There's no functional change here.

src/pygeoprocessing/routing/routing.pyx
richpsharp: Correct.

src/pygeoprocessing/routing/routing.pyx
richpsharp: It wasn't referring to an actual parameter which is why i didn't have double ticks, just the suffix of those but I changed it in 0fd7e78 from `height` to _height so that's more obvious.

src/pygeoprocessing/routing/routing.pyx
richpsharp: changed in 02f5637

src/pygeoprocessing/routing/routing.pyx
richpsharp: changed in 40ad0cf

tests/test_routing.py
richpsharp: changed in c05182d

tests/test_routing.py
richpsharp: Yes because there's a -1 pixel that pours off the edge, so to drain you only need the other pixels to be at the same height.

tests/test_routing.py
richpsharp: GOOD ONE

tests/test_routing.py
richpsharp: I added two more cases that use a `max_pit_fill_count` of 1 that'll result in no pit filling at all. Added in 75a05f5

tests/test_routing.py
richpsharp: fixed in ff64743e1148d989d559e87db6c767465d130753

src/pygeoprocessing/routing/routing.pyx
richpsharp: Removed in 1e3f6bf

richpsharp:
> Hey @richpsharp I just gave this a look through. I followed it pretty well I think and didn't find anything glaring. I have a few comments though that should be minor.

Thanks @dcdenu4, I think I addressed all your comments. Can you take a second look when you get a chance?

dcdenu4 (approved):
Thanks Rich! Looks okay to me.

PR #200, opened 2021-05-25
Bugfix/revert to geom bounds

richpsharp:
Reverting create vector extents code to use geometries rather than layer bounds. Made update to history.


emlys (approved):
Thanks Rich, hooray for `git revert`!

PR #201, opened 2021-06-02
Expand transform_bounding_box error message

emlys:
Expanding this error message to include some more info would be helpful so we don't have to do it on the invest side every time `transform_bounding_box` is used. natcap/invest#557


richpsharp (approved):
Hi @emlys, could you also add a short mention of this in `HISTORY.rst` since it's a user facing change? Otherwise I've already approved it so feel free to do the merge once all the checks pass. (p.s. sorry I tried to see if I could add a line in HISTORY myself but i couldn't figure out how to edit a file you hadn't changed yet)

PR #202, opened 2021-06-07
fixes #199, skip if pixel is nodata and update test to cover a case l…

### LINKED ISSUE #199: distance_to_channel_mfd producing negative values
And they're not just the nodata value.
Minimal example to reproduce, using the flow direction and streams rasters from NDR test data intermediate outputs:
```
import numpy
import pygeoprocessing
import pygeoprocessing.routing

flow_dir_path = '/Users/emily/Documents/ndr_test_wksp/intermediate_outputs/flow_direction.tif'
streams_path = '/Users/emily/Documents/ndr_test_wksp/intermediate_outputs/stream.tif'
target_path = '/Users/emily/Documents/demo_distance_to_stream_bug.tif'

pygeoprocessing.routing.distance_to_channel_mfd(
    (flow_dir_path, 1),
    (streams_path, 1),
    target_path)

arr = pygeoprocessing.raster_to_numpy_array(target_path)
nodata = pygeoprocessing.get_raster_info(target_path)['nodata'][0]
valid_mask = ~numpy.isclose(arr, nodata)
print(numpy.where(arr[valid_mask] < 0))
print(arr[valid_mask][arr[valid_mask] < 0])
```
```
$ python demo.py 
(array([17765, 17766, 17792, 17847, 17848, 17849, 17850, 17851, 17862,
       17863, 17864, 17865]),)
[-0.07142857 -0.07142857 -0.07142857 -0.07142857 -0.07142857 -0.07142857
 -0.07142857 -0.07142857 -0.07142857 -0.07142857 -0.07142857 -0.07142857]
```

richpsharp:
This issue is fixed by explicitly testing if a pixel has a connection to the stream, if not the pixel is skipped and the distance to the stream is written as a nodata value.


tests/test_routing.py
```diff
@@ -858,7 +858,7 @@ def test_distance_to_channel_mfd(self):
             [2400, 17984, 17984, 17984, 17984, 17984, 17984, 17984, 17984,
              17984, 26880],
             [2400, 17984, 17984, 17984, 17984, 17984, 17984, 17984, 17984,
-             17984, 26880]], dtype=numpy.int32)
+             1178599424, 26880]], dtype=numpy.int32)
```
emlys, line 861: this is changing an edge pixel to flow off the array?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2839,7 +2838,7 @@ def distance_to_channel_mfd(
                         pixel.value += flow_dir_weight * (
                             weight_val + n_distance)
 
-                    if preempted:
+                    if preempted or pixel.value == distance_nodata:
```
emlys, line 2841: so the small negative values were happening because the nodata value `-1` would be go on to get divided: `pixel.value / sum_of_flow_weights`?

emlys, line 2841: what does `preempted` represent here? it's `1` when this pixel hasn't been visited before?

emlys (approved):
Thanks @richpsharp, just had a couple of questions to make sure I understand.

tests/test_routing.py
richpsharp: yes! in fact I was using your 1178599424 as the example you dug up in your earlier question.

src/pygeoprocessing/routing/routing.pyx
richpsharp: exactly, it was treating -1 as a valid value.

richpsharp: the algorithm above works by finding a pixel that has an undefined distance to stream value, then attempts to work that out by walking downstream. if one of the flow directions is undefined we save off what direction we were considering, push that pixel on the stack, then push the uncalculated downstream pixel on the stack. in that case the algorithm was "preempted" because it wasn't able to finish calculating the distance to stream for the pixel. is that okay?

richpsharp: the algorithm above works by finding a pixel that has an undefined distance to stream value, then attempts to work that out by walking downstream. if one of the flow directions is undefined we save off what direction we were considering, push that pixel on the stack, then push the uncalculated downstream pixel on the stack. in that case the algorithm was "preempted" because it wasn't able to finish calculating the distance to stream for the pixel. is that okay?

PR #206, opened 2021-07-02
Bugfix/stitch rasters bad offset 205

richpsharp:
Noticed by inspection that the `stitch_rasters` function not only was using `x_window_size` for all the general x/y checks but also check for right/bottom bounds as > rather than >= (non-inclusive). Fixed both and updated history.


beckyck (approved):
Looks good to me Rich!

PR #210, opened 2021-08-03
Handle nodata better in reclassify_raster

emlys:
I wasn't able to make a reproducible test case for this, so no test. Even Perrine's original data that caused the issue, worked for me today.

### DIFF
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -41,16 +41,13 @@ class ReclassificationMissingValuesError(Exception):
 
     Attributes:
         msg (str) - error message
-        missing_values (list) - a list of the missing values from the raster
-            that are not keys in the dictionary
 
     """
 
-    def __init__(self, msg, missing_values):
+    def __init__(self, msg):
         """See Attributes for args docstring."""
         self.msg = msg
-        self.missing_values = missing_values
-        super().__init__(msg, missing_values)
+        super().__init__(msg)
 
 
 LOGGER = logging.getLogger(__name__)
@@ -1851,10 +1848,10 @@ def reclassify_raster(
             if not all(has_map):
                 missing_values = unique[~has_map]
                 raise ReclassificationMissingValuesError(
-                    f'The following {missing_values.size} raster values'
-                    f' {missing_values} from "{base_raster_path_band[0]}"'
-                    ' do not have corresponding entries in the ``value_map``:'
-                    f' {value_map}.', missing_values)
+                    f'The following {missing_values.size} raster values '
+                    f'{missing_values} from "{base_raster_path_band[0]}" '
+                    'do not have corresponding entries in the ``value_map``: '
+                    f'{value_map}.')
         index = numpy.digitize(original_values[valid_mask], keys, right=True)
         out_array[valid_mask] = values[index]
         return out_array
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -41,13 +41,16 @@ class ReclassificationMissingValuesError(Exception):
 
     Attributes:
         msg (str) - error message
+        missing_values (list) - a list of the missing values from the raster
+            that are not keys in the dictionary
 
     """
 
-    def __init__(self, msg):
+    def __init__(self, msg, missing_values):
         """See Attributes for args docstring."""
         self.msg = msg
-        super().__init__(msg)
+        self.missing_values = missing_values
+        super().__init__(msg, missing_values)
 
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


src/pygeoprocessing/geoprocessing.py
```diff
@@ -41,16 +41,13 @@ class ReclassificationMissingValuesError(Exception):
...
-            that are not keys in the dictionary
 
     """
 
-    def __init__(self, msg, missing_values):
+    def __init__(self, msg):
         """See Attributes for args docstring."""
         self.msg = msg
-        self.missing_values = missing_values
-        super().__init__(msg, missing_values)
```
davemfish, line 53: It looks like the reason we have this `missing_values` attribute was for the benefit our invest `utils.reclassify_raster` wrapper:

```
    try:
        pygeoprocessing.reclassify_raster(
            raster_path_band, value_map, target_raster_path, target_datatype,
            target_nodata, values_required=True)
    except pygeoprocessing.ReclassificationMissingValuesError as err:
        error_message = (
                f"Values in the {raster_name} raster were found that are not"
                f" represented under the '{column_name}' column of the"
                f" {table_name} table. The missing values found in the"
                f" {raster_name} raster but not the table are:"
                f" {err.missing_values}.")
        raise ValueError(error_message)
```
Maybe we should just leave this class as-is for now?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1834,28 +1831,30 @@ def reclassify_raster(
...
-                    ' do not have corresponding entries in the ``value_map``:'
-                    f' {value_map}.', missing_values)
-        index = numpy.digitize(original_values.ravel(), keys, right=True)
-        return values[index].reshape(original_values.shape)
+                    f'The following {missing_values.size} raster values '
+                    f'{missing_values} from "{base_raster_path_band[0]}" '
+                    'do not have corresponding entries in the ``value_map``: '
+                    f'{value_map}.')
+        index = numpy.digitize(original_values[valid_mask], keys, right=True)
+        out_array[valid_mask] = values[index]
```
davemfish, line 1860: Nice to have have the update to the more modern numpy APIs here that don't require 1-d arrays. 👍

davemfish (changes requested):
Hi @emlys , this solution looks great to me! Thanks for thinking hard about the float edge case. It's definitely something we can put in the issue queue if we want to. It didn't sink in until after our call that the case you might be most worried about is if someone *thinks* they have integer-like values (e.g. 3.0) but they end up on the wrong side of a `numpy.digitize` bin. I was more fixated on the fact that no one would reasonably ask to reclassify a number like 3.00000000001, so we shouldn't design for it. 

Anyway I think it's good to get this clear fix in like you have it, and deal with the rest another time if needed.

I glanced at the tests we already have for `reclassify_raster` and I think they're adequate to cover these changes. I noticed we actually do test with a float raster, which seems appropriate. 

One minor change requested, see my note about the Exception class.

src/pygeoprocessing/geoprocessing.py
emlys: Oh okay. Maybe stormwater should be using `utils.reclassify_raster`? I changed it because when you call `pygeoprocessing.reclassify_raster` directly, the error message has that array of missing values tacked on the end with no explanation: `pygeoprocessing.geoprocessing.ReclassificationMissingValuesError: ('The following 1 raster values [-3.4e+38] from "/Users/emily/Documents/stormwater_workspace/intermediate/lulc_aligned.tif" do not have corresponding entries in the ``value_map``: {1: 1, 2: 1, 3: 0, 4: 1, 5: 1, 6: 0, 7: 1, 8: 0, 9: 0, 10: 1, 11: 0, 12: 0, 13: 0, 14: 0}.', array([-3.4e+38], dtype=float32))`

emlys: What do you think about this ac4b8d6? Moving the message up to the exception class, it can stay compatible with `utils.reclassify_raster` while also having a clear message by itself

emlys:
Thanks @davemfish!
> It didn't sink in until after our call that the case you might be most worried about is if someone thinks they have integer-like values (e.g. 3.0) but they end up on the wrong side of a numpy.digitize bin.

Yeah, that's my biggest concern. When `values_required=True`, hopefully the `numpy.isin` step would catch that problem. And it's good to know we're already testing with floats. I'll leave the issue open.

src/pygeoprocessing/geoprocessing.py
davemfish: > Maybe stormwater should be using utils.reclassify_raster?

I believe the main/only reason we have that `utils` function is to better format the error message for invest users that don't know what the `value_map` refers to. So yeah, I think the stormwater model should probably use the `utils` function for consistency across invest.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -46,11 +46,14 @@ class ReclassificationMissingValuesError(Exception):
...
+    def __init__(self, missing_values, raster_path, value_map):
         """See Attributes for args docstring."""
-        self.msg = msg
+        self.msg = (
+            f'The following {missing_values.size} raster values '
+            f'{missing_values} from "{raster_path}" do not have corresponding '
+            f'entries in the value map: {value_map}.')
         self.missing_values = missing_values
-        super().__init__(msg, missing_values)
+        super().__init__(self.msg)
```
davemfish, line 56: I think this is fine. If I understand it correctly, it's the change from `super().__init__(msg, missing_values)` to `super().__init__(self.msg)` that fixes the problem of the extra array printed in the message. Forming the message inside the exception seems fine, but not strictly necessary if we preferred not to pass along the `raster_path` and `value_map` args into the exception.

davemfish (approved):
Okay I think we're good, @emlys

src/pygeoprocessing/geoprocessing.py
davemfish: @emlys I guess I'll leave it up to you to merge after you read the comment above, just to be sure we're both understanding this the same way. If we are, I think we're good!

src/pygeoprocessing/geoprocessing.py
dcdenu4: Yea, the `utils` wrapper for this function was designed with InVEST users in mind, given how often we were seeing questions on the forums related to this error message. So the `utils` wrapper is meant for developers to have more control over what would be useful for the user to know if they have an error when reclassifying.

src/pygeoprocessing/geoprocessing.py
emlys: Yes, both arguments to `__init__` were getting printed. Since this exception is only used in this one function, I don't really care either if the message lives in the exception class or the function.

PR #213, opened 2021-08-06
Fix reclassify_raster bug when reclassifying to float

emlys:


src/pygeoprocessing/geoprocessing.py
```diff
@@ -1842,7 +1842,8 @@ def reclassify_raster(
 
     def _map_dataset_to_value_op(original_values):
         """Convert a block of original values to the lookup values."""
-        out_array = numpy.full(original_values.shape, target_nodata)
+        out_array = numpy.full(original_values.shape, target_nodata,
+                               dtype=numpy.float32)
```
phargogh: Since the datatype is user-defined, shouldn't this type be dependent on the input type? `_GDAL_TYPE_TO_NUMPY_LOOKUP` could be useful for this.

tests/test_geoprocessing.py
```diff
@@ -201,6 +201,32 @@ def test_reclassify_raster_empty_value_map(self):
                 (raster_path, 1), empty_value_map, target_path,
                 gdal.GDT_Float32, target_nodata, values_required=False)
 
+    def test_reclassify_raster_int_to_float(self):
+        """PGP.geoprocessing: test reclassify to float types."""
```
phargogh, line 205: This seems like a good test for basic functionality of `reclassify_raster`, regardless of the previous comment.

tests/test_geoprocessing.py
```diff
@@ -3167,7 +3193,8 @@ def test_stitch_rasters(self):
         stitch_by_etch_target_path = os.path.join(
             self.workspace_dir, 'stitch_by_etch.tif')
         pygeoprocessing.numpy_array_to_raster(
-            numpy.full((256, 256), -1, dtype=numpy.float32), -1, (1, -1), (0, 0),
+            numpy.full((256, 256), -1, dtype=numpy.float32), -
+            1, (1, -1), (0, 0),
```
phargogh: Could you bump that `-` down to the next line for readability?

tests/test_geoprocessing.py
```diff
@@ -4303,7 +4330,7 @@ def test_convolve_with_byte_kernel(self):
         pygeoprocessing.numpy_array_to_raster(
             array, 255, (20, -20), (0, 0), projection_wkt, signal_path)
         pygeoprocessing.numpy_array_to_raster(kernel.astype(numpy.uint8),
-            255, (20, -20), (0, 0), projection_wkt, int_kernel_path)
+                                              255, (20, -20), (0, 0), projection_wkt, int_kernel_path)
```
phargogh: While the proposed indentation is consistent with PEP8 indentation recommendations, I'm a little surprised that the long-line would be left as-is.  Is this < 80 characters?

phargogh (changes requested):
Thanks @emlys!  A couple of minor suggestions, the main one being that we use the user's raster type when creating the target array.

tests/test_geoprocessing.py
emlys: sure! not sure why my autopep8 broke the line there

tests/test_geoprocessing.py
emlys: no, it was longer. more strange stuff from autopep8

src/pygeoprocessing/geoprocessing.py
emlys: I think it would come out the same because `raster_calculator` sets the final datatype, but it makes more sense to use the `target_datatype` parameter here too. Is that what you mean?

phargogh: Yes, I did mean that it would make sense to use the `target_datatype` parameter here as well, which I see is what you've added in 385243a.

phargogh (approved):

PR #215, opened 2021-08-13
Fix distance_to_channel_mfd weighting

### LINKED ISSUE #214: distance_to_channel_mfd still factors in distance to edge
I suspect that the fix to #171 is not doing quite what we want. Pixels that don't drain to a stream at all are `nodata`, as expected. But pixels that have only part of their flow drain to a stream are still defined.

As I understand it `dist_to_channel_mfd` is weighted by the proportion of flow in each direction. Like it's the average distance to a stream, not the shortest. And it appears that the portions flowing off the edge or into nodata are getting a distance of 0.  I made a diagram to help keep track of what I was looking at 😄 
image

emlys:
Fixes #214

The problem was that each pixel's distance was averaged over all flow directions. This fixes it so that the average is taken only over the directions that reach a stream.


tests/test_routing.py
```diff
@@ -1032,6 +1032,56 @@ def test_distance_to_channel_mfd_with_weights(self):
...
+        # - g, h, and i flow off the bottom edge
+        # - flow paths are a->b->c->f->i, a->e->i, a->e->h, and d->g
+        # - d, g, and h are nodata because none of their flow reaches a stream
+        # - part of a and e's flow doesn't reach a stream. their distances
+        #   only measure the fraction that does reach the stream.
+        nodata = -1
+        expected_distance = numpy.array([
+            [(8+2*2**.5)/3, 3,      2],
+            [nodata,        2**.5,  1],
+            [nodata,        nodata, 0]])
```
phargogh, line 1071: Especially because MFD pixels are packed ints, this extra information about the MFD values is very helpful!

I also like the `expected_distance` being in a more human-readable form.

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2603,7 +2612,43 @@ def distance_to_channel_mfd(
         raster_driver_creation_tuple=DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS):
     """Calculate distance to channel with multiple flow direction.
 
-    Parameters:
+    MFD distance to channel for a pixel ``i`` is the average distance that
+    water travels from ``i`` until it reaches a stream, defined as::
```
phargogh, line 2616: Nice inclusion of an RST block literal.

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -2768,14 +2814,17 @@ def distance_to_channel_mfd(
...
 
+                # the algorithm starts from an arbitrary pixel
+                # it pushes its downstream neighbors to a stack, working depth-first
+                # until reaching either a stream or a raster edge
                 if visited_managed_raster.get(xi_root, yi_root) == 0:
                     visited_managed_raster.set(xi_root, yi_root, 1)
                     # arguments are x,y position of pixel, then last D8 flow
                     # direction processed (0-7), and last is the running
                     # accumulation distance accumulated by this pixel so far
                     # initialized to nodata
```
phargogh: These comments look to be out of date for this function now, given that we're using MFD and a different struct.

phargogh (changes requested):
Makes sense to me!  I had one trivial comment about an inline comment that I think should be updated for the new `MFDFlowPixelType` struct, but that's about it.

src/pygeoprocessing/routing/routing.pyx
emlys: removed this comment block

emlys:
Thanks @phargogh! I removed that outdated comment.

phargogh (approved):
Thanks @emlys !

PR #221, opened 2021-08-31
Allow nodata reclassification

dcdenu4:
Adding ability for nodata to be reclassified if it is in the `value_map` for `reclassify_raster`.

Added two new tests and updated history.

### DIFF
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -1837,8 +1837,6 @@ def reclassify_raster(
     """
     if len(value_map) == 0:
         raise ValueError("value_map must contain at least one value")
-    if target_nodata is None:
-        raise ValueError("target_nodata must be a valid number, received None")
     if not _is_raster_path_band_formatted(base_raster_path_band):
         raise ValueError(
             "Expected a (path, band_id) tuple, instead got '%s'" %
@@ -1856,6 +1854,14 @@ def reclassify_raster(
                 nodata_dest_value = val
                 del value_map_copy[key]
                 break
+
+    if target_nodata is None and nodata_dest_value is None:
+        raise ValueError(
+            "target_nodata was set to None and the base raster nodata"
+            " value was not represented in the value_map. reclassify_raster"
+            " does not assume the base raster nodata value should be used"
+            " as the target_nodata value. Set target_nodata to a valid number"
+            " and/or add a base raster nodata value mapping to value_map.")
 
     keys = sorted(numpy.array(list(value_map_copy.keys())))
     values = numpy.array([value_map_copy[x] for x in keys])
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -1850,6 +1850,7 @@ def reclassify_raster(
             if numpy.isclose(key, nodata):
                 nodata_dest_value = val
                 del value_map_copy[key]
+                break
 
     keys = sorted(numpy.array(list(value_map_copy.keys())))
     values = numpy.array([value_map_copy[x] for x in keys])
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -1798,8 +1798,10 @@ def reclassify_raster(
     """Reclassify pixel values in a raster.
 
     A function to reclassify values in raster to any output type. By default
-    the values except for nodata must be in ``value_map``. If nodata is
-    present in ``value_map`` nodata values are reclassified.
+    the values except for nodata must be in ``value_map``. If the base
+    raster nodata value is in ``value_map``, then nodata values are
+    reclassified to the mapped value. This would essentially leave no nodata
+    holes in the target raster.
 
     Args:
         base_raster_path_band (tuple): a tuple including file path to a raster
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -1834,6 +1834,8 @@ def reclassify_raster(
     """
     if len(value_map) == 0:
         raise ValueError("value_map must contain at least one value")
+    if target_nodata is None:
+        raise ValueError("target_nodata must be a valid number, received None")
     if not _is_raster_path_band_formatted(base_raster_path_band):
         raise ValueError(
             "Expected a (path, band_id) tuple, instead got '%s'" %
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -1798,7 +1798,8 @@ def reclassify_raster(
     """Reclassify pixel values in a raster.
 
     A function to reclassify values in raster to any output type. By default
-    the values except for nodata must be in ``value_map``.
+    the values except for nodata must be in ``value_map``. If nodata is
+    present in ``value_map`` nodata values are reclassified.
 
     Args:
         base_raster_path_band (tuple): a tuple including file path to a raster
@@ -1806,11 +1807,13 @@ def reclassify_raster(
         value_map (dictionary): a dictionary of values of
             {source_value: dest_value, ...} where source_value's type is the
             same as the values in ``base_raster_path`` at band ``band_index``.
-            Must contain at least one value.
+            Cannot be empty and must contain a mapping for all raster values
+            if ``values_required=True``. If nodata is mapped, nodata will be
+            reclassified, otherwise nodata will be set to ``target_nodata``.
         target_raster_path (string): target raster output path; overwritten if
             it exists
         target_datatype (gdal type): the numerical type for the target raster
-        target_nodata (numerical type): the nodata value for the target raster
+        target_nodata (numerical type): the nodata value for the target raster.
             Must be the same type as target_datatype
         values_required (bool): If True, raise a ValueError if there is a
             value in the raster that is not found in ``value_map``.
@@ -1837,8 +1840,19 @@ def reclassify_raster(
             base_raster_path_band)
     raster_info = get_raster_info(base_raster_path_band[0])
     nodata = raster_info['nodata'][base_raster_path_band[1]-1]
-    keys = sorted(numpy.array(list(value_map.keys())))
-    values = numpy.array([value_map[x] for x in keys])
+    # If nodata was included in the value_map pop it from our lists
+    # and handle it separately. Doing this on the off chance nodata is
+    # a max or min float which can cause floating decimal discrepencies.
+    value_map_copy = value_map.copy()
+    nodata_dest_value = target_nodata
+    if nodata is not None:
+        for key, val in value_map.items():
+            if numpy.isclose(key, nodata):
+                nodata_dest_value = val
+                del value_map_copy[key]
+
+    keys = sorted(numpy.array(list(value_map_copy.keys())))
+    values = numpy.array([value_map_copy[x] for x in keys])
 
     def _map_dataset_to_value_op(original_values):
         """Convert a block of original values to the lookup values."""
@@ -1849,6 +1863,7 @@ def reclassify_raster(
             valid_mask = numpy.full(original_values.shape, True)
         else:
             valid_mask = ~numpy.isclose(original_values, nodata)
+            out_array[~valid_mask] = nodata_dest_value
 
         if values_required:
             unique = numpy.unique(original_values[valid_mask])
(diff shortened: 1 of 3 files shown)


src/pygeoprocessing/geoprocessing.py
```diff
@@ -1837,8 +1840,20 @@ def reclassify_raster(
             base_raster_path_band)
     raster_info = get_raster_info(base_raster_path_band[0])
     nodata = raster_info['nodata'][base_raster_path_band[1]-1]
-    keys = sorted(numpy.array(list(value_map.keys())))
-    values = numpy.array([value_map[x] for x in keys])
+    # If nodata was included in the value_map pop it from our lists
+    # and handle it separately. Doing this on the off chance nodata is
+    # a max or min float which can cause floating decimal discrepencies.
+    value_map_copy = value_map.copy()
+    nodata_dest_value = target_nodata
```
emlys, line 1850: What should happen if `target_nodata` is `None`? If `target_nodata` is `None` but there are nodata pixels in the input, we might end up introducing `nan`s into the output on line 1867. I think that case is invalid but should we check for it?

emlys (commented):
Thanks for tackling this @dcdenu4! I'm still not sure I fully understand what the function _should_ be doing with the combination of `value_map` and `target_nodata`. I made a [gist](https://gist.github.com/emlys/e398a868ad0662bc1fe3e00d947a1cde) with some test cases of what I think should happen, but not sure about it. With these changes, they all pass except case #2 (change nodata value without reclassifying).

There's probably a few ways this could work but whichever way you go with, I'd add to the docstring describing the behavior in each of these cases. The difference between `target_nodata` and including the nodata value in the `value_map` isn't intuitive to me.

tests/test_geoprocessing.py
```diff
@@ -227,6 +227,56 @@ def test_reclassify_raster_int_to_float(self):
...
+        value_map = {
+            test_value: 0,
+            nodata: 1,
+        }
+        target_nodata = -1
+        pygeoprocessing.reclassify_raster(
+            (raster_path, 1), value_map, target_path, gdal.GDT_Float32,
+            target_nodata, values_required=True)
+        target_array = pygeoprocessing.raster_to_numpy_array(target_path)
+        self.assertAlmostEqual(numpy.sum(target_array), 2)
```
emlys, line 254: I would also check that the nodata value is set as expected

src/pygeoprocessing/geoprocessing.py
dcdenu4: This is a great question, I just did a quick check in `geoprocessing.py` and I don't see any precedent for allowing `target_nodata` to be`None`. `convolve_2d` has `target_nodata=None` as the default but then sets it to minimum value of a float32 if it isn't changed. I'll have this function do something similar.

dcdenu4: Although, the docstring does say "numerical type". I'm actually more inclined to report an error if `target_nodata` is `None`.

dcdenu4:
Hey @emlys, thanks for the review and making that gist! I agree that handling `nodata` is a little confusing.

>With these changes, they all pass except case #2 (change nodata value without reclassifying).

I think the expected result in the 2nd gist example should replace the `base_nodata` value with `10`? https://gist.github.com/emlys/e398a868ad0662bc1fe3e00d947a1cde#file-demo_reclassify_cases-py-L60. That is to say, `nodata` was not in the `value_map` so update the base `nodata` to the `target_nodata`. That's how I think of it anyway. 

My general handle on it is:
If the current nodata value is in the `value_map` then the user wants to reclassify current nodata to something else. Usually this would be something other than a nodata value, but could still be whatever `target_nodata` is, though it'd be unnecessary to do so. If current nodata is not in the `value_map` then the user wants to persist nodata values to the target raster but might want to update the target rasters nodata value to `target_nodata`.

src/pygeoprocessing/geoprocessing.py
phargogh: This is a really interesting question. As far as precedent, doesn't raster_calculator allow for the nodata value to be None? I think the source code for raster_calculator explicitly handles that case.

phargogh:
I agree that the default behavior should be to reclassify `base_nodata` to `target_nodata`.  If the user wants to define a new reclassification rule for `base_nodata` to be mapped to something else (or `target_nodata`), then they have that override available.

For the case of when:
1. `base_nodata` is set
2. `target_nodata` is NOT set (`target_nodata is None`)
3. `base_nodata` is NOT in the `value_map

Then I think we should not guess or pick a new nodata value ... this feels like programmer error to me and I think the function should raise an exception.

I think that would cover all of the nodata/value_map cases ... what do you think?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1831,14 +1836,28 @@ def reclassify_raster(
     """
     if len(value_map) == 0:
         raise ValueError("value_map must contain at least one value")
+    if target_nodata is None:
+        raise ValueError("target_nodata must be a valid number, received None")
```
phargogh: I _think_ there could be an argument for having `target_nodata=None` (though I don't know if we use this in InVEST).  It seems to me that the main case we're trying to prevent is a broken reclassification rule ... if `base_nodata` is defined but `target_nodata` is not, then how do we translate that?  In that case, I absolutely think we should raise an exception.

phargogh (commented):
I just wanted to add a thought on the base/target nodata handling and when we might want to raise an exception.

src/pygeoprocessing/geoprocessing.py
dcdenu4: > I think the source code for raster_calculator explicitly handles that case.

It does guard against an undefined target nodata value, you're right. I think this is a case where intuitively it doesn't make sense to me why you would pass in a None value for target nodata, however, if there is precedent of guarding against target nodata values of None, then we can certainly do it here as well.

`convolve_2d` at the very least handles None nodata by raising an error in a specific case, and setting to min float32 in another.

src/pygeoprocessing/geoprocessing.py
dcdenu4: Yea, I think you're right in that for this case the function does not have enough information on what to do, so should raise an exception. I was coming from the mindset that `target_nodata` should not be allowed to be None at all, which would avoid this situation. I'd love to hear the argument for `target_nodata=None` as being something a user would explicitly want. I guess I'd prefer to avoid over engineering if not allowing `target_nodata` to be None doesn't restrict the use of the function in anyway...

dcdenu4:
> Then I think we should not guess or pick a new nodata value ... this feels like programmer error to me and I think the function should raise an exception.

Totally agree.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1798,19 +1798,24 @@ def reclassify_raster(
...
             same as the values in ``base_raster_path`` at band ``band_index``.
-            Must contain at least one value.
+            Cannot be empty and must contain a mapping for all raster values
+            if ``values_required=True``. If nodata is mapped, nodata will be
+            reclassified, otherwise nodata will be set to ``target_nodata``.
         target_raster_path (string): target raster output path; overwritten if
             it exists
         target_datatype (gdal type): the numerical type for the target raster
-        target_nodata (numerical type): the nodata value for the target raster
+        target_nodata (numerical type): the nodata value for the target raster.
```
emlys, line 1817: I can't figure out how to make multi-line commit suggestions, but what about:
```suggestion
        target_nodata (numerical type): the nodata value for the target raster.
            Must be the same type as target_datatype. All nodata pixels in the 
            base raster will be reclassified to this value, and the target 
            raster's nodata value will be set to this value.
```

src/pygeoprocessing/geoprocessing.py
phargogh: Looking through the InVEST source code, the only concrete example of this that I found is that GLOBIO allows a user-defined GLOBIO LULC, and we use the nodata value, whatever it is (even if it's `None`).

I also looked through the first 10 pages of a github search for `reclassify_raster` and the GLOBIO case is the only one I can find where the nodata value _could_ be set.  So I'll concede that this is probably an edge case.

But what makes me wary of requiring a nodata value is that it's [an explicitly optional part of the GDAL raster data model](https://gdal.org/user/raster_data_model.html#raster-band).  That's not to say that we have to support everything that GDAL supports ... GDAL rasters support mask bands, for instance, which we've never used to my knowledge.

I still think that the best example is not so much a matter of explicitly wanting a nodata value as much as it is that a nodata value might not be needed.  If a user doesn't need a nodata value, why should one be required?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1798,19 +1798,24 @@ def reclassify_raster(
     """Reclassify pixel values in a raster.
 
     A function to reclassify values in raster to any output type. By default
-    the values except for nodata must be in ``value_map``.
+    the values except for nodata must be in ``value_map``. If the base
+    raster nodata value is in ``value_map``, then nodata values are
+    reclassified to the mapped value. This would essentially leave no nodata
+    holes in the target raster.
```
emlys: How about adding something like, "Include the base raster nodata value in ``value_map`` to reclassify nodata pixels to a valid value (not setting this as the target raster's nodata value). Set ``target_nodata`` to reclassify nodata pixels to a different nodata value (setting this as the target raster's nodata value)."

emlys:
@dcdenu4, this looks good! Just had a couple suggestions to expand on the docstring.

src/pygeoprocessing/geoprocessing.py
phargogh: Looks like your multi-line suggestion worked, Emily!

emlys: I'm not sure if it will overwrite line 1819 also, or just add in between them

src/pygeoprocessing/geoprocessing.py
dcdenu4: I gave it a go here: https://github.com/natcap/pygeoprocessing/pull/221/commits/31b9c7164353426e15d7c43d801d3e5823038e57

>Set target_nodata to reclassify nodata pixels to a different nodata value (setting this as the target raster's nodata value)

I left off this last bit because `target_nodata` is a unique argument and `setting` it didn't quite feel like the right phrasing to me. I also think this part should be plenty explained in the Args section. Let me know what you think.

src/pygeoprocessing/geoprocessing.py
dcdenu4: I didn't commit your exact suggestion but put pieces of it in here: https://github.com/natcap/pygeoprocessing/commit/31b9c7164353426e15d7c43d801d3e5823038e57

I basically just elaborated a little onto yours and left the last part: `and the target raster's nodata value will be set to this value.` out since it's already mentioned in the first sentence.

Let me know what you think!

src/pygeoprocessing/geoprocessing.py
dcdenu4: I believe I addressed this here: https://github.com/natcap/pygeoprocessing/pull/221/commits/705d9d066677416c22d24228f85d69901e07b76b

Allowing target nodata values of None as long as they don't break the case you mention above (target_nodata = None and base raster nodata not present in value_map)

dcdenu4:
@phargogh 

>For the case of when:
>
>    base_nodata is set
>    target_nodata is NOT set (target_nodata is None)
>    base_nodata is NOT in the `value_map
>
>Then I think we should not guess or pick a new nodata value ... this feels like programmer error to me and I think the function should raise an exception.

I think I addressed this case here: https://github.com/natcap/pygeoprocessing/pull/221/commits/705d9d066677416c22d24228f85d69901e07b76b

dcdenu4:
Hey @emlys and @phargogh thanks for the feedback. I tried to address your comments with some updated docstrings, comments, error handling, and additional test. Let me know what you think!

src/pygeoprocessing/geoprocessing.py
emlys: That looks good to me!

src/pygeoprocessing/geoprocessing.py
emlys: I like your version! These things are often tricky to describe neatly.

emlys (approved):
Approving but @phargogh I'll let you merge once you get a chance to review also!

phargogh (approved):
Thanks so much for the patch and for clarifying this ambiguity!

Issue #42, opened 2020-03-06
Improve support for signed byte rasters in reclassification


phargogh:
Some time ago in a forums post (https://community.naturalcapitalproject.org/t/habitat-q-error/431/15), I helped a user with a reclassification issue in the InVEST Habitat Quality model that ended up being a bug in how GDAL was handling signed byte rasters and how `reclassify_raster` was handling nodata values.

In order to do this well (and in part because I looked at this issue some time ago), we need to better understand how `reclassify_raster` handles signed byte rasters and the nodata values themselves.

richpsharp:
hi @phargogh, I'm going through the pgp issue queue and tried to see if I could identify this problem. Just from the forum posts I'm not sure exactly the values that were at play, but in an attempt to recreate the issue I made a test that reclassifies a signed byte raster with a negative nodata value and negative valid values and wasn't able to reproduce any issues. Maybe we fixed this somehow? Could you take a look at my test in this branch: https://github.com/richpsharp/pygeoprocessing/tree/bugfix/GEOPROCESSING-42-reclassify-signed/ here: `tests/test_geoprocessing.py::PyGeoprocessing10::test_reclassify_signed_byte_raster` and see if I was doing it right? If you think the issue is still there can you post me the user's data or more information? Or if it looks okay can you mark this issue closed and I'll delete my branch?

phargogh:
Hey @richpsharp I just dug back through my email and found that the user had a malformed LULC raster, where a pixel value of `128` was being used (and was probably supposed to represent nodata), `128` was NOT in the reclassification table (this was InVEST HQ, so the sensitivity table), and the LULC also did not have a defined nodata value.

The raster did happen to be a signed byte raster, so I probably assumed that there was an issue when there wasn't.

phargogh:
While reading through the GDAL docs for a completely different reason, I came across [this tidbit about signed rasters in the GDAL data model description](https://gdal.org/user/raster_data_model.html?highlight=signedbyte#image-structure-domain):

> PIXELTYPE: This may appear on a GDT_Byte band (or the corresponding dataset) and have the value SIGNEDBYTE to indicate the unsigned byte values between 128 and 255 should be interpreted as being values between -128 and -1 for applications that recognise the SIGNEDBYTE type.

[...]

PR #222, opened 2021-09-03
Adding a test for signed_byte raster_calculator operations.

phargogh:
From what I've read in the source code, signed byte support in GDAL appears to be very good, provided that the appropriate metadata is set (and it's this metadata that GDAL uses to determine whether a raster is a signed byte type).

This test just adds a functional test for raster calculator to make sure that we can interpret signed byte values correctly in a local op.

RE #42

### DIFF
--- a/tests/test_geoprocessing.py
+++ b/tests/test_geoprocessing.py
@@ -2075,6 +2075,37 @@ class TestGeoprocessing(unittest.TestCase):
         expected_result[127, 127] = nodata
         numpy.testing.assert_allclose(result, expected_result)
 
+    def test_raster_calculator_signed_byte(self):
+        """PGP.geoprocessing: test that signed byte pixels interpreted."""
+        pixel_array = numpy.ones((128, 128), numpy.byte)
+        # ArcGIS will create a signed byte raster with an unsigned value of 255
+        # that actually is supposed to represent -1, even though the nodata
+        # value will be set as -1.
+        pixel_array[0, 0] = 255  # 255 ubyte is -1 byte
+        nodata_base = -1
+        base_path = os.path.join(self.workspace_dir, 'base.tif')
+        _array_to_raster(
+            pixel_array, nodata_base, base_path,
+            creation_options=['PIXELTYPE=SIGNEDBYTE'])
+
+        # The local_op should receive at least one value less than 0
+        def local_op(byte_values):
+            byte_values[byte_values < 0] = 2
+            return byte_values
+
+        target_path = os.path.join(self.workspace_dir, 'target.tif')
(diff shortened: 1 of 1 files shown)


tests/test_geoprocessing.py
```diff
@@ -2075,6 +2075,37 @@ def local_op(scalar, raster_array, col_array):
...
+        pixel_array = numpy.ones((128, 128), numpy.byte)
+        # ArcGIS will create a signed byte raster with an unsigned value of 255
+        # that actually is supposed to represent -1, even though the nodata
+        # value will be set as -1.
+        pixel_array[0, 0] = 255  # 255 ubyte is -1 byte
+        nodata_base = -1
+        base_path = os.path.join(self.workspace_dir, 'base.tif')
+        _array_to_raster(
+            pixel_array, nodata_base, base_path,
+            creation_options=['PIXELTYPE=SIGNEDBYTE'])
```
dcdenu4, line 2089: nice

tests/test_geoprocessing.py
```diff
@@ -2075,6 +2075,37 @@ def local_op(scalar, raster_array, col_array):
...
+
+        target_path = os.path.join(self.workspace_dir, 'target.tif')
+        pygeoprocessing.raster_calculator(
+            [(base_path, 1)], local_op, target_path, gdal.GDT_Byte, None)
+
+        target_array = pygeoprocessing.raster_to_numpy_array(target_path)
+        # We expect that any values less than 0 are converted to 2.
+        # This ensures that
+        # we expect a negative result even though we put in a positive because
+        # we know signed bytes will convert.
```
dcdenu4:
```suggestion
        # This ensures that we expect a negative result even though we
        # put in a positive because we know signed bytes will convert.
```

tests/test_geoprocessing.py
```diff
@@ -2075,6 +2075,37 @@ def local_op(scalar, raster_array, col_array):
...
+        self.assertEqual(target_array[0, 0], 2)
+        self.assertEqual(target_array.sum(), 128 * 128 + 1)
+        self.assertEqual(list(numpy.unique(target_array)), [1, 2])
```
dcdenu4, line 2106: Would it be worth testing here that `nodata` was set correctly? I can't remember or quickly find if `GDT_Byte` is signed or unsigned.

dcdenu4 (commented):
Hey @phargogh , just had a quick question / comment.

tests/test_geoprocessing.py
phargogh: It should for sure be tested!  Added in 6d52804

dcdenu4 (approved):
Looks good.

PR #231, opened 2021-12-02
Test pygeoprocessing against python 3.10 (and replace setup.py commands)

### LINKED ISSUE #225: Test against Python 3.10
Python 3.10 us supposed to be released today (Oct 4, 2021) so let's make sure everything is working as expected.

### LINKED ISSUE #230: setup.py is being deprecated
While working on #225 , I found the following warnings in the build logs for `python setup.py install`:

```
 /usr/local/miniconda/envs/pyenv/lib/python3.10/site-packages/setuptools/installer.py:27: SetuptoolsDeprecationWarning: setuptools.installer is deprecated. Requirements should be satisfied by a PEP 517 installer.
  warnings.warn(
/usr/local/miniconda/envs/pyenv/lib/python3.10/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
```

So we should look into how best to replace `setuptools.installer` with something more modern.

phargogh:
This PR fixes #225 (for testing against python 3.10) and also #230 (for deprecating setup.py commands).  The latter change is because `python setup.py install` fails in python 3.10 so it's time to use a different set of tools.

Most of the changes to the `.yml` files should just be whitespace, aside from a little matrix modification to test pygeoprocessing on 3.10 using the latest version of GDAL (since none of the earlier python versions have builds for GDAL on conda-forge).

### DIFF
--- a/.github/workflows/build-py-dists.yml
+++ b/.github/workflows/build-py-dists.yml
@@ -8,7 +8,7 @@ jobs:
             fail-fast: false
             matrix:
                 os: [windows-latest, macos-latest]
-                python-version: [3.7, 3.8, 3.9]
+                python-version: [3.7, 3.8, 3.9, "3.10"]
                 python-arch: [x86, x64]
                 include:
                     - python-version: 3.7
@@ -17,6 +17,8 @@ jobs:
                       numpy: "numpy~=1.17"
                     - python-version: 3.9
                       numpy: "numpy~=1.19"
+                    - python-version: "3.10"
+                      numpy: "numpy~=1.21"
 
                 exclude:
                     - os: macos-latest
--- a/.github/workflows/build-py-dists.yml
+++ b/.github/workflows/build-py-dists.yml
@@ -24,9 +24,6 @@ jobs:
                     - os: macos-latest
                       python-arch: x86
 
-                    - python-version: "3.10"
-                      python-arch: x86
-
         steps:
             - uses: actions/checkout@v2
               with:
--- a/.github/workflows/build-py-dists.yml
+++ b/.github/workflows/build-py-dists.yml
@@ -24,6 +24,9 @@ jobs:
                     - os: macos-latest
                       python-arch: x86
 
+                    - python-version: "3.10"
+                      python-arch: x86
+
         steps:
             - uses: actions/checkout@v2
               with:
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -38,7 +38,7 @@ jobs:
         shell: bash -l {0}
         run: |
           conda install mamba -n base -c conda-forge
-          mamba upgrade -y pip setuptools
+          conda upgrade -y pip setuptools
 
       - name: Install PyGeoprocessing
         shell: bash -l {0}
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -53,7 +53,7 @@ jobs:
         shell: bash -l {0}
         run: |
             mamba install --file requirements.txt
-            mamba install gdal==${{ matrix.gdal }} pytest
+            mamba install -c conda-forge gdal==${{ matrix.gdal }} pytest
             python setup.py install
 
       - name: Test with pytest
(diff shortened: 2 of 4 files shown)


.github/workflows/pythonpackage.yml
```diff
@@ -5,48 +5,65 @@ on: [push, pull_request]
...
+        include:
+          - python-version: "3.10"
+            gdal: "3.4.0"
+            os: ubuntu-latest
+
+          - python-version: "3.10"
+            gdal: "3.4.0"
+            os: windows-latest
+
+          - python-version: "3.10"
```
emlys, line 24: Are we `include`ing these combinations instead of adding 3.10 to the matrix `python-version` array because we don't expect python 3.10 to work with the other GDAL versions? Or just doesn't seem necessary to test all of them?

.github/workflows/build-py-dists.yml
```diff
@@ -8,7 +8,7 @@ jobs:
             fail-fast: false
             matrix:
                 os: [windows-latest, macos-latest]
-                python-version: [3.7, 3.8, 3.9]
+                python-version: [3.7, 3.8, 3.9, "3.10"]
```
emlys, line 11: Curious why the quotes are needed? And do the other versions need them as well?

.github/workflows/build-py-dists.yml
```diff
@@ -17,11 +17,16 @@ jobs:
...
                       numpy: "numpy~=1.19"
+                    - python-version: "3.10"
+                      numpy: "numpy~=1.21"
 
                 exclude:
                     - os: macos-latest
                       python-arch: x86
 
+                    - python-version: "3.10"
+                      python-arch: x86
```
emlys: Does python 3.10 not support `x86`? I skimmed the changelog and didn't find anything relevant. Maybe worth a comment here explaining why it's excluded.

emlys: Since invest no longer supports `x86`, could we drop it here as well? Would be nice to halve the number of matrix combinations that have to run.

emlys (commented):
Looks good @phargogh, just had a few questions!

.github/workflows/build-py-dists.yml
phargogh: The first time I tried to add `3.10` to the build matrix, I was _shocked_ to see that GHA had added not Python 3.10, but Python 3.1 and everything failed.  It turns out that the YML parser tries to do some intelligent casting, so the quotes are needed here to specify that we do indeed want the string `"3.10"` and not the float `3.1`.

.github/workflows/build-py-dists.yml
phargogh: Ha, I think I put this in here without thinking ... maybe it's because `conda-forge` dropped their `x86` builds?  In any case, this shouldn't apply here because GHA has their own repository of python versions, so I've removed this exclusion in 8a86422

phargogh: >Since invest no longer supports x86, could we drop it here as well? Would be nice to halve the number of matrix combinations that have to run.

You know, that's an interesting point, and it makes me wonder about the PyPI download stats of the x86 and x64 windows wheels.  I'm not at all against dropping them, but I think for the moment I'd like to keep the x86 builds until we've had a chance to take a closer look, so I've created https://github.com/natcap/pygeoprocessing/issues/232 for this so we can think it through.  How does that sound?

.github/workflows/pythonpackage.yml
phargogh: Yeah, the issue is that the only GDAL builds for python 3.10 are for GDAL 3.4.0+.  So if I add `"3.10"` to `python-version`, then now I have to exclude 2 versions of GDAL across 3 OSes, which is more than I'm including.

I could very well be using the "exclude" section of the build matrix incorrectly, but I found it kind of challenging and frustrating to exclude a range of builds from the matrix.

So it seemed most direct to just state the build matrix and then include a few bonus runs for python 3.10.

.github/workflows/build-py-dists.yml
emlys: Sounds good!

emlys (approved):
Sorry I forgot about this one @phargogh!

Issue #235, opened 2022-01-11
Proposal: `raster_calculator` wrapper


emlys:
Proposal for a higher-level version of `raster_calculator` that abstracts away some details of working with rasters and arrays. The goal would be to be able to do arithmetic with rasters and scalars interchangeably, like how `numpy` lets you use arrays and scalars interchangeably.

This would abstract away:
* masking out results where any input is nodata
* details of calling `raster_calculator`: indicating the band of each raster, and marking scalar values as `'raw'`
* (optionally) choosing an output dtype
* (optionally) choosing an output nodata value

## Justification
For #234, I skimmed through the models and found that pixelwise multiplication and addition are reimplemented in several models.

I also counted over 3 dozen functions that use `raster_calculator` to do multi-part arithmetic operations on rasters (e.g., adding two rasters pixelwise, then multiplying by a scalar). These functions are highly similar and the unique logic usually comes down to one line of array math. The rest is taken up by reading and comparing nodata values, masking, and calling `raster_calculator`.

So a solution for #234 that _also_ handles other, or combinations of, pixelwise arithmetic, could save a lot more repeated code (vs. making individual functions for the most-used operations).

## Signature and docstring for proposed function
```
def raster_calculator_2(op, *inputs, target_path='', target_nodata=None, 
                      target_dtype=None):
    """
    Apply a pixelwise function to a series of raster and/or scalar inputs.

    The output raster will have nodata where any input raster has nodata.
    Raster inputs are split into aligned blocks, and the function is 
    applied individually to each stack of blocks (as numpy arrays). 

    Args:
        op (function): Function to apply to the inputs. It should accept a 
            number of arguments equal to the length of ``*inputs``. It should 
           return a numpy array with the same shape as its array input(s).
        *inputs: Strings (raster paths) and/or scalar inputs to ``op``, in the 
            order that they will be passed to ``op``. All raster inputs 
            should be aligned and have the same dimensions. At least one input 
            must be a raster. 
        target_path (str): path to write out the output raster.
        target_nodata (number): Nodata value to use for the output raster. 
            Optional. If not provided, a suitable nodata value will be chosen.
        target_dtype (numpy.dtype): dtype to use for the output. Optional. If 
            not provided, a suitable dtype will be chosen.

    Returns:
        ``None``
```
_Note: A suitable dtype could be chosen by taking the most precise dtype of all inputs, so that no information is lost.
A suitable nodata value could be derived from the dtype. See also natcap/invest#783._

[...]

PR #236, opened 2022-01-11
`raster_calculator` wrapper

emlys:
Draft solution for #235. I'm requesting everyone's review because this (or a different solution) could affect a lot of code in invest. See the similar, expression-based function [`evaluate_raster_calculator_expression`](https://github.com/natcap/pygeoprocessing/blob/e994119e4c04f1b68867da87d416085935bdbdf4/src/pygeoprocessing/symbolic.py#L14) for comparison.

This adds a new function `raster_op` (suggestions for the name are welcome!) and unit tests for it. Here are some key points:

* Takes in a callable rather than an expression for maximum flexibility
* User may provide a target dtype, or let an appropriate one be chosen automatically
* User may provide a target nodata value, or let an appropriate one be chosen automatically
* If the user provides a target nodata value and it can't be safely cast to the target dtype (whether automatically chosen or not), a `ValueError` is raised. This is an easy mistake that would probably never be intentional.
* In contrast, all target dtypes are allowed even if casting the data to that dtype loses information. E.g. a float32 input and a target dtype of int8 is valid. I'm not sure if we have any use case for intentionally casting to a smaller type - I'd be equally happy to make this raise `ValueError`.
* Unlike `evaluate_raster_calculator_expression`, I did not include `default_nan` and `default_inf` options to fill in any `nan` and `inf` values. I can't think of any use cases for those in invest - I think it's fine to let `nan`s and `inf`s pass through. If we do want to replace `nan` or `inf`, that can easily be done in the `op`.

### DIFF
--- a/src/pygeoprocessing/geoprocessing_core.pyx
+++ b/src/pygeoprocessing/geoprocessing_core.pyx
@@ -28,9 +28,13 @@ from osgeo import osr
 import numpy
 import pygeoprocessing
 
-DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS = ('GTIFF', (
-    'TILED=YES', 'BIGTIFF=YES', 'COMPRESS=LZW',
-    'BLOCKXSIZE=256', 'BLOCKYSIZE=256'))
+
+DEFAULT_CREATION_OPTIONS = ('TILED=YES', 'BIGTIFF=YES', 'COMPRESS=LZW',
+                             'BLOCKXSIZE=256', 'BLOCKYSIZE=256')
+INT8_CREATION_OPTIONS = DEFAULT_CREATION_OPTIONS + (
+    'PIXELTYPE=SIGNEDBYTE',)
+DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS = ('GTIFF', DEFAULT_CREATION_OPTIONS)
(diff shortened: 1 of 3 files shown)


src/pygeoprocessing/geoprocessing_core.pyx
```diff
@@ -28,9 +28,13 @@ from osgeo import osr
...
-DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS = ('GTIFF', (
-    'TILED=YES', 'BIGTIFF=YES', 'COMPRESS=LZW',
-    'BLOCKXSIZE=256', 'BLOCKYSIZE=256'))
+
+DEFAULT_CREATION_OPTIONS = ('TILED=YES', 'BIGTIFF=YES', 'COMPRESS=LZW',
+                             'BLOCKXSIZE=256', 'BLOCKYSIZE=256')
+INT8_CREATION_OPTIONS = DEFAULT_CREATION_OPTIONS + (
+    'PIXELTYPE=SIGNEDBYTE',)
+DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS = ('GTIFF', DEFAULT_CREATION_OPTIONS)
+INT8_GTIFF_CREATION_TUPLE_OPTIONS = ('GTIFF', INT8_CREATION_OPTIONS)
```
emlys, line 37: It was convenient to make a separate variable for the int8 options.

PR #237, opened 2022-01-15
pin numpy versions in pyproject.toml

emlys:
Pin `numpy` to the minimum compatible versions in pyproject.toml `build-system` `requires`. Necessary for natcap/invest.users-guide#62. Pygeoprocessing is getting built from source on ReadTheDocs using whatever the latest numpy version is (1.22.0). It later gets downgraded to numpy 1.21.5 to be compatible with invest. The pygeoprocessing built with the newer numpy version conflicts.

Also removed some redundant lines in the workflow yml.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,4 +5,12 @@
 # dynamically import GDAL via python's import system.  This behavior means
 # that we can provide a much easier build experience so long as GDAL is
 # available at runtime.
-requires = ['setuptools', 'wheel', 'setuptools_scm', 'numpy', 'cython']
+requires = [
+    'setuptools', 'wheel', 'setuptools_scm', 'cython',
+    # use minimum compatible numpy for each python version
+    # https://github.com/cython/cython/issues/4452
+    'numpy==1.16.5; python_version=="3.7"',
+    'numpy==1.17.3; python_version=="3.8"',
+    'numpy==1.19.3; python_version=="3.9"',
+    'numpy==1.21.2; python_version=="3.10"']
+build-backend = "setuptools.build_meta"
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -5,4 +5,12 @@
...
-requires = ['setuptools', 'wheel', 'setuptools_scm', 'numpy', 'cython']
+requires = [
+    'setuptools', 'wheel', 'setuptools_scm', 'cython',
+    # use minimum compatible numpy for each python version
+    # https://github.com/cython/cython/issues/4452
+    'numpy==1.16.5; python_version=="3.7"',
+    'numpy==1.17.3; python_version=="3.8"',
+    'numpy==1.19.3; python_version=="3.9"',
+    'numpy==1.21.2; python_version=="3.10"']
+build-backend = "setuptools.build_meta"
```
emlys, line 16: Not quite sure how this is different from the [fallback option](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#fallback-behaviour), but it's included in [an example from the setuptools docs](https://setuptools.pypa.io/en/latest/build_meta.html#how-to-use-it).

phargogh (changes requested):
`pyproject.toml` looks good to me!  Thanks for pinning these.

Since we have similar conditional numpy version logic in `.github/workflows/build-py-dists.yml`, could you update that workflow file to build wheels with `python -m build --wheel` so that we can remove the duplication of numpy version pinning?

Since this is also kind of a user-facing change (it affects the source and binary distributions), could you also add a note to `HISTORY`?

emlys:
> could you update that workflow file to build wheels with python -m build --wheel so that we can remove the duplication of numpy version pinning?

Good idea! Did the same thing for building the sdist as well.

> could you also add a note to HISTORY?

Done!

phargogh (approved):
Great, thank you!  I'm adding my approval here, assuming the tests pass.

Issue #243, opened 2022-03-02
LOGGER.warn deprecationWarnings in pygeoprocessing


phargogh:
We have a user report of some logging warnings in pygeoprocessing being elevated to Exceptions in SWY, so this will need to be fixed here and then InVEST re-built.

https://community.naturalcapitalproject.org/t/swy-error-warn-method-invest-3-10-2/2534

phargogh:
Resolved in #245

PR #245, opened 2022-03-03
Elevate warnings to exceptions when running tests

phargogh:
This PR elevates warnings encountered in tests to exceptions.  I also went ahead and fixed all the warnings I encountered, and added some functionality to test the one mentioned in #243 as well as another related warning that will help make the `calculate_disjoint_polygon_set` function more resilient.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,3 +19,8 @@ build-backend = "setuptools.build_meta"
 [tool.pytest.ini_options]
 # Raise warnings to exceptions.
 filterwarnings = 'error'
+
+
+[tool.setuptools_scm]
+version_scheme = "post-release"
+local_scheme = "node-and-date"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,3 +14,8 @@ requires = [
     'numpy==1.19.3; python_version=="3.9"',
     'numpy==1.21.2; python_version=="3.10"']
 build-backend = "setuptools.build_meta"
+
+
+[tool.pytest.ini_options]
+# Raise warnings to exceptions.
+filterwarnings = 'error'
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -2278,7 +2278,6 @@ def calculate_disjoint_polygon_set(
                 f'empty geometry in {vector_path} FID: {poly_feat.GetFID()}, '
                 'skipping...')
             continue
-        LOGGER.info(poly_geom_ref.ExportToWkt())
         # with GDAL>=3.3.0 ExportToWkb returns a bytearray instead of bytes
         shapely_polygon_lookup[poly_feat.GetFID()] = (
             shapely.wkb.loads(bytes(poly_geom_ref.ExportToWkb())))
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -2267,7 +2267,7 @@ def calculate_disjoint_polygon_set(
     for poly_feat in vector_layer:
         poly_geom_ref = poly_feat.GetGeometryRef()
         if poly_geom_ref is None:
-            LOGGER.warn(
+            LOGGER.warning(
                 f'no geometry in {vector_path} FID: {poly_feat.GetFID()}, '
                 'skipping...')
             continue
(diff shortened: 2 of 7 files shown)


pyproject.toml
```diff
@@ -14,3 +14,13 @@ requires = [
     'numpy==1.19.3; python_version=="3.9"',
     'numpy==1.21.2; python_version=="3.10"']
 build-backend = "setuptools.build_meta"
+
+
+[tool.pytest.ini_options]
+# Raise warnings to exceptions.
+filterwarnings = 'error'
```
emlys, line 21: Hooray for `pyproject.toml`!

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2267,10 +2269,16 @@ def calculate_disjoint_polygon_set(
...
+            LOGGER.warning(
                 f'no geometry in {vector_path} FID: {poly_feat.GetFID()}, '
                 'skipping...')
             continue
+        if poly_geom_ref.IsEmpty():
+            LOGGER.warning(
+                f'empty geometry in {vector_path} FID: {poly_feat.GetFID()}, '
+                'skipping...')
+            continue
+        LOGGER.info(poly_geom_ref.ExportToWkt())
```
emlys: Did you mean to leave this in? If so, maybe add a prefix to the logging message so it's clearer where it comes from?
```suggestion
        LOGGER.info('calculate_disjoint_polygon_set: adding polygon ' + poly_geom_ref.ExportToWkt())
```

emlys: Lines 2265-2267 say
```
# As much as I want this to be in a comprehension, a comprehension version
# of this loop causes python 3.6 to crash on linux in GDAL 2.1.2 (which is
# what's in the debian:stretch repos.)
```
since we're not supporting those versions any more, is it worth refactoring this?

tests/test_geoprocessing.py
```diff
@@ -753,6 +753,20 @@ def test_calculate_disjoint_polygon_set(self):
                 expected_value = row_index // 2 * n + col_index // 2
                 new_feature.SetField('expected_value', expected_value)
                 layer.CreateFeature(new_feature)
+
+        # Now create one additional feature that has no geometry in order to
+        # exercise a warning.
+        new_feature = ogr.Feature(layer_defn)
+        new_feature.SetField('expected_value', 0)
+        layer.CreateFeature(new_feature)
```
emlys, line 796: Do we also need to assert that the warning was raised?

emlys (changes requested):
@phargogh this looks good! Just had a couple questions

pyproject.toml
phargogh: This was a really exciting part of this!  A fun fact here is that the [pytest team have plans to make far better use of what `pyproject.toml` can offer](https://docs.pytest.org/en/6.2.x/customize.html#pyproject-toml), but until they do they're just porting over their old INI options here (hence the `tool.pytest.ini_options` table)

src/pygeoprocessing/geoprocessing.py
phargogh: > Did you mean to leave this in?

Ooo, yeah I think this'll be better to leave out.  Since logging flushes to a file on each write, this loop would really easily bottleneck with this logger message in there.  Removed in ae0d5ed

phargogh: > since we're not supporting those versions any more, is it worth refactoring this?

I think the loop has enough complexity now where a comprehension, while possible, would be less readable and less maintainable.  We'd have to kind of bend over backwards to shoehorn the same logger.warnings in there.  So if you're OK with it, I think I'd rather just remove the comment and then leave it as a standard for-loop.

Comment removed in  48ed37c

tests/test_geoprocessing.py
phargogh: It would be best to!  Added for these two features in this test in 676ef1f

phargogh:
OK!  I've responded to all of your comments @emlys and added some assertions for the logging along with a new context manager for capturing the logging and rendering it as a list of `logging.LogRecord` objects for access to the underlying attributes.  Back to you!

emlys (approved):
thanks @phargogh!

PR #250, opened 2022-03-21
Testing that ws_id will be copied over to delineated watersheds.

### LINKED ISSUE #249: Allow a ws_id field in watershed delineation
In InVEST, it's very useful to have a `ws_id` field because of the consistency with the routed models.  The current watershed delineation uses the `ws_id` field internally and then just doesn't copy its values over to the new vector.  It'd be great if we could use some other internal ID field instead.

Related to https://github.com/natcap/invest/issues/908

phargogh:
It turned out that the conditional I thought was preventing `ws_id` from being copied over had no meaningful effect at all.  I've updated the tests to verify that if there's a `ws_id` field in the source vector, it'll be copied over into the target vector.

Fixes #249

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -41,7 +41,7 @@ jobs:
         uses: conda-incubator/setup-miniconda@v2
         with:
           activate-environment: pyenv
-          auto-update-conda: false
+          auto-update-conda: true
           python-version: ${{ matrix.python-version }}
           channels: conda-forge
 
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -14,20 +14,32 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.7, 3.8, 3.9]
-        gdal: ["3.2.2", "3.3.0", "3.4.0"]
+        gdal: ["3.2.2", "3.3.0", "3.4.3", "3.5"]
         os: [ubuntu-latest, windows-latest, macos-latest]
 
         include:
           - python-version: "3.10"
-            gdal: "3.4.0"
+            gdal: "3.4.3"
             os: ubuntu-latest
 
           - python-version: "3.10"
-            gdal: "3.4.0"
+            gdal: "3.4.3"
             os: windows-latest
 
           - python-version: "3.10"
-            gdal: "3.4.0"
+            gdal: "3.4.3"
+            os: macos-latest
+
+          - python-version: "3.10"
+            gdal: "3.5"
+            os: ubuntu-latest
+
+          - python-version: "3.10"
+            gdal: "3.5"
+            os: windows-latest
+
+          - python-version: "3.10"
+            gdal: "3.5"
             os: macos-latest
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/pythonpackage.yml
```diff
@@ -59,8 +59,10 @@ jobs:
 
       - name: Install PyGeoprocessing
         run: |
-            conda install --file requirements.txt
-            conda install gdal==${{ matrix.gdal }} setuptools build
+            # Installing all of these in the same line will force conda to
+            # resolve all requirements at once, reducing the chance of DLL
+            # conflicts.
+            conda install --file requirements.txt gdal==${{ matrix.gdal }} setuptools build
```
emlys, line 77: oh interesting! I'd be curious to read about how conda resolves dependencies when done all at once vs. sequentially. do you happen to have a reference for that?

emlys (approved):
@phargogh it looks like the macos tests are still failing with the gdal dylib issue. The changes are unrelated to that, but maybe take out the comment in `pythonpackage.yml` if it's unclear that combining the conda commands is helpful? (keeping them all on one line is nice even if it doesn't fix the problem tho)

.github/workflows/pythonpackage.yml
phargogh: I don't have a documentation reference for it, but it does appear to be the case just from debugging this PR.  [Here's the latest build on `main`](https://pipelines.actions.githubusercontent.com/bGFNKdWfd0FvLn6TeVKOltVTPj0bb20eF1nGnpYGH1aTzpEgid/_apis/pipelines/1/runs/819/signedlogcontent/21?urlExpires=2022-03-24T23%3A21%3A07.3392228Z&urlSigningMethod=HMACV1&urlSignature=VLgds40U8QnPWloFlJDFKO%2B6wS%2FUKbkf9xcZsKGxX1Q%3D), where if you look at the logs you can see that the latest version of GDAL is being installed (3.4.2), but is then being downgraded to the build matrix-defined version of GDAL, 3.2.2.  With this all in one line, we _only_ install the one version of GDAL.

phargogh:
Thanks for taking a look @emlys !  I still think keeping them all in the same line is helpful, even if there's still this GDAL DLL issue. I'll look into the DLL issue a bit more and write back when I've found a fix.

phargogh:
@emlys I think this one is ready for re-review as well now that the mac tests are passing again.

.github/workflows/pythonpackage.yml
```diff
@@ -59,10 +59,19 @@ jobs:
...
             python -m build --wheel
             python -m pip install $(find dist -name "*.whl")
 
+      - name: Mac dylib hack
+        if: matrix.os == 'macos-latest'
+        run: |
+            # Specific fontconfig needed for binary build issues on mac.
+            # See natcap/pygeoprocessing#252
+            conda install fontconfig=2.13.1
```
dcdenu4: Oofta!

phargogh: Yep, not ideal at all, but until various packages are rebuilt on `conda-forge` it's a suitable workaround!

emlys: Sorry this slipped through the cracks! Now that it's been a while, do we still need this?

phargogh: To the best of my knowledge, we do not!  Removed in 8c7eaeb so let's see what happens in the tests.

emlys (approved):

PR #261, opened 2022-06-08
Using the correct numpy dtype in new_raster_from_base. 

### LINKED ISSUE #260: Speed up raster filling in new_raster_from_base() with correct array dtypes
In reading through `pygeoprocessing.new_raster_from_base`, I noticed that the band filling portion of the function is using `numpy.empty()` without using the correct array dtype, which means that in most cases we're allocating more memory than is needed for these arrays and there's a minor (but notable) performance penalty for doing this.  For example, if we only need to create an unsigned byte array, creating a float32 array will use 4 times the memory when it doesn't need to.

A minor point, but it's easy to implement so why not?

phargogh:
Such a minor change and really no changes observed in runtime in a reasonably-sized raster.  It's possible that some runtime improvements might be observed on very large rasters.

Fixes #260

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -14,20 +14,32 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.7, 3.8, 3.9]
-        gdal: ["3.2.2", "3.3.0", "3.4.0"]
+        gdal: ["3.2.2", "3.3", "3.4", "3.5"]
         os: [ubuntu-latest, windows-latest, macos-latest]
 
         include:
           - python-version: "3.10"
-            gdal: "3.4.0"
+            gdal: "3.4"
             os: ubuntu-latest
 
           - python-version: "3.10"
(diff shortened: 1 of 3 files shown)


dcdenu4 (approved):
Looks good! Thanks James!

dcdenu4:
I noticed that the mac tests are failing with some gdal related issues. I'm re-running one of the jobs now.

dcdenu4 (commented):
Thanks @phargogh , is there an issue associated with those gdal DLL issues that we can reference for our future selves?

phargogh:
> is there an issue associated with those gdal DLL issues that we can reference for our future selves?

@dcdenu4 sadly no, I don't think so.  The failure started seemingly randomly and I fixed it by trial-and-error.

I have a sneaking suspicion that this unpredictability in the `conda-forge` builds might be resolved by pinning the package versions (issue created for this at https://github.com/natcap/pygeoprocessing/issues/265), but this should be explored.

PR #262, opened 2022-06-08
Plugging shared_memory leak

### LINKED ISSUE #247: Shared memory objects leaking and being cleaned up on process shutdown
I recently noticed when running an InVEST model that I see the following warning very reliably:

```
resource_tracker.py:216: UserWarning: resource_tracker: There appear to be 28 leaked shared_memory objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '
```

To reproduce:
* Run an InVEST model (or a pygeoprocessing function that uses `shared_memory`, like `raster_calculator`
* Exit the process normally
* Observe warning

phargogh:
Objects were being created when they should not have been, and then sticking around because they shouldn't have been created in the first place.

I'm really not sure how to test this, but I can confirm that the shared memory warning is no longer present.

Fixes #247

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,9 @@ Release History
 
 Unreleased Changes
 ------------------
+* Fixing a memory leak in ``pygeoprocessing.raster_calculator`` where
+  shared memory objects were being inadvertently created when they should not
+  have been and then they were not subsequently destroyed.
 * ``calculate_disjoint_polygon_set`` will now skip over empty geometries.
   Previously, the presence of empty geometries would cause an error to be
   raised.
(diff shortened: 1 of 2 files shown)


davemfish (approved):
Easy! It looks like we're not compatible with gdal 3.4.0, but also looks like you already addressed that and `main` is passing. So I'll leave it to you to merge @phargogh

phargogh:
I just merged in the latest changes from main, so fingers crossed that resolves the tests.

phargogh:
Hey, looks like tests are passing!  I'll go ahead and merge this, then.

Issue #266, opened 2022-07-01
Pylint `no-member` errors on pygeoprocessing module


phargogh:
When I run `pylint` on a program that uses pygeoprocessing, I get a pile of `no-member` errors when I shouldn't.  For example:

```
src/natcap/invest/scenic_quality/scenic_quality.py:895:22: E1101: Module 'pygeoprocessing' has no 'get_raster_info' member (no-member)
src/natcap/invest/scenic_quality/scenic_quality.py:898:4: E1101: Module 'pygeoprocessing' has no 'new_raster_from_base' member (no-member)
src/natcap/invest/scenic_quality/scenic_quality.py:914:22: E1101: Module 'pygeoprocessing' has no 'iterblocks' member (no-member)
src/natcap/invest/scenic_quality/scenic_quality.py:981:18: E1101: Module 'pygeoprocessing' has no 'get_raster_info' member (no-member)
src/natcap/invest/scenic_quality/scenic_quality.py:1004:4: E1101: Module 'pygeoprocessing' has no 'raster_calculator' member (no-member)
src/natcap/invest/scenic_quality/scenic_quality.py:1030:4: E1101: Module 'pygeoprocessing' has no 'raster_calculator' member (no-member)
```

phargogh:
Resolved in #267

PR #267, opened 2022-07-01
Fixing module-level imports to work with pylint and JEDI-vim

phargogh:
This is a minor thing, but it's been a mild issue for _years_.  So it's fixed now and my autocomplete is working as expected in `vim` and PyLint is able to find the right attributes.

I'm really not sure how to test this, so the tests aren't updated.

RE:#266

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -79,4 +79,4 @@ jobs:
             python -m pip install $(find dist -name "*.whl")
 
       - name: Test with pytest
-        run: pytest
+        run: python -m pytest tests
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,10 @@ Release History
 
 Unreleased Changes
 ------------------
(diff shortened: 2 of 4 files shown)


dcdenu4 (approved):
Thanks @phargogh , kind of annoying that we have to manually keep this list. I'm not sure why the tests are failing. I tested locally with Python 3.9, gdal 3.4.3 and could not reproduce the error.

phargogh:
@dcdenu4 I agree about the unfortunate nature of needing to manually expose things at that level, but I'm not sure that there's a way around it.  Even in the python stdlib, a bunch of the standard packages do the same thing so I guess this is normal.  As frustrating as it is, it feels like python is trying to give us granular control of what's exposed at the module level, which makes sense.

RE: Test failures, it turned out that numpy 1.23.0 (released recently) changed a warning message which meant the existing warning filters were invalid in one of the tests.  Both warning message formats are now handled.

dcdenu4 (approved):
Thanks for handling that @phargogh !

PR #271, opened 2022-08-10
Add value counts under polygons in zonal statistics

### LINKED ISSUE #270: Proposal: optionally return unique pixel counts under each polygon
The InVEST HRA model has a couple of columns in an output table (`R_%LOW`, `R_%MED`, `R_%HIGH` in `SUMMARY_STATISTICS.csv`) that are derived from the pixel counts of each classification in a raster that has pixel values in the set `{0, 1, 2, 3}`.  As it stands, the only way to get accurate reports of this information is to basically reimplement `zonal_statistics`, adding in the counting.  These outputs cannot be derived from the current output of `zonal_statistics`.

It would be very handy if `pygeoprocessing.zonal_statistics` allowed for pixel counts under each polygon to be reported.  The assumption is that this only makes sense on discrete (integer) rasters. Recording these values on a floating-point raster could easily exhaust available memory, which is why this is best kept as an optional parameter that defaults to `False`.

I propose modifying the function signature of `pygeoprocessing.zonal_statistics` to include a new parameter, `include_pixel_counts`:

```python

def zonal_statistics(
        base_raster_path_band, aggregate_vector_path,
        aggregate_layer_name=None, ignore_nodata=True,
        polygons_might_overlap=True, include_pixel_counts=False,
        working_dir=None):
```

Using the HRA example, the return value for FID `60` would be:  11, 62, 4
```python
{60: {
    'min': 1,
    'max': 3,
    'sum': 147,
    'count': 77,
    'nodata_count': 0,
    'pixel_counts': {
        1: 11,
        2: 62,
        3: 4,
    }
}
```

phargogh:
This PR adds pixel counts under each polygon and warns the user if pixel counts have been requested on a floating-point raster.

Fixes #270

### DIFF
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -1160,7 +1160,7 @@ def zonal_statistics(
         There may be some degenerate cases where the bounding box vs. actual
         geometry intersection would be incorrect, but these are so unlikely as
         to be manually constructed. If you encounter one of these please email
-        the description and dataset to richsharp@stanford.edu.
+        the description and dataset to jdouglass@stanford.edu.
 
     Args:
         base_raster_path_band (tuple): a str/int tuple indicating the path to
@@ -1486,7 +1486,7 @@ def zonal_statistics(
         # is as small as a pixel. There could be some degenerate cases that
         # make this estimation very wrong, but we do not know of any that
         # would come from natural data. If you do encounter such a dataset
-        # please email the description and datset to richsharp@stanford.edu.
+        # please email the description and datset to jdouglass@stanford.edu.
         unset_fid_block = clipped_band.ReadAsArray(
             xoff=xoff, yoff=yoff, win_xsize=win_xsize, win_ysize=win_ysize)
 
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -1196,9 +1196,9 @@ def zonal_statistics(
         of 'min' 'max' 'sum' 'count' and 'nodata_count'.  Example::
 
             {0: {'min': 0,
-                 'max': 1,
-                 'sum': 1.7,
-                 'count': 3,
+                 'max': 14,
+                 'sum': 42,
+                 'count': 8,
                  'nodata_count': 1,
                  'value_counts': {
                      2: 5,
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -1526,7 +1526,7 @@ def zonal_statistics(
         if include_value_counts:
             aggregate_stats[unset_fid]['value_counts'].update(
                 dict(zip(*numpy.unique(
-                    value_unset_fid_block, return_counts=True))))
+                    valid_unset_fid_block, return_counts=True))))
 
     unset_fids = aggregate_layer_fid_set.difference(aggregate_stats)
     LOGGER.debug(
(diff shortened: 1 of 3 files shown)


phargogh:
@emlys this should be ready to review now!

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1234,12 +1244,39 @@ def zonal_statistics(
...
+            "values.")
+
     # -1 here because bands are 1 indexed
     raster_nodata = raster_info['nodata'][base_raster_path_band[1]-1]
     temp_working_dir = tempfile.mkdtemp(dir=working_dir)
     clipped_raster_path = os.path.join(
         temp_working_dir, 'clipped_raster.tif')
 
+    def _make_sample_dict():
+        """Build a sample aggregate dict.
```
emlys: I'm not sure I'm understanding why it's a "sample" dict?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1234,12 +1244,39 @@ def zonal_statistics(
...
+
+        Returns:
+            A sample aggregate dict with starter values, used for a
+            ``collections.defaultdict`` when a key does not already exist.
+        """
+        sample_aggregate_dict = {
+            'min': None, 'max': None, 'count': 0, 'nodata_count': 0,
+            'sum': 0.0}
+        if include_value_counts:
+            sample_aggregate_dict['value_counts'] = collections.Counter()
```
emlys: neat!

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1194,7 +1199,12 @@ def zonal_statistics(
                  'max': 1,
                  'sum': 1.7,
                  'count': 3,
-                 'nodata_count': 1
+                 'nodata_count': 1,
+                 'value_counts': {
+                     2: 5,
+                     4: 1,
+                     14: 2,
```
emlys, line 1206: minor, but it would be nice if these example statistics were consistent with each other

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1487,6 +1524,10 @@ def zonal_statistics(
         aggregate_stats[unset_fid]['count'] = valid_unset_fid_block.size
         aggregate_stats[unset_fid]['nodata_count'] = numpy.count_nonzero(
             unset_fid_nodata_mask)
+        if include_value_counts:
+            aggregate_stats[unset_fid]['value_counts'].update(
```
emlys, line 1532: this might be worth a comment that `aggregate_stats[unset_fid]['value_counts']` is a `collections.Counter` and so `update` adds to it rather than overwrites.

emlys (changes requested):

src/pygeoprocessing/geoprocessing.py
phargogh: Yeah, "sample" might not be the best terminology here ... "new" is better.  I've revised this in 8aa8227 to clarify what this is used for.  "sample" was borrowed from the prior implementation where the term maybe made a little more sense based on how it was used, as a starter value.

src/pygeoprocessing/geoprocessing.py
phargogh: I think this is the first time I've maybe ever needed to use a counter!  Pretty exciting.

src/pygeoprocessing/geoprocessing.py
phargogh: Yes it would!  Patched in e32064c

src/pygeoprocessing/geoprocessing.py
phargogh: I agree, patched in 7dccc77

emlys (approved):
thanks @phargogh!

PR #282, opened 2022-11-19
Add filename to raster_calculator logging

### LINKED ISSUE #281: Add the filename to raster_calculator progress logging
`raster_calculator`'s progress logging works great in a single-threaded context, as we often (but not always) have in InVEST.  There are plenty of other cases, particularly when we start adding CPUs and doing multiple `raster_calculator` operations simultaneously where simple logging like `INFO:pygeoprocessing.geoprocessing:86.6% complete` just isn't clear enough.

I think it would be very worthwhile to _at least_ include the target filename in the logging message so that we can distinguish between multiple progress logging messages that happen to be logged at or about the same time.

phargogh:
This has been bothering me for a while.  Sorry for the very short PR!  Didn't really seem like a test was necessary for something like this.

Fixes #281

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,8 @@ Release History
 
 Unreleased Changes
 ------------------
+* Adding the target filename to progress logging in
+  ``pygeoprocessing.raster_calculator``.
 * ``pygeoprocessing.zonal_statistics`` will now optionally include a count of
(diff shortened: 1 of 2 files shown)


dcdenu4 (approved):
That does seem handy. Thanks @phargogh.

PR #284, opened 2022-12-02
Explicitly support Python 3.11

phargogh:
This PR adds support for python 3.11 without removing support for any existing versions.


dcdenu4 (approved):
Thanks @phargogh, looks good.

PR #287, opened 2022-12-09
Feature/269 allow touching nonoverlapping polygons in disjoint polygon set

### LINKED ISSUE #269: Optionally allow touching but nonoverlapping polygons in `calculate_disjoint_polygon_set`
In working on Urban Nature Access, we had a case where we had a large cluster of polygons with boundaries that touched but whose interiors were all distinct.  Using the current implementation of `calculate_disjoint_polygon_set`, there were 8 groupings of polygons. I propose including an optional flag in `calculate_disjoint_polygon_set` to not consider touching polygons as intersecting.

The scope of work here would be a couple hours to implement, test and PR in.

phargogh:
This PR introduces a flag to `pygeoprocessing.disjoint_polygon_set` that allows geometries with intersecting boundaries but nonintersecting interiors to be part of the same polygon subset.

Fixes #269

### DIFF
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -1183,8 +1183,8 @@ def zonal_statistics(
             aggregation coverage close to optimally by rasterizing sets of
             polygons that don't overlap.  However, this step can be
             computationally expensive for cases where there are many polygons.
-              this flag to False directs the function rasterize in one
-            step.
+            Setting this flag to ``False`` directs the function rasterize in
+            one step.
         include_value_counts (boolean): If True, the function tallies the
             number of pixels of each value under the polygon.  This is useful
             for classified rasters but could exhaust available memory when run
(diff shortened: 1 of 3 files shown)


src/pygeoprocessing/geoprocessing.py
```diff
@@ -2366,6 +2371,12 @@ def calculate_disjoint_polygon_set(
         else:
             polygon = poly_geom
         for intersect_poly_fid in possible_intersection_set:
+            # If geometries touch (share 1+ boundary point), then do not count
+            # it as an intersection.
+            if geometries_may_touch and polygon.touches(
+                    shapely_polygon_lookup[intersect_poly_fid]):
+                continue
```
davemfish, line 2378: Cool, so "touching" objects actually do not intersect, by definition.

davemfish (approved):
Hey @phargogh , this looks good to me, thanks!

PR #289, opened 2022-12-13
Raise ValueError if get_gis_type results in UNKOWN_TYPE

### LINKED ISSUE #244: pygeoprocessing.get_gis_type should raise an exception if gdal can't open the file
Throughout pygeoprocessing, we do some basic checking for the case where gdal can't open up a dataset.  `get_raster_info` and `get_vector_info` both do this, and it's reasonable to expect that `get_gis_type` would also do this.

dcdenu4:
`get_gis_type` returns `gdal.UNKOWN_TYPE` (0) if the file cannot be opened as a raster or vector. This is inconsistent with the rest of the API including `get_raster_info` and `get_vector_info` which raise a `ValueError` in that circumstance. It's also not intuitive that a Microsoft Word file would return a GIS type of `UNKOWN_TYPE`. PGP is a GIS processing library that only works with valid GIS types.

Fixes #244

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -10,6 +10,9 @@ Unreleased Changes
   will be logged when invoked on floating-point rasters, as using this on
   continuous rasters can result in excessive memory consumption. To use this
   feature, set ``include_value_counts=True`` when calling ``zonal_statistics``.
+* ``pygeoprocessing.get_gis_type`` will now raise a ``ValueError`` if the file
+  cannot be opened as ``gdal.OF_RASTER`` or ``gdal.OF_VECTOR``.
+  https://github.com/natcap/pygeoprocessing/issues/244
 
 2.3.4 (2022-08-22)
 ------------------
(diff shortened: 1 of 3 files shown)


dcdenu4:
@phargogh, I'd be curious your thoughts on keeping `UNKNOWN_TYPE` around as a constant in `__init__.py`. In PGP, `get_gis_type` is the only place it's used, and InVEST only uses it in HRA, directly related to a `get_gis_type` call.

phargogh:
@dcdenu4 much to my surprise, GDAL does actually support drivers that are neither rasters nor vectors!  So yes, let's keep the `UNKNOWN_TYPE` flag.

Here's my rationale:

```python
from osgeo import gdal

for i in range(gdal.GetDriverCount()):
    try:
        metadata = gdal.GetDriver(i).GetMetadata()
    except Exception as error:
        print(error, 'ERROR')
    else:
        raster_capable = 'DCAP_RASTER' in metadata
        vector_capable = 'DCAP_VECTOR' in metadata
        if not raster_capable and not vector_capable:
            print(metadata['DMD_LONGNAME'])
```
which outputs:
```
Geographic Network generic file based model
Geographic Network generic DB based model
```

phargogh (approved):

phargogh:
Source code looks good to me; I'm just waiting for the remaining tests to finish.  There were a couple of conda `500` internal server errors, which are fine and flaky.

PR #292, opened 2022-12-19
Implement ``raster_reduce``

### LINKED ISSUE #285: Proposal: `raster_reduce` for raster operations that don't produce a raster
It could be useful to have an equivalent of `raster_calculator` for operations that don't produce a raster output. If `raster_calculator` is like `map` over the blocks of a raster, then this would be like `functools.reduce` over the blocks of a raster.

Aggregations are probably the most common use case. For the Morgan Stanley project I needed to produce histogram data for a raster. `raster_calculator` seems like the easiest approach, but it requires writing to a target raster which was totally irrelevant for my use case. I ended up with:
```python
def raster_histogram(raster_path, bins=10):
    """Return histogram data for the given raster divided into bins."""

    histogram = numpy.zeros(bins)

    def add_to_histogram(array):
        nonlocal histogram
        histogram += numpy.histogram(array, bins=bins)[0]
        return array

    pygeoprocessing.raster_calculator(
        [(raster_path, 1)],
        add_to_histogram,
        'foo.tif',
        gdal.GDT_Int32,
        TARGET_NODATA
    )
    return histogram
```

Given a proposed new function `raster_reduce(raster_path, function, initializer)` that behaves like [`functools.reduce`](https://docs.python.org/3/library/functools.html#functools.reduce), that could be rewritten as

```python
def raster_histogram(raster_path, bins=10):
    """Return histogram data for the given raster divided into bins."""

    def add_to_histogram(array, histogram):
        return histogram + numpy.histogram(array, bins=bins)[0]

    return pygeoprocessing.raster_reduce(
        [(raster_path, 1)],
        add_to_histogram,
        numpy.zeros(bins))
```

I'm not sure if there are any use cases for this function in invest, but I imagine it would be useful for data analysis.

emlys:
Fixes #285 
I didn't include any error handling like `raster_calculator` has, so the implementation is very simple. Let me know if you think it's missing anything!

### DIFF
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -624,6 +624,7 @@ def raster_reduce(function, raster_path_band, initializer, mask_nodata=True,
             flattened array of non-nodata pixels from each block is passed to
             the ``function``. if False, each block is passed directly to the
             ``function`` without masking.
+        largest_block (int): largest block parameter to pass to ``iterblocks``
 
     Returns:
         aggregate value, the final value returned from ``function``
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -535,6 +535,59 @@ def raster_calculator(
                 pass
 
 
+def raster_reduce(function, raster_path, initializer, band=1):
+    """Cumulatively apply a reducing function to each block of a raster.
+
+    This effectively reduces the entire raster to a single value, but it works
+    by blocks to be memory-efficient.
+
+    The ``function`` signature should be ``function(aggregator, block)``, where
+    ``aggregator`` is the aggregated value so far, and ``block`` is a 2D numpy
+    array containing the data from the block to reduce next.
+
+    ``function`` is called once on each block. On the first ``function`` call,
+    ``aggregator`` is initialized with ``initializer``. The return value from
+    each ``function`` call is passed in as the ``aggregator`` argument to the
+    subsequent ``function`` call. When all blocks have been reduced, the return
+    value of the final ``function`` call is returned.
+
+    Example:
+        Calculate the sum of all values in a raster::
+
+            raster_reduce(lambda total, block: total + numpy.sum(block),
+                          raster_path, 0)
+
+        Calculate a histogram of all values in a raster::
+
+            def add_to_histogram(histogram, block):
+                return histogram + numpy.histogram(block, bins=10)[0]
+
+            raster_reduce(add_to_histogram, raster_path, numpy.zeros(10))
+
+        Calculate the sum of all values in a raster, excluding nodata::
+
+            nodata = pygeoprocessing.get_raster_info(raster_path)['nodata'][0]
+            def sum_excluding_nodata(total, block):
+                return total + numpy.sum(block[block != nodata])
+
+            raster_reduce(sum_excluding_nodata, raster_path, 0)
+
+    Args:
+        raster_path (str): path of the raster to reduce
+        function (func): function to apply to each raster block
+        initializer (obj): value to initialize the aggregator for the
+            first function call
+        band (int): index of the raster band to use. defaults to 1.
+
+    Returns:
+        aggregate value, the final value returned from ``function``
+    """
+    aggregator = initializer
+    for (_, block) in iterblocks((raster_path, band)):
+        aggregator = function(aggregator, block)
+    return aggregator
+
+
 def align_and_resize_raster_stack(
         base_raster_path_list, target_raster_path_list, resample_method_list,
         target_pixel_size, bounding_box_mode, base_vector_path_list=None,
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -590,7 +590,7 @@ def raster_reduce(function, raster_path_band, initializer,
     for (_, block) in iterblocks(raster_path_band,
                                  largest_block=largest_block):
         aggregator = function(aggregator, block)
-        pixels_processed += block.shape[0] * block.shape[1]
+        pixels_processed += block.size
         last_time = _invoke_timed_callback(
             last_time, lambda: LOGGER.info(
                 f'{raster_path_band[0]} reduce '
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -535,7 +535,8 @@ def raster_calculator(
                 pass
 
 
-def raster_reduce(function, raster_path, initializer, band=1):
+def raster_reduce(function, raster_path_band, initializer,
+                  largest_block=_LARGEST_ITERBLOCK):
     """Cumulatively apply a reducing function to each block of a raster.
 
     This effectively reduces the entire raster to a single value, but it works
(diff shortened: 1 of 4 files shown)


src/pygeoprocessing/geoprocessing.py
```diff
@@ -535,6 +535,59 @@ def raster_calculator(
                 pass
 
 
+def raster_reduce(function, raster_path, initializer, band=1):
```
phargogh: I can see the parallels to python's reduce, which is neat!  A couple thoughts:

1. What would you think about using the `(path, band)` notation, for consistency with other raster operations in this file?
2. Would it be worth having the path/band tuple as the first parameter?  I see that most functions in `pygeoprocessing` that involve rasters do this, but also the function as it stands is more like the stdlib reduce.  What do you think?
3. One thing that's interesting and convenient about the `functools` reduce is that the `initializer` is optional.  Would that be worth including here as well?
4. Since this is a wrapper around `iterblocks`, I wonder if it would be helpful to also expose the `largest_block` parameter, in case the user knows something about their system and wants to fiddle with the block size.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -535,6 +535,59 @@ def raster_calculator(
...
+        function (func): function to apply to each raster block
+        initializer (obj): value to initialize the aggregator for the
+            first function call
+        band (int): index of the raster band to use. defaults to 1.
+
+    Returns:
+        aggregate value, the final value returned from ``function``
+    """
+    aggregator = initializer
+    for (_, block) in iterblocks((raster_path, band)):
```
phargogh: Thinking about `raster_calculator` and how long it can take to iterate through potentially very large rasters, maybe some timed progress logging would be helpful here?

tests/test_geoprocessing.py
```diff
@@ -4833,3 +4833,21 @@ def test_geometry_field_mismatch(self):
...
+    def test_raster_reduce(self):
+        """PGP: test raster_reduce can calculate a sum."""
+        array = numpy.ones((512, 512))
+        raster_path = os.path.join(self.workspace_dir, 'raster.tif')
+        _array_to_raster(array, -1, raster_path)
+
+        def sum_blocks(total, block): return total + numpy.sum(block)
+        spy_sum_blocks = unittest.mock.Mock(wraps=sum_blocks)
+        result = pygeoprocessing.raster_reduce(spy_sum_blocks, raster_path, 0)
+        self.assertEqual(result, 512*512)
```
phargogh: Since this is derived from `array`, maybe we could just use `numpy.sum(array)` to make it clear that we're just summing pixel values?

tests/test_geoprocessing.py
```diff
@@ -4833,3 +4833,21 @@ def test_geometry_field_mismatch(self):
...
+
+        for i, (_, (total, block), _) in enumerate(spy_sum_blocks.mock_calls):
```
phargogh, line 4873: I think I get what this is doing, but maybe a comment would help make the intent clear?

tests/test_geoprocessing.py
```diff
@@ -4833,3 +4833,21 @@ def test_geometry_field_mismatch(self):
...
+            self.assertEqual(total, i * 256 * 256)
+            numpy.testing.assert_array_equal(block, numpy.ones((256, 256)))
+
+
+
```
phargogh: Bonus lines?

phargogh (commented):
I mostly just have a bunch of questions/comments about the API and a couple other things, let me know what you think!

tests/test_geoprocessing.py
emlys: good idea!

src/pygeoprocessing/geoprocessing.py
emlys: 1: Sure!
2: I did put the arguments in this order to resemble `functools.reduce`, but consistency with pygeoprocessing is an equally valid reson. I don't really have a preference either way. 
4: Good idea!

emlys: 3: I think the `initializer` is necessary with the current design, but a maybe not with a different design.

With `functools.reduce(function, iterable, initializer)`, the initializer is optional because you can infer it. If `initializer` isn't provided, it's set to the return value of the first `function` call. 

But `raster_reduce` has two parts: it aggregates each block to a single value and then reduces _those_ values to a single value. In the current implementation, the `function` does both of those operations. For instance with the histogram example, the correct value of `initializer` to infer would be the result of `numpy.histogram` called on the first block. But there's no way to know that because `numpy.histogram` is wrapped up in the `function`.

But it would be possible if we split out the block aggregation and reducing operations into separate functions, something like:
```
def raster_reduce(raster_path, block_aggregate_fn, raster_reduce_fn, initializer=None):
    # reduce each block to a single value
    block_values = map(block_aggregate_fn, iterblocks(raster_path))
    if initializer is not None:
        block_values = [initializer] + block_values
    # reduce that series of block values to a single value
    return functools.reduce(raster_reduce_fn , block_values)
```
Then the histogram example could be written as
```
def block_histogram(block): return numpy.histogram(block, bins=10)[0]
def sum(x, y): return x + y
raster_reduce(raster_path, block_histogram, sum)
```
This is kind of hard to wrap my head around, does that make any sense? And would that alternative implementation be easier to use?

phargogh: Aha!  Yes, you're absolutely right.  Thanks for walking me through this.  The return type of `function` might not be an array the size and shape (or even type) of `block`, so in general we can't just take the first block and use that as the initializer.

Ah, higher-order functions are so interesting!  I think we should stick with your original formulation with the initializer so that we minimize the number of functions one has to understand in order to get this working in code.  Conceptually the combination of map and reduce in the alternate implementation makes an interesting distinction, but I think the original implementation might be more straightforward.

phargogh: I had one other thought that I wonder what you think about:  would it make sense or be useful to optionally filter out nodata values?  We could always write our reduce function to handle nodata, but that requires knowing the nodata ahead of time when it could easily be read and the block indexed within `raster_reduce`.  If you agree, would it also make sense to flatten out the block for cases where we are not doing nodata filtering?  Just a thought!

src/pygeoprocessing/geoprocessing.py
```diff
@@ -535,6 +535,72 @@ def raster_calculator(
...
+    """
+    aggregator = initializer
+    last_time = time.time()
+    pixels_processed = 0
+    x_size, y_size = get_raster_info(raster_path_band[0])['raster_size']
+    n_pixels = x_size * y_size
+    for (_, block) in iterblocks(raster_path_band,
+                                 largest_block=largest_block):
+        aggregator = function(aggregator, block)
+        pixels_processed += block.shape[0] * block.shape[1]
```
phargogh: Such a minor thing, but slightly cheaper :)

```suggestion
        pixels_processed += block.size
```

phargogh (changes requested):
This is looking really cool @emlys !  And thanks for walking me through the need for the initializer, that totally makes sense that we need it.  I just had one trivial tweak to suggest and a question about whether it'd make sense to handle nodata ... let me know what you think!

src/pygeoprocessing/geoprocessing.py
emlys: Good idea!

emlys:
@phargogh I added a nodata masking option like you suggested, and made that the default. For consistency I also `flatten`ed the block if `mask_nodata` is False. Let me know what you think!

phargogh (approved):
Looks great!  I agree, flattening for consistency sounds right to me.

Just waiting until tests pass to merge!

emlys:
Merging now that tests passed

emlys:
Github is glitching here... the merge failed even though there were no conflicts, so I merged locally, then it marked the PR as closed.

PR #293, opened 2022-12-19
use oldest-supported-numpy meta package in pyproject.toml

emlys:
Use the [`oldest-supported-numpy`](https://github.com/scipy/oldest-supported-numpy) package to manage the oldest compatible numpy version for each python version. This is a meta-package developed by the scipy org. It saves us the trouble of keeping `pyproject.toml` correct and up to date when we add support for new python versions. We had the incorrect numpy version for python 3.10 anyway.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,14 +6,7 @@
 # that we can provide a much easier build experience so long as GDAL is
 # available at runtime.
 requires = [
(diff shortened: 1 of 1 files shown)


phargogh (approved):
Exciting!  I'm just waiting for the tests to complete before merging.

phargogh:
These test failures should be addressed with the changes already in `main`, so I'm going to take the plunge and merge this in.

PR #294, opened 2022-12-19
Removing x86 builds from github actions.

### LINKED ISSUE #232: Remove x86 windows wheels from the build workflow?
We still build x86 wheels for windows, but it's unclear if these are actually useful.  InVEST is now an x64 binary and even `conda-forge` has stopped building x86 packages.  Here are a few questions that come to mind that are worth considering:

* Does anyone at natcap still use x86 python who needs to use pygeoprocessing?
* How do the PyPI download stats for the x86 pygeoprocessing wheels compare to the x64 wheels?  [PyPInfo](https://github.com/ofek/pypinfo/blob/master/README.rst) looks useful for this.

phargogh:
It turns out that about 0.28% of the total downloads of `pygeoprocessing` from PyPI have been for win32, so it makes sense to remove those wheels from our GHA workflows and distribution processes so that the GHA build time can be put to better use.

Fixes #232

### DIFF
--- a/.github/workflows/build-py-dists.yml
+++ b/.github/workflows/build-py-dists.yml
@@ -9,7 +9,7 @@ jobs:
             matrix:
                 os: [windows-latest, macos-latest]
                 python-version: [3.7, 3.8, 3.9, "3.10", "3.11"]
-                python-arch: x64
+                python-arch: [x64]
 
         steps:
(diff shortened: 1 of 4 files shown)


emlys (approved):
Great! Just waiting for tests to pass.

PR #295, opened 2022-12-20
Add a D8 stream extraction function to `pygeoprocessing.routing`

### LINKED ISSUE #272: Proposal: Write a stream extraction function for D8
I recently wrote a script for someone where I had to also write out the raster_calculator op for filtering by a TFA. It's not a hard function to write, but it is something that we do in multiple places in InVEST.

I propose that we roll this in to `pygeoprocessing.routing` and consolidate the usage of it throughout InVEST. Here are the obvious cases where a TFA filter is used:

* https://github.com/natcap/invest/blob/main/src/natcap/invest/delineateit/delineateit.py#L299
* https://github.com/natcap/invest/blob/main/src/natcap/invest/routedem.py#L125

NDR, SDR, SWY use MFD and so are using the existing stream extraction function.

phargogh:
This PR adds a new function, `pygeoprocessing.routing.extract_streams_d8`, which should be helpful for users working with D8 routing since one won't need to implement their own D8 stream thresholding function.

Fixes #272

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,11 @@ Release History
 
 Unreleased Changes
 ------------------
+* Added a D8 stream extraction function at
+  ``pygeoprocessing.routing.extract_streams_d8`` which takes a D8 flow
+  accumulation raster and a flow accumulation threshold, setting all pixels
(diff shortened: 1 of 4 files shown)


davemfish (approved):
Very nice!

PR #297, opened 2022-12-31
Add a function to build overviews

### LINKED ISSUE #280: Proposal: add a function to build overviews
We currently have no convenient way to build overviews within pygeoprocessing, despite their utility, and the gdal raster method to build overviews does not really discuss what the "levels" should be or what those mean.  A helper function could make it easier to compute overviews and have the logging directed to the usual logging stream.

I propose adding a function to take care of this with the following features:

* Allow overviews to be built at reasonable levels if no override levels are defined.
* Allow geotiffs to have internal overviews created instead of external overviews.
* Allow the interpolation scheme to be defined for the overview creation.

A prototype of most of the above is available at https://github.com/phargogh/prototype-dem-tile-fetcher-and-processor/blob/main/hydrosheds-global-3s-v1-con.py#L31

The primary use case I'm thinking about for overviews is to speed up visualization of very large raster (potentially gigabytes) in a desktop GIS software like QGIS.

phargogh:
This PR adds a function to build overviews with a couple of parameters:

* Whether the overviews should be internal or external
* The resampling algorithm to use
* Whether to overwrite existing overviews if they exist
* The overview levels to use, or if the function should just generate a bunch of them.

In addition to having a function to build overviews, I also extended `get_raster_info()` to include a new `overviews` key that lists the dimensions of the overviews.  This is similar to what we see in the output of `gdalinfo`.  I also added another missing `file_list` parameter that was present in `get_raster_info()`'s output but not in the docstring.

Fixes #280

### DIFF
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -4110,7 +4110,7 @@ def build_overviews(raster_path_band_tuple, internal=False,
 
     LOGGER.debug(f"Using overviews {overview_scales}")
     result = raster.BuildOverviews(
-        pszResampling=resample_method,
+        resampling=resample_method,
         overviewlist=overview_scales,
         callback=overviews_progress
     )
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -2197,7 +2197,8 @@ def warp_raster(
 
     if resample_method.lower() not in _GDAL_WARP_ALGORITHMS:
         raise ValueError(
-            f'Invalid resample method: "{resample_method}"')
+            f'Invalid resample method: "{resample_method}". '
+            f'Must be one of {_GDAL_WARP_ALGOS_FOR_HUMAN_EYES}')
 
     gdal.Warp(
         warped_raster_path, base_raster,
@@ -4177,7 +4178,8 @@ def build_overviews(
     """
     if resample_method.lower() not in _GDAL_WARP_ALGORITHMS:
         raise ValueError(
-            f'Invalid overview resample method: "{resample_method}"')
+            f'Invalid overview resample method: "{resample_method}". '
+            f'Must be one of {_GDAL_WARP_ALGOS_FOR_HUMAN_EYES}')
 
     def overviews_progress(*args, **kwargs):
         pct_complete, name, other = args
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -25,6 +25,7 @@ import shapely.ops
 import shapely.prepared
 import shapely.wkb
 from osgeo import gdal
+from osgeo import gdalconst
 from osgeo import ogr
 from osgeo import osr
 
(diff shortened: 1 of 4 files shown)


src/pygeoprocessing/geoprocessing.py
```diff
@@ -4065,6 +4078,108 @@ def _mult_op(base_array, base_nodata, scale, datatype):
...
 
 
+def build_overviews(
+        raster_path, internal=False, resample_method='nearest',
+        overwrite=False, levels='auto'):
+    """Build overviews for a raster dataset.
+
+    Args:
+        raster_path (str): A path to a raster on disk and the band index to
+            build overviews for.
```
dcdenu4: Minor discrepancy between the text and the input. "and the band index to build ..." but the input is just a string. Should this be `raster_path_band (tuple)`?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -4065,6 +4078,108 @@ def _mult_op(base_array, base_nodata, scale, datatype):
...
+
+        Generate overviews for 4 levels, at 1/2, 1/4, 1/8 and 1/16 the
+        resolution::
+
+            build_overviews(raster_path, levels=[2, 4, 8, 16])
+
+    Returns:
+        ``None``
+    """
+    # WarpOptions.this is None when an invalid option is passed, and it's a
```
dcdenu4: This is a nice tidbit. Are we able to get the possible resample methods from GDAL and provide those in the error message?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -4065,6 +4078,108 @@ def _mult_op(base_array, base_nodata, scale, datatype):
...
+
+    def overviews_progress(*args, **kwargs):
+        pct_complete, name, other = args
+        percent = round(pct_complete * 100, 2)
+        if time.time() - overviews_progress.last_progress_report > 5.0:
+            LOGGER.info(f"Overviews progress: {percent}%")
+            overviews_progress.last_progress_report = time.time()
+    overviews_progress.last_progress_report = time.time()
+
+    raster_path = raster_path
```
dcdenu4: Is this necessary or handling a local scope case?

dcdenu4 (changes requested):
Thanks @phargogh, I just had a few questions / comments.

src/pygeoprocessing/geoprocessing.py
phargogh: Whoops!  Forgot to update this message when I changed it from `(path, band)` to just `path`.  I couldn't find an obvious way to tell GDAL to only generate overviews for a single band, so now the function just does it for the whole raster.  Patched in 5678032

src/pygeoprocessing/geoprocessing.py
phargogh: It is not necessary!  This is left over from a find/replace and is removed in 5678032.

src/pygeoprocessing/geoprocessing.py
phargogh: This ended up being kind of neat!  GDAL doesn't have something like a module-level mapping to get the human-readable types, and the mapping of `gdal.GRA_*` to human-readable names is [local to the `gdal.WarpOptions` parsing function.](https://github.com/OSGeo/gdal/blob/bfe1cfea364a0e185a792db67db7de2fc6aad902/swig/python/osgeo/gdal.py#L854).  I put together something in 7bdd1dbc8a4fc62015c0a737e4712b5022b75708 to get those human-readable names out of `gdal.WarpOptions` (avoiding `ast` inspection in case we can't read the source code like in pyinstaller binaries) and making them available at the module level.  This should make pygeoprocessing resilient to any changes to the set of resample algorithms GDAL offers, so long as they keep `gdal.WarpOptions` up to date!

454dda32c9251191f8f882b1b7a803d594e11693 takes this a step further and uses these human-readable resample algorithms in the docstrings where they are used.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -80,6 +81,23 @@ def __init__(self, missing_values, raster_path, value_map):
...
+# those strings is to do this conversion of gdalconst.GRA_* types to the
+# human-readable labels via gdal.WarpOptions like so.
+_GDAL_WARP_ALGORITHMS = []
+for _warp_algo in (_attrname for _attrname in dir(gdalconst)
+                   if _attrname.startswith('GRA_')):
+    # Appends ['-r', 'near'] to _GDAL_WARP_ALGORITHMS if _warp_algo is
+    # gdalconst.GRA_NearestNeighbor.  See gdal.WarpOptions for the dict
+    # defining this mapping.
+    gdal.WarpOptions(options=_GDAL_WARP_ALGORITHMS,
+                     resampleAlg=getattr(gdalconst, _warp_algo))
```
dcdenu4, line 94: Very cool!

dcdenu4 (approved):
Thanks @phargogh! That's a really neat solution for the resampling methods. Thanks for looking into the weeds to pull that out.

It looks like there is a conflict with `History.rst` but I'll go ahead and approve now.

phargogh:
Thanks @dcdenu4 !  I made two additional changes beyond the merge conflict in order to fix a comment that was made out of date by recent changes and also to include the warp algorithms in the text of exceptions.  So I figure that might deserve a glance over/re-review.

dcdenu4 (approved):
Awesome, thanks @phargogh !

PR #298, opened 2023-01-12
Add `isnan()` checking to routing's`_is_close()`

### LINKED ISSUE #248: Pitfilling is buggy with a nan nodata value
We had a [recent issue on the forums](https://community.naturalcapitalproject.org/t/swy-error-warn-method-invest-3-10-2/2534/12) where the user had a DEM with a `nan` nodata value causing pitfilling to produce a streaky DEM with not-at-all-reasonable DEM values (colorized layer is the pitfilled DEM, grayscale is the source DEM):

Screen Shot 2022-03-16 at 2 06 41 PM

When I converted the DEM's nodata value from `nan` to `-3.4028235e+38`, pitfilling worked as expected.

See the forums for the corrected, functional DEM.  To reproduce, I think we should be able to use any Float32 DEM with a `nan` nodata value.

phargogh:
This PR adds support for IEEE754 `nan` values to our routing `_is_close()` function, which is exclusively used in `pygeoprocessing.routing` to check nodata values.

The addition of `isnan` should not incur a significant cost as [the glibc implementation I found online](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/ieee754/dbl-64/s_isnan.c;h=1f1535dea33e99520e13717ab57f5c9a7f6509c1;hb=HEAD) has no branching at all and is mostly bitshifting -- cheap for hardware.  Even then, only 1 value will need to be checked for `nan`-ness before it falls back to the usual arithmetic of `_is_close()`.  The benefit for this small cost, that we shouldn't have to address this again on the forums(!), at least for pitfilling, should be great.

Fixes #248

### DIFF
--- a/src/pygeoprocessing/routing/routing.pyx
+++ b/src/pygeoprocessing/routing/routing.pyx
@@ -31,6 +31,7 @@ from cython.operator cimport dereference as deref
 from cython.operator cimport preincrement as inc
 from libc.time cimport time as ctime
 from libc.time cimport time_t
+from libc.math cimport isnan
 from libcpp.deque cimport deque
 from libcpp.list cimport list as clist
 from libcpp.pair cimport pair
@@ -182,6 +183,8 @@ cdef cppclass GreaterPixel nogil:
         return 0
 
 cdef int _is_close(double x, double y, double abs_delta, double rel_delta):
+    if isnan(x) and isnan(y):
+        return 1
     return abs(x-y) <= (abs_delta+rel_delta*abs(y))
 
 # a class to allow fast random per-pixel access to a raster for both setting
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,11 @@ Release History
 
 Unreleased Changes
 ------------------
+* Nodata value checking in ``pygeoprocessing.routing`` now correctly handles
+  comparison of ``nan`` values.  This is explicitly tested in
+  ``pygeoprocessing.routing.fill_pits``, but should also improve the
+  experience of other routing functions as well.
+  https://github.com/natcap/pygeoprocessing/issues/248
 * Added a function to build overviews for a raster,
   ``pygeoprocessing.build_overviews``. Related to this,
   ``pygeoprocessing.get_raster_info()`` now includes an ``'overviews'`` key
--- a/tests/test_routing.py
+++ b/tests/test_routing.py
@@ -149,6 +149,26 @@ class TestRouting(unittest.TestCase):
         dem_array[3:8, 3:8] = 0.0
         numpy.testing.assert_almost_equal(result_array, dem_array)
 
+    def test_pit_filling_nodata_nan(self):
+        """PGP.routing: test pitfilling with nan nodata value."""
+        base_path = os.path.join(self.workspace_dir, 'base.tif')
+        dem_array = numpy.zeros((11, 11), dtype=numpy.float32)
+        nodata = numpy.nan
+        dem_array[3:8, 3:8] = -1
+        dem_array[0, 0] = -1
+        dem_array[1, 1] = nodata
+        _array_to_raster(dem_array, nodata, base_path)
+
+        fill_path = os.path.join(self.workspace_dir, 'filled.tif')
+        pygeoprocessing.routing.fill_pits(
+            (base_path, 1), fill_path, working_dir=self.workspace_dir)
+
+        result_array = pygeoprocessing.raster_to_numpy_array(fill_path)
+        self.assertEqual(result_array.dtype, numpy.float32)
+        # the expected result is that the pit is filled in
+        dem_array[3:8, 3:8] = 0.0
+        numpy.testing.assert_almost_equal(result_array, dem_array)
+
     def test_flow_dir_d8(self):
         """PGP.routing: test D8 flow."""
         dem_path = os.path.join(self.workspace_dir, 'dem.tif')


src/pygeoprocessing/routing/routing.pyx
```diff
@@ -182,6 +183,8 @@ cdef cppclass GreaterPixel nogil:
         return 0
 
 cdef int _is_close(double x, double y, double abs_delta, double rel_delta):
+    if isnan(x) and isnan(y):
```
dcdenu4, line 186: Is it possible that `x` or `y` here is `nan` but not both? If so, does it still make sense to continue?

dcdenu4 (commented):
Thanks @phargogh, just one quick question.

src/pygeoprocessing/routing/routing.pyx
phargogh: Yes, it is absolutely possible that only one of these values may be `nan`!  Fortunately, the math of `abs(x-y) <= (abs_delta+rel_delta*abs(y))` works as expected (I'm using python here, but C and C++ both have equivalent results).  The conclusion here is that comparing a float value to `nan` returns False.

```python
def isclose(x, y, abs_delta, rel_delta):
     return abs(x-y) <= (abs_delta+rel_delta*abs(y))
```

```python
>>> isclose(1.5, float('nan'), 1e-8, 1e-5)
False
```

The problem this PR is trying to resolve is that comparing two `nan` values also is `False`:
```python
>>> isclose(float('nan'), float('nan'), 1e-8, 1e-5)
False
```
So this PR handles the case by returning `True` when both `x` and `y` are `nan`.

phargogh: Your comment also made me wonder if we might have a speedup if we did handle the case where only one of the values were `nan`, since we can avoid the arithmetic entirely if we do know that.

_Conclusion: I think what's in the PR is the best solution because it's easier to understand and the timings are so close anyways that I'm having a tough time getting any definitive timings._

I tested a couple different combinations, including the original (`base`) and found that on my laptop with 10 billion iterations, handling the exclusive-or case in a way that made sense took _way_ more time than I thought it would, doubling the function's apparent runtime.

It's worth noting that any improvements in times are only in the hundredsths or thousandths of a second faster over 10 billion runs, so averaged over 10 billion runs we're talking about differences of trillionths of a second per function call.  This is a level of timing precision that is unreliable, so I'm getting conflicting results over repeated runs.

So, that being said, the usually fastest version of `isclose` that I've found on my laptop a tossup between what's in the PR and a somewhat more convoluted implementation that sometimes reduces the number of times `isnan` is called:
```pyrex
cdef int _is_close(double x, double y, double abs_delta, double rel_delta):
    if isnan(x) and isnan(y):
        return 1
    return abs(x-y) <= (abs_delta+rel_delta*abs(y))

cdef int _is_close_xor_switch(double x, double y, double abs_delta, double rel_delta):
    if isnan(x):
        return isnan(y)
    return abs(x-y) <= (abs_delta+rel_delta*abs(y))
```

Here are one example of timing results:
```
2.0 elapsed time (base):  4.7931s 1
2.0 elapsed time (pr):    4.693s 1
2.0 elapsed time (xor1):  8.8043s 1
2.0 elapsed time (xorf):  8.7695s 1
2.0 elapsed time (xors):  4.6966s 1
2.0 elapsed time (xor):   8.7564s 1
nan elapsed time (base):  4.6996s 0
nan elapsed time (pr):    4.6963s 0
nan elapsed time (xor1):  7.9259s 0
nan elapsed time (xorf):  7.9238s 0
nan elapsed time (xors):  4.6906s 0
nan elapsed time (xor):   7.9262s 0
```

Here's the source code:
```pyrex
cdef int _is_close_base(double x, double y, double abs_delta, double rel_delta):
    return abs(x-y) <= (abs_delta+rel_delta*abs(y))

cdef int _is_close(double x, double y, double abs_delta, double rel_delta):
    if isnan(x) and isnan(y):
        return 1
    return abs(x-y) <= (abs_delta+rel_delta*abs(y))

cdef int _is_close_xor(double x, double y, double abs_delta, double rel_delta):
    if isnan(x) and isnan(y):
        return 1
    if isnan(x) ^ isnan(y):
        return 0
    return abs(x-y) <= (abs_delta+rel_delta*abs(y))

cdef int _is_close_xor_once(double x, double y, double abs_delta, double rel_delta):
    cdef int x_is_nan = isnan(x)
    cdef int y_is_nan = isnan(y)
    if x_is_nan and y_is_nan:
        return 1
    if x_is_nan ^ y_is_nan:
        return 0
    return abs(x-y) <= (abs_delta+rel_delta*abs(y))

cdef int _is_close_xor_first(double x, double y, double abs_delta, double rel_delta):
    cdef int x_is_nan = isnan(x)
    cdef int y_is_nan = isnan(y)
    if x_is_nan ^ y_is_nan:
        return 0
    if x_is_nan & y_is_nan:
        return 1
    return abs(x-y) <= (abs_delta+rel_delta*abs(y))

cdef int _is_close_xor_switch(double x, double y, double abs_delta, double rel_delta):
    if isnan(x):
        return isnan(y)
    else:
        return abs(x-y) <= (abs_delta+rel_delta*abs(y))

# re-run this once I have an internet connection.
def test_isclose():
    cdef unsigned long n = 10000000000  # 10 billion iterations
    cdef unsigned long i
    cdef unsigned long sum_ = 0
    cdef double y_value
    for y_value in [2, float('nan')]:
        start_time = time.time()
        sum_ = 0
        for i in range(n):
            sum_ += _is_close_base(i, y_value, 1e-8, 1e-5)
        print(f'{y_value} elapsed time (base):  {round(time.time() - start_time, 4)}s {sum_}')

        start_time = time.time()
        sum_ = 0
        for i in range(n):
            sum_ += _is_close(i, y_value, 1e-8, 1e-5)
        print(f'{y_value} elapsed time (pr):    {round(time.time() - start_time, 4)}s {sum_}')

        start_time = time.time()
        sum_ = 0
        for i in range(n):
            sum_ += _is_close_xor_once(i, y_value, 1e-8, 1e-5)
        print(f'{y_value} elapsed time (xor1):  {round(time.time() - start_time, 4)}s {sum_}')

        start_time = time.time()
        sum_ = 0
        for i in range(n):
            sum_ += _is_close_xor_first(i, y_value, 1e-8, 1e-5)
        print(f'{y_value} elapsed time (xorf):  {round(time.time() - start_time, 4)}s {sum_}')

        start_time = time.time()
        sum_ = 0
        for i in range(n):
            sum_ += _is_close_xor_switch(i, y_value, 1e-8, 1e-5)
        print(f'{y_value} elapsed time (xors):  {round(time.time() - start_time, 4)}s {sum_}')

        start_time = time.time()
        sum_ = 0
        for i in range(n):
            sum_ += _is_close_xor(i, y_value, 1e-8, 1e-5)
        print(f'{y_value} elapsed time (xor):   {round(time.time() - start_time, 4)}s {sum_}')

```

dcdenu4 (approved):
Wow, thanks @phargogh for that run down! That is fun to see and think about. Sticking with how it currently stands in this PR makes sense to me too.

PR #302, opened 2023-01-31
Allow user-defined target layer name in `reproject_vector`

### LINKED ISSUE #301: Allow the target layer name to be defined by user input in `reproject_vector`
`pygeoprocessing.reproject_vector` currently takes the layer name from the source vector's layer definition.  It would be useful to the Urban Nature Access model if we could define the name of the layer, such as if we wanted to ad a suffix or change it entirely.

phargogh:
This PR allows the user to define a target layer name when reprojecting a vector.  A special case has been implemented to handle the case of an ESRI Shapefile, where the layer name matches the filename and cannot be defined by hand.

Fixes #301

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,11 @@ Release History
 
 Unreleased Changes
 ------------------
+* The function ``pygeoprocessing.reproject_vector`` now accepts an optional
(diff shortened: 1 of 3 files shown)


davemfish (approved):
Very nice @phargogh !

PR #303, opened 2023-02-08
Add a library of convolution kernels

### LINKED ISSUE #268: Proposal: library of convolution kernels
`pygeoprocessing.convolve_2d` has some pretty stringent requirements for how a kernel must be formatted (must be a spatial raster, best if it's tiled, spatial reference needs to be set, but it doesn't matter what it is), and so  just formatting your kernel correctly is a pretty high bar before `convolve_2d` can even be used.

I propose:

* Creating a library of convolution kernels, `pygeoprocessing.kernels` containing kernel-related utilities
* Moving kernel creation functions used in InVEST that follow standard kernel models (e.g. gaussian, linear, etc.) over to `pygeoprocessing.kernels`
* Creating some helper functions (as needed) in this subpackage to help standardize the creation of kernels.

phargogh:
This PR implements a library of convolution kernels, taken from InVEST where 2 or more models use the kernel.  Some kernels, such as the power and inverted parabolic functions in Urban Nature Access, are not included here as a result.

The major changes here are:

* `pygeoprocessing.geoprocessing.numpy_array_to_raster` has been modified to make the pixel size and origin optional, as this spatial information is not needed for kernels.
* `pygeoprocessing.kernels.kernel_from_numpy_array` is simply a wrapper around `numpy_array_to_raster`, pre-filling in the parameters that kernels (and `convolve_2d`) don't care about.
* `pygeoprocessing.kernels.create_kernel` creates a distance-based kernel from a user-defined function.  The function may be a callable, or an `eval`-able string.
* A selection of common kernel functions are also provided.

I elected to create separate top-level functions for the various decay kernels in order to make it easier to inspect the required parameters and so that the user would only need to make a single function call.  Other approaches, like always defining functions as strings or always defining functions as functions of numpy arrays, require additional work to be able to use (string formatting and something like `functools.partial`, respectively).

Please let me know what you think, including about the names of the functions!

Fixes #268

### DIFF
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -3936,7 +3936,7 @@ def numpy_array_to_raster(
         new_raster.SetProjection(projection_wkt)
     if origin is not None and pixel_size is not None:
         new_raster.SetGeoTransform(
-            [origin[0], pixel_size[0], 0.0, origin[1], 0.0, pixel_size[1]])
+            [origin[0], pixel_size[0], 0, origin[1], 0, pixel_size[1]])
     elif origin is not None or pixel_size is not None:
         raise ValueError(
             "Origin and pixel size must both be defined or both be None")
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -2797,7 +2797,9 @@ def convolve_2d(
         kernel_path_band (tuple): a 2 tuple of the form
             (filepath to kernel raster, band index), all pixel values should
             be valid -- output is not well defined if the kernel raster has
-            nodata values.
+            nodata values.  To create a kernel raster, see the documentation
+            and helper functions available in the
+            :doc:`pygeoprocessing kernels module <pygeoprocessing.kernels>`.
         target_path (string): filepath to target raster that's the convolution
             of signal with kernel.  Output will be a single band raster of
             same size and projection as ``signal_path_band``. Any nodata pixels
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -741,7 +741,7 @@ def align_and_resize_raster_stack(
             ``geoprocessing.DEFAULT_OSR_AXIS_MAPPING_STRATEGY``. This parameter
             should not be changed unless you know what you are doing.
 
-    Return:
+    Returns:
         None
 
     Raises:
@@ -765,7 +765,6 @@ def align_and_resize_raster_stack(
             file.
         ValueError
             If ``pixel_size`` is not a 2 element sequence of numbers.
-
     """
     # make sure that the input lists are of the same length
     list_lengths = [
@@ -2854,7 +2853,7 @@ def convolve_2d(
             tuple/list as the second. Defaults to a GTiff driver tuple
             defined at geoprocessing.DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS.
 
-    Return:
+    Returns:
         ``None``
 
     Raises:
@@ -2865,7 +2864,6 @@ def convolve_2d(
             if ``signal_path_band`` or ``kernel_path_band`` is a row based
             blocksize which would result in slow runtimes due to gdal
             cache thrashing.
-
     """
     if target_datatype is not gdal.GDT_Float64 and target_nodata is None:
         raise ValueError(
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -3900,9 +3900,12 @@ def numpy_array_to_raster(
     Args:
         base_array (numpy.array): a 2d numpy array.
         target_nodata (numeric): nodata value of target array, can be None.
-        pixel_size (tuple): square dimensions (in ``(x, y)``) of pixel.
-        origin (tuple/list): x/y coordinate of the raster origin.
-        projection_wkt (str): target projection in wkt.
+        pixel_size (tuple): square dimensions (in ``(x, y)``) of pixel. Can be
+            None to indicate no stated pixel size.
+        origin (tuple/list): x/y coordinate of the raster origin. Can be None
+            to indicate no stated origin.
+        projection_wkt (str): target projection in wkt.  Can be None to
+            indicate no projection/SRS.
         target_path (str): path to raster to create that will be of the
             same type of base_array with contents of base_array.
         raster_driver_creation_tuple (tuple): a tuple containing a GDAL driver
@@ -3933,8 +3936,12 @@ def numpy_array_to_raster(
         options=raster_driver_creation_tuple[1])
     if projection_wkt is not None:
         new_raster.SetProjection(projection_wkt)
-    new_raster.SetGeoTransform(
-        [origin[0], pixel_size[0], 0.0, origin[1], 0.0, pixel_size[1]])
+    if origin is not None and pixel_size is not None:
+        new_raster.SetGeoTransform(
+            [origin[0], pixel_size[0], 0.0, origin[1], 0.0, pixel_size[1]])
+    elif origin is not None or pixel_size is not None:
+        raise ValueError(
+            "Origin and pixel size must both be defined or both be None")
     new_band = new_raster.GetRasterBand(1)
     if target_nodata is not None:
         new_band.SetNoDataValue(target_nodata)
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -3897,6 +3897,14 @@ def numpy_array_to_raster(
         raster_driver_creation_tuple=DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS):
     """Create a single band raster of size ``base_array.shape``.
 
+    The GDAL datatype of the target raster is determined by the numpy dtype of
+    ``base_array``.
+
+    Note:
+        The ``origin`` and ``pixel_size`` parameters must both be defined
+        properly as 2-tuples of floats, or else must both be set to ``None``.
+        A ``ValueError`` will be raised otherwise.
+
     Args:
         base_array (numpy.array): a 2d numpy array.
         target_nodata (numeric): nodata value of target array, can be None.
--- a/src/pygeoprocessing/kernels.py
+++ b/src/pygeoprocessing/kernels.py
@@ -61,14 +61,11 @@ def dichotomous_kernel(target_kernel_path, max_distance, normalize):
 # UNA calls this a density kernel
 # really, this is quite specific to UNA
 def parabolic_decay_kernel(target_kernel_path, max_distance, normalize=True):
-    """Create an inverted parabola that reaches a value of 0 at ``max_distance``
-
+    """Create an inverted parabola that reaches 0 at ``max_distance``
     """
-    def _density(dist):
-        return (0.75 * (1 - (dist / max_distance) ** 2))
-
     create_kernel(
-        target_kernel_path, _density, max_distance, normalize=normalize)
+        target_kernel_path, "0.75 * (1-(dist / max_dist) ** 2)",
+        max_distance, normalize=normalize)
 
 
 def numexpr_kernel(target_kernel_path, max_distance, expression, extras=None):
@@ -82,20 +79,16 @@ def numexpr_kernel(target_kernel_path, max_distance, expression, extras=None):
 
 
 def exponential_decay_kernel(target_kernel_path, max_distance,
-                             distance_factor=1, normalize=True):
-    def _exp_decay(dist):
-        return numpy.exp(-(dist*distance_factor) / max_distance)
-
+                             normalize=True):
     create_kernel(
-        target_kernel_path, _exp_decay, max_distance, normalize=normalize)
+        target_kernel_path, "exp(-dist / max_dist)",
+        max_distance, normalize=normalize)
 
 
 def linear_decay_kernel(target_kernel_path, max_distance, normalize=True):
-    def _linear_decay(dist):
-        return (max_distance - dist) / max_distance
-
     create_kernel(
-        target_kernel_path, _linear_decay, max_distance, normalize=normalize)
+        target_kernel_path, "(max_dist - dist) / max_dist",
+        max_distance, normalize=normalize)
 
 
 def gaussian_decay_kernel(target_kernel_path, sigma, n_std_dev=3,
--- a/src/pygeoprocessing/kernels.py
+++ b/src/pygeoprocessing/kernels.py
@@ -65,10 +65,10 @@ def parabolic_decay_kernel(target_kernel_path, max_distance, **kwargs):
         max_distance, **kwargs)
 
 
-def exponential_decay_kernel(target_kernel_path, max_distance,
-                             **kwargs):
+def exponential_decay_kernel(target_kernel_path, expected_distance,
+                             max_distance, **kwargs):
     create_kernel(
-        target_kernel_path, "exp(-dist / max_dist)",
+        target_kernel_path, f"exp(-dist / {expected_distance})",
         max_distance, **kwargs)
 
 
--- a/src/pygeoprocessing/kernels.py
+++ b/src/pygeoprocessing/kernels.py
@@ -1,0 +1,99 @@
+import logging
+import math
+
+import numpy
+import pygeoprocessing
+from osgeo import gdal
+from osgeo import osr
+
+FLOAT32_NODATA = numpy.finfo(numpy.float32).min
+LOGGER = logging.getLogger(__name__)
+
+
+def _create_distance_based_kernel(
+        target_kernel_path, function, max_distance, normalize):
+    """
+    Create a kernel raster based on pixel distance from the centerpoint.
+
+    Args:
+        target_kernel_path (string): The path to where the target kernel should
+            be written on disk.  If this file does not have the suffix
+            ``.tif``, it will be added to the filepath.
+        function (callable): A python callable that takes as input a
+            2D numpy array and returns a 2D numpy array.  The input array will
+            contain float32 distances to the centerpoint pixel of the kernel.
+        max_distance (float): The maximum distance of kernel values from
+            the center point.  Values outside of this distance will be set to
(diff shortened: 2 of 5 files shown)


src/pygeoprocessing/geoprocessing.py
```diff
@@ -3933,8 +3934,12 @@ def numpy_array_to_raster(
         options=raster_driver_creation_tuple[1])
     if projection_wkt is not None:
         new_raster.SetProjection(projection_wkt)
-    new_raster.SetGeoTransform(
-        [origin[0], pixel_size[0], 0.0, origin[1], 0.0, pixel_size[1]])
+    if origin is not None and pixel_size is not None:
+        new_raster.SetGeoTransform(
+            [origin[0], pixel_size[0], 0.0, origin[1], 0.0, pixel_size[1]])
```
emlys:
```suggestion
            [origin[0], pixel_size[0], 0, origin[1], 0, pixel_size[1]])
```

src/pygeoprocessing/geoprocessing.py
```diff
@@ -3933,8 +3934,12 @@ def numpy_array_to_raster(
...
+    elif origin is not None or pixel_size is not None:
+        raise ValueError(
+            "Origin and pixel size must both be defined or both be None")
```
emlys, line 3960: Is this a GDAL requirement?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -3900,9 +3898,12 @@ def numpy_array_to_raster(
...
         target_nodata (numeric): nodata value of target array, can be None.
-        pixel_size (tuple): square dimensions (in ``(x, y)``) of pixel.
-        origin (tuple/list): x/y coordinate of the raster origin.
-        projection_wkt (str): target projection in wkt.
+        pixel_size (tuple): square dimensions (in ``(x, y)``) of pixel. Can be
+            None to indicate no stated pixel size.
+        origin (tuple/list): x/y coordinate of the raster origin. Can be None
+            to indicate no stated origin.
+        projection_wkt (str): target projection in wkt.  Can be None to
+            indicate no projection/SRS.
```
emlys, line 3924: I like the idea of making these optional! does the GDAL data model and all supported raster formats allow the pixel size, origin, and projection to be undefined?

src/pygeoprocessing/kernels.py
```diff
@@ -0,0 +1,295 @@
...
+            be written on disk.  If this file does not have the suffix
+            ``.tif``, it will be added to the filepath.
+        function (callable): A python callable that takes as input a
+            1D numpy array and returns a 1D numpy array.  The input array will
+            contain float32 distances to the centerpoint pixel of the kernel.
+        max_distance (float): The maximum distance of kernel values from
+            the center point.  Values outside of this distance will be set to
+            ``0.0``.
+        pixel_radius=None (float): The radius (in pixels) of the target kernel.
+            If ``None``, the radius will be ``math.ceil(max_distance)``.
```
emlys: Some thoughts on naming... 

I found `pixel_radius` kind of confusing because radius makes me think circles, but this is about the size of a square. Apparently "radius" can also apply to polygons, but [it's the distance from center to vertex](https://www.mathopenref.com/polygonradius.html), which this is not. Seems like the technical term for this measurement is ["apothem"](https://en.wikipedia.org/wiki/Apothem) but that's maybe too obscure. And because we add a center pixel, this measurement gets +0.5 pixel in the output.

src/pygeoprocessing/kernels.py
```diff
@@ -0,0 +1,295 @@
...
+    # The kernel is technically a TIFF, not a GeoTiff.
+    pygeoprocessing.numpy_array_to_raster(
+        numpy_array, target_nodata=None, pixel_size=None, origin=None,
+        projection_wkt=None, target_path=target_kernel_path)
+
+
+def dichotomous_kernel(
+        target_kernel_path: Text,
+        max_distance: Union[int, float],
+        pixel_radius: Union[int, float] = None,
```
emlys: Is it necessary to have both `max_distance` and `pixel_radius` here? i.e., Is there a case where the kernel needs to be padded with a border of extra 0s?

src/pygeoprocessing/kernels.py
```diff
@@ -0,0 +1,295 @@
...
+    create_kernel(
+        target_kernel_path=target_kernel_path,
+        function=_normal_decay,
+        max_distance=(sigma * n_std_dev),
+        pixel_radius=pixel_radius,
+        normalize=normalize
+    )
+
+
+def create_kernel(
```
emlys: Suggest naming this `create_distance_kernel` or `create_distance_decay_kernel`, since kernel is a generic term and not necessarily distance based

src/pygeoprocessing/kernels.py
```diff
@@ -0,0 +1,295 @@
...
+        target_kernel_path: str,
+        function: Union[str, Callable],
```
emlys: similarly, maybe a more specific name here like `distance_decay_function`

src/pygeoprocessing/kernels.py
```diff
@@ -0,0 +1,295 @@
...
+def exponential_decay_kernel(
+        target_kernel_path: Text,
+        max_distance: Union[int, float],
+        expected_distance: Union[int, float],
+        pixel_radius: Union[int, float] = None,
+        normalize: bool = True) -> None:
+    """Create an exponential decay kernel.
+
+    This kernel will reach 1/e at ``expected_distance``, but will continue to
+    have nonzero values out to ``max_distance``.
```
emlys: It could be helpful to include the actual decay function in the docstring

src/pygeoprocessing/kernels.py
```diff
@@ -0,0 +1,295 @@
...
+        normalize=normalize
+    )
+
+
+def linear_decay_kernel(
+        target_kernel_path: Text,
+        max_distance: Union[int, float],
+        pixel_radius: Union[int, float] = None,
+        normalize: bool = True) -> None:
+    """Create a linear decay kernel.
```
emlys, line 145: same as above, include decay function in docstring

src/pygeoprocessing/kernels.py
```diff
@@ -0,0 +1,295 @@
...
+    )
+
+
+def normal_distribution_kernel(
+        target_kernel_path: Text,
+        sigma: Union[int, float],
+        n_std_dev: Union[int, float] = 3,
+        pixel_radius: Union[int, float] = None,
+        normalize: bool = True):
+    """Create an decay kernel following a normal distribution.
```
emlys, line 179: same as above, include decay function in docstring

src/pygeoprocessing/kernels.py
```diff
@@ -0,0 +1,295 @@
...
+
+    Returns:
+        ``None``
+    """
+    def _linear_decay(dist):
+        return (max_distance - dist) / max_distance
+
+    create_kernel(
+        target_kernel_path=target_kernel_path,
+        function="(max_dist - dist) / max_dist",
```
emlys: this is also defined in the function above, which is unused

src/pygeoprocessing/kernels.py
```diff
@@ -0,0 +1,295 @@
...
+    """Create an decay kernel following a normal distribution.
+
+    Args:
+        target_kernel_path: The path to where the kernel will be written.
+            Must have a file extension of ``.tif``.
+        sigma: The width (in pixels) of a standard deviation.
+        n_std_dev: The number of standard deviations to include in the kernel.
+            The kernel will have values of 0 when at a distance of
+            ``(sigma * n_std_dev)`` away from the centerpoint.
+        pixel_radius: The radius of the target kernel, in pixels.  If ``None``,
```
emlys: like with dichotomous decay, when would pixel_radius need to be different from sigma * n_std_dev ?

emlys (changes requested):
@phargogh this is really cool! The type hints are neat. I had some questions and I'm not sure I totally understand the purpose and difference of `max_distance` vs `pixel_radius`.

src/pygeoprocessing/geoprocessing.py
phargogh: Sort of - GDAL requires that the geotransform is a 6-element array of `double`s, so we would need to assume some reasonable origin or pixel size.  I suppose we could theoretically do `nan` or something like that, but it seems easier to just require that the geotransform be set properly or else not at all.

src/pygeoprocessing/geoprocessing.py
phargogh: Much to my surprise, yes!  I do not think this was always the case.  At least with GeoTiffs, you can just not provide these parameters and then you have a TIFF, which works the same way.  It's great for kernels and other things where we either don't have or don't need the spatial information.

phargogh: I also asked [this question on the GIS StackExchange](https://gis.stackexchange.com/questions/451261/is-there-a-standard-spatial-reference-to-indicate-that-a-raster-is-aspatial) to see if there exists a standard way of expressing that a raster is not a spatial dataset, and it just seems like there isn't a super clear answer.  Folks in the comments said "make it a TIFF" or "use a GeoPackage".

src/pygeoprocessing/kernels.py
phargogh: I agree, `apothem` is much more accurate (even if not commonly recognized), particularly since each pixel more accurately represents an individual point.  With that model, we don't even need the +0.5 pixel adjustment!

src/pygeoprocessing/kernels.py
phargogh: Interestingly, I found the opposite to be true when working these into InVEST!  Stormwater's dichotomous kernel is built in such a way that it was helpful to have `pixel_radius = math.floor(max_distance)`.

```python
    raster_info = pygeoprocessing.get_raster_info(raster_path)
    pixel_radius = radius / abs(raster_info['pixel_size'][0])
    kernels.dichotomous_kernel(
        kernel_path,
        max_distance=pixel_radius,
        pixel_radius=math.floor(pixel_radius),
        normalize=False)
```

So I think the more generalized argument to be made is when you want have a function where you want `max_distance` to be beyond the edge of the bounds of the kernel.

src/pygeoprocessing/kernels.py
phargogh: Good point!  Renamed to `create_distance_decay_kernel()` in b6fa7a6

src/pygeoprocessing/kernels.py
phargogh: Renamed to `distance_decay_function` in 7bc24b4

src/pygeoprocessing/kernels.py
phargogh: I've renamed `pixel_radius` to `apothem` in 4e99b96

src/pygeoprocessing/kernels.py
phargogh: I went back and forth on this about how best to make it happen, but I fully agree.  Patched in b558503.

src/pygeoprocessing/kernels.py
phargogh: I went back and forth on this about how best to make it happen, but I fully agree.  Patched in b558503.

src/pygeoprocessing/kernels.py
phargogh: I went back and forth on this about how best to make it happen, but I fully agree.  Patched in b558503.

src/pygeoprocessing/kernels.py
phargogh: Yep!  Patched in b558503be8987633f0de60aba8b9eb169ffd4292

src/pygeoprocessing/kernels.py
phargogh: Yeah, for a normal distribution, I think there's a very good case to make for removing the pixel radius.  And, unlike the dichotomous kernel, we don't have a need for it in InVEST.  Removed in 3efbd92.

phargogh:
Thanks for your thoughtful comments @emlys !  I've responded to all of your comments and also improved the module-level documentation.

src/pygeoprocessing/geoprocessing.py
emlys: makes sense!

emlys: Would it be worth noting this requirement in the docstring?

phargogh: Definitely.  Added in 8fb9829

src/pygeoprocessing/geoprocessing.py
emlys: Interesting! I tried this out and QGIS still identifies it as a geotiff even if the pixel size, origin, and projection are not provided.

emlys: Though, it does the same with an arbitrary non-spatial .tiff image.

phargogh: Yes, it's the same driver as far as GDAL is concerned.  I guess it's because GeoTiff is just an extension of the TIFF spec, that a Tiff can be opened as a GeoTiff?

src/pygeoprocessing/kernels.py
```diff
@@ -0,0 +1,345 @@
...
+    """Create a binary kernel indicating presence/absence within a distance.
+
+    This is equivalent to ``int(dist <= max_distance)`` for each pixel in the
+    kernel, where ``dist`` is the euclidean distance of the pixel's centerpoint
+    to the centerpoint of the center pixel.
+
+    Args:
+        target_kernel_path: The path to where the kernel will be written.
+            Must have a file extension of ``.tif``.
+        max_distance: The maximum distance of the kernel, in pixels. Kernel
```
emlys: Sorry, this wording is still tripping me up... what about something like
```suggestion
        max_distance: The distance threshold, in pixels. Kernel
```

src/pygeoprocessing/kernels.py
```diff
@@ -0,0 +1,348 @@
...
+    the kernel.
+
+    Args:
+        target_kernel_path: The path to where the kernel will be written.
+            Must have a file extension of ``.tif``.
+        max_distance: The maximum distance of the kernel, in pixels. Kernel
+            pixels that are greater than ``max_distance`` from the centerpoint
+            of the kernel will have values of ``0.0``.
+        apothem: The apothem of the target kernel, in pixels.  If ``None``,
+            then ``math.ceil(max_distance)`` will be used.
```
emlys:
```suggestion
            then ``math.floor(max_distance)`` will be used.
```

src/pygeoprocessing/kernels.py
```diff
@@ -0,0 +1,348 @@
...
+    Args:
+        target_kernel_path: The path to where the kernel will be written.
+            Must have a file extension of ``.tif``.
+        max_distance: The maximum distance of the kernel, in pixels. Kernel
+            pixels that are greater than ``max_distance`` from the centerpoint
+            of the kernel will have values of ``0.0``.
+        expected_distance: The distance, in pixels, from the centerpoint at
+            which decayed values will equal ``1/e``.
+        apothem: The apothem of the target kernel, in pixels.  If ``None``,
+            then ``math.ceil(max_distance)`` will be used.
```
emlys:
```suggestion
            then ``math.floor(max_distance)`` will be used.
```

src/pygeoprocessing/kernels.py
```diff
@@ -0,0 +1,348 @@
...
+    to the centerpoint of the center pixel.
+
+    Args:
+        target_kernel_path: The path to where the kernel will be written.
+            Must have a file extension of ``.tif``.
+        max_distance: The distance threshold, in pixels. Kernel
+            pixels that are greater than ``max_distance`` from the centerpoint
+            of the kernel will have values of ``0.0``.
+        apothem: The apothem of the target kernel, in pixels.  If ``None``,
+            then ``math.ceil(max_distance)`` will be used.
```
emlys:
```suggestion
            then ``math.floor(max_distance)`` will be used.
```

emlys (commented):
Thanks @phargogh! I have two last questions:
1. Now that `apothem` defaults to `math.floor(max_distance)`, (which solves the special case with stormwater), is there still a use case for `apothem` at all?
2. What is the use case for passing in `distance_decay_function` as a string to evaluate, rather than a callable?

phargogh:
> Now that apothem defaults to math.floor(max_distance), (which solves the special case with stormwater), is there still a use case for apothem at all?

Nope, you're absolutely right!  I've removed the apothem stuff in fc3e186.

> What is the use case for passing in distance_decay_function as a string to evaluate, rather than a callable?

Simply put, convenience.  There exist kernels that are just easier to express as a string.  For example, Urban Nature Access will eventually have a power function where the kernel is defined by $d_{ij}^{(-\beta)}$ (see https://github.com/natcap/invest/issues/1175).  Each kernel would be expected to have a different $\beta$ value.  As a string, this is super easy to express:

```python
# beta is read in from a table
kernel_func = f"dist ** (-{beta})"
```

As a callable, the clearest way to express this is with a higher-order function:

```python
def _create_power_function(beta):
    """Descriptive docstring."""
    def _actual_power_function_used_by_kernel_creation_func(distance):
        """Another descriptive docstring."""
        return distance ** (-beta)

kernel_func = _create_power_function(beta)
```

Or else a function that's redefined for each `beta` in a loop (though this would only work when synchronously called):
```python
for beta in my_beta_params_from_table:
    def _power_function(distance):
        """Descriptive docstring."""
        return distance ** (-beta)
    kernel_func = _power_function
```

Alternatively, this could be abbreviated with `functools.partial` like so
```python
import functools

def _power_function(distance, beta):
    """Descriptive docstring."""
    return distance ** (-beta)

kernel_func = functools.partial(_power_function, beta=beta)
```

Really, there's only a subset of kernel functions that would be fit to be represented as a string expression.  Functions that are either very complex, such as the normal distribution kernel defined in this module, or require special indexing (they approach infinity or have a division-by-zero under certain conditions) are clearer to express as a function than as an expression.

Anyways, let me know what you think!  I'm not necessarily set on the idea of supporting a string expression, but I do like the idea of supporting things that make it easier for us to express things that feel like they should be easy to express.

emlys (approved):
Looks good!

PR #304, opened 2023-02-09
Replace `<lambda>` in geoprocessing progress logging with the function name

### LINKED ISSUE #300: Can we improve the `<lambda>` in timed progress logging?
Whenever we have a bunch of long-running `raster_calculator()` calls, particularly when we're doing some multiprocessing, it's easy to end up with a whole pile of logging messages that look like this:

```
2023-01-27 16:51:49,702:pid6291:INFO:pygeoprocessing.geoprocessing:<lambda>:35.2% complete
```

Worse, even though many of these come from `raster_calculator`, the timed progress logging could really be coming from almost any one of the geoprocessing functions in the module.  With python's ability to walk the call stack at runtime, we have to be able to programmatically and deliberately set the `__name__` attribute to the calling function.  So instead of seeing `<lambda>`, we should see `raster_calculator` or maybe even a user-defined string if it's appropriate.

phargogh:
This PR uses a new subclass of `logging.LoggerAdapter` to:

1. Handle logging that should only happen on a specified interval and
2. Make sure that the correct parameters are passed in to adjust the function name of the stack frame we want, not the default stack frame.

As far as logging itself goes, this PR transforms this:
```
2023-02-09 12:33:46,486:pid56575:INFO:pygeoprocessing.geoprocessing:raster_calculator:starting stats_worker
2023-02-09 12:33:46,486:pid56575:INFO:pygeoprocessing.geoprocessing:raster_calculator:started stats_worker <Thread(Thread-1, started daemon 12951224320)>
2023-02-09 12:33:51,536:pid56575:INFO:pygeoprocessing.geoprocessing:<lambda>:foo.tif 0.7% complete
```

into this:
```
2023-02-09 12:33:46,486:pid56575:INFO:pygeoprocessing.geoprocessing:raster_calculator:starting stats_worker
2023-02-09 12:33:46,486:pid56575:INFO:pygeoprocessing.geoprocessing:raster_calculator:started stats_worker <Thread(Thread-1, started daemon 12951224320)>
2023-02-09 12:33:51,536:pid56575:INFO:pygeoprocessing.geoprocessing:raster_calculator:foo.tif 0.7% complete
```

Although the simplest fix would have been to just add the stack frame adjustment to each timed logging call, that solution was both harder to test and meant that we had to remember to add the timed progress logging _with_ the appropriate parameters for the stack frame adjustment.  The solution here is easier to test and assumes a reasonable stack frame adjustment if the user hasn't already defined one.

After submitting the PR I also found a change in the `logging` API, where this isn't actually supported on python 3.7.  Rather than trying to sink some time into recreating later functionality, I opted to just skip this feature unless we're on python 3.8+.

Fixes #300

### DIFF
--- a/tests/test_geoprocessing.py
+++ b/tests/test_geoprocessing.py
@@ -7,6 +7,7 @@ import os
 import pathlib
 import queue
 import shutil
+import sys
 import tempfile
 import time
 import types
@@ -1532,9 +1533,15 @@ class TestGeoprocessing(unittest.TestCase):
             self.assertEqual(record.msg, expected_message)
 
             # check that the function name logged is the name of this test
-            # (which is the calling function)
-            self.assertEqual(
-                record.funcName, self.test_timed_logging_adapter.__name__)
+            # (which is the calling function).  This is only possible on python
+            # 3.8 or later.
+            if sys.version_info >= (3, 8):
+                self.assertEqual(
+                    record.funcName, self.test_timed_logging_adapter.__name__)
+
+        # The rest of this test only applies to python 3.8+
+        if sys.version_info < (3, 8):
+            return
 
         # Check the custom stack frame adjustment
         with capture_logging(pygeoprocessing_logger) as captured_logging:
--- a/tests/test_geoprocessing.py
+++ b/tests/test_geoprocessing.py
@@ -1536,6 +1536,26 @@ class TestGeoprocessing(unittest.TestCase):
             self.assertEqual(
                 record.funcName, self.test_timed_logging_adapter.__name__)
 
+        # Check the custom stack frame adjustment
+        with capture_logging(pygeoprocessing_logger) as captured_logging:
+            timed_logger = TimedLoggingAdapter(interval_s=0.1)
+
+            def _sub_stackframe():
+                time.sleep(0.5) # make sure we pass the interval threshold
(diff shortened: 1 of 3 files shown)


tests/test_geoprocessing.py
```diff
@@ -1506,13 +1507,61 @@ def test_interpolate_points(self):
...
+
+            _sub_stackframe()
+
+        self.assertEqual(len(captured_logging), 1)
+        self.assertEqual(captured_logging[0].msg, 'DANGER')
+
+        # check that the function name logged is the name of this test
+        # (which is the parent of the calling function)
+        self.assertEqual(
+            record.funcName, self.test_timed_logging_adapter.__name__)
```
davemfish, line 1566: `record` here is still referring to the messages from the previous block, so just passing by coincidence, I think.

I really like this `_sub_stackframe()` test though! It's exactly the example I was hoping to find when reading about `stacklevel` kwarg in `TimedLoggingAdapter`

davemfish (changes requested):
Hi @phargogh , I like this solution a lot! I found one little bug in the test, I think. Other than that, looks great

tests/test_geoprocessing.py
phargogh: Good catch!  Patched in 0522693

davemfish (approved):

PR #311, opened 2023-03-21
Update `zonal_statistics` to work on multiple raster layers

### LINKED ISSUE #306: `zonal_statistics` bug when inputs have different CRS
`zonal_statistics` calls `calculate_disjoint_polygon_set` with the `bounding_box` of the original base raster, not the clipped/reprojected raster. If the base raster and the aggregate vector have different CRS, this is incorrect and may exclude features from the calculation.

### LINKED ISSUE #310: Optimize `zonal_statistics` for multiple raster layers
I needed to run `zonal_statistics` on each of several rasters, using the same aggregate vector. There's some duplicated work going on there - rasterizing the same vector once for each call. Maybe `zonal_statistics` could work on one or multiple base rasters?

The warp step also takes a lot longer than I'd like.

emlys:
Fixes #310 

This ended up being a pretty big refactor of the first half of zonal_statistics. It also fixes #306, another bug where some of the logged progress numbers didn't add up, and at least one other previously unreported bug (see test comment below).

As implemented, the old API is still valid and you can still provide a single raster band, so we won't need to make any changes to invest.

### DIFF
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -1371,6 +1371,18 @@ def zonal_statistics(
     that are not aggregated use their bounding box to intersect with the
     raster for overlap statistics.
 
+    Process:
+    1. Align and clip rasters. All rasters are warped to match the projection
+       of the aggregate vector. All rasters are clipped/padded to have the
+       same bounding box, which is the union of all their bounding boxes.
+    2. If polygons might overlap, divide them into disjoint sets.
+    3. For each disjoint set:
+           rasterize the polygons - create a raster of fids
+           for each raster band input:
+
+
+
+
     Note:
         There may be some degenerate cases where the bounding box vs. actual
         geometry intersection would be incorrect, but these are so unlikely as
@@ -1451,32 +1463,6 @@ def zonal_statistics(
         raise RuntimeError(
             f'Could not open layer {aggregate_layer_name} on {aggregate_vector_path}')
 
-    # create a new aggregate ID field to map base vector aggregate fields to
-    # local ones that are guaranteed to be integers.
-    local_aggregate_field_name = 'original_fid'
-    rasterize_layer_args = {
-        'options': [
-            'ALL_TOUCHED=FALSE',
-            f'ATTRIBUTE={local_aggregate_field_name}']
-    }
-
-    # clip base raster to aggregating vector intersection
-    temp_working_dir = tempfile.mkdtemp(dir=working_dir)
-    raster_nodatas = []
-    raster_pixel_sizes = []
-    for i, base_raster_path_band in enumerate(base_raster_path_band_list):
-        raster_info = get_raster_info(base_raster_path_band[0])
-        if (raster_info['datatype'] in {gdal.GDT_Float32, gdal.GDT_Float64}
-                and include_value_counts):
-            LOGGER.warning(
-                'Value counts requested on a floating-point raster. This can '
-                'cause excessive memory usage if the raster has continuous '
-                'values.')
-        # -1 here because bands are 1 indexed
-        raster_nodatas.append(raster_info['nodata'][base_raster_path_band[1]-1])
-        raster_pixel_sizes.append(raster_info['pixel_size'])
-        clipped_raster_path = os.path.join(temp_working_dir, f'clipped_raster_{i}.tif')
-
     def _new_aggregate_dict():
         """Build a new aggregate dict for ``collections.defaultdict``.
 
@@ -1496,16 +1482,43 @@ def zonal_statistics(
             aggregate_dict['value_counts'] = collections.Counter()
         return aggregate_dict
 
+    # clip base raster to aggregating vector intersection
+    temp_working_dir = tempfile.mkdtemp(dir=working_dir)
+    base_raster_paths = []
+    base_raster_bands = []
+    base_raster_nodatas = []
+    base_raster_pixel_sizes = []
+    warped_raster_paths = []
+    for i, base_raster_path_band in enumerate(base_raster_path_band_list):
+        base_raster_paths.append(base_raster_path_band[0])
+        base_raster_bands.append(base_raster_path_band[1])
+        raster_info = get_raster_info(base_raster_path_band[0])
+        if (raster_info['datatype'] in {gdal.GDT_Float32, gdal.GDT_Float64}
+                and include_value_counts):
+            LOGGER.warning(
+                'Value counts requested on a floating-point raster. This can '
+                'cause excessive memory usage if the raster has continuous '
+                'values.')
+        # -1 here because bands are 1 indexed
+        base_raster_nodatas.append(raster_info['nodata'][base_raster_path_band[1]-1])
+        base_raster_pixel_sizes.append(raster_info['pixel_size'])
+        warped_raster_paths.append(
+            os.path.join(temp_working_dir, f'warped_raster_{i}.tif'))
+
+    # Align and clip all the rasters
+    # Align to the aggregate vector's projection
+    # Clip to
     try:
         # nearest neighbor should be okay for all types of data
         # because we aren't changing pixel size
         align_and_resize_raster_stack(
-            [base_raster_path_band[0]], [clipped_raster_path], ['near'],
-            raster_info['pixel_size'], 'intersection',
+            base_raster_path_list=[b[0] for b in base_raster_path_bands],
+            target_raster_path_list=warped_raster_paths,
+            resample_method_list=['near' for _ in base_raster_path_bands],
+            target_pixel_size=base_raster_pixel_sizes,
+            bounding_box_mode='union',
             base_vector_path_list=[aggregate_vector_path],
             raster_align_index=0)
-        clipped_raster = gdal.OpenEx(clipped_raster_path, gdal.OF_RASTER)
-        clipped_band = clipped_raster.GetRasterBand(base_raster_path_band[1])
     except ValueError as e:
         if 'Bounding boxes do not intersect' in repr(e):
             LOGGER.error(
@@ -1520,10 +1533,7 @@ def zonal_statistics(
             # and we didn't raise an exception, execution could get weird.
             raise
 
-    # make a shapefile that non-overlapping layers can be added to
-    driver = ogr.GetDriverByName('MEMORY')
-    disjoint_vector = driver.CreateDataSource('disjoint_vector')
-    spat_ref = aggregate_layer.GetSpatialRef()
+    #
 
     # Initialize these dictionaries to have the shapefile fields in the
     # original datasource even if we don't pick up a value later
@@ -1533,6 +1543,8 @@ def zonal_statistics(
         [agg_feat.GetFID() for agg_feat in aggregate_layer])
     agg_feat = None
     # Loop over each polygon and aggregate
+
+    # Calculate disjoint polygon sets
     if polygons_might_overlap:
         LOGGER.info('creating disjoint polygon set')
         disjoint_fid_sets = calculate_disjoint_polygon_set(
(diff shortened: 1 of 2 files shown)


tests/test_geoprocessing.py
```diff
@@ -1079,13 +1079,13 @@ def test_zonal_stats_all_outside(self):
         array = numpy.fliplr(numpy.flipud(
             numpy.array(range(n*n), dtype=numpy.int32).reshape((n, n))))
         _array_to_raster(
-            array, None, raster_path, projection_epsg=4326,
-            origin=(origin_x+n, origin_y-n), pixel_size=(-1, 1))
+            array, 255, raster_path, projection_epsg=4326,
```
emlys, line 1082: This seems like it was a bug in the test: lines 1075-1076 say it's meant to test intersection with a nodata pixel, but we weren't marking that pixels' value (255) as being nodata.

tests/test_geoprocessing.py
```diff
@@ -1338,7 +1338,8 @@ def test_zonal_statistics_nodata_is_zero(self):
         raster_path = os.path.join(self.workspace_dir, 'small_raster.tif')
         _array_to_raster(
             numpy.array([[1, 0], [1, 0]], dtype=numpy.int32), 0, raster_path,
-            origin=(origin_x, origin_y), pixel_size=(1.0, -1.0))
+            origin=(origin_x, origin_y), pixel_size=(1.0, -1.0),
+            projection_epsg=4326)
```
emlys, line 1446: This was a bug in the test: without specifying `projection_epsg`, it was defaulting to a different projection and did not actually overlap with the vector.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1358,7 +1358,7 @@ def interpolate_points(
 
 
 def zonal_statistics(
-        base_raster_path_band, aggregate_vector_path,
+        base_raster_path_bands, aggregate_vector_path,
```
phargogh: `base_raster_path_band_list`?  The docstring describes this as a list, and `_list` would be consistent with the rest of the pygeoprocessing API that uses a list of path/bands tuples.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1408,373 +1421,357 @@ def zonal_statistics(
...
+                '`base_raster_path_band` not formatted as expected.  Expected '
+                f'(path, band_index), received {repr(base_raster_path_band)}')
+
+    # Check that all input rasters are aligned. This should hold true if they
+    # have the same projection, geotransform, and bounding box.
+    for attr in ['geotransform', 'bounding_box', 'projection_wkt']:
+        vals = set()
+        for path, _ in base_raster_path_bands:
+            raster_info = get_raster_info(path)
+            vals.add(str(raster_info[attr]))
```
phargogh, line 1476: Nice!  Using `str` here should help us avoid most floating-point precision issues.

I don't know if this would complicate things further or even be necessary, but would it be worth adding some additional precision formatting?  Like, only the first 6 digits of a float after the decimal should be included, something like that, just in case the precision gets into numerically murky territory.  I'm not certain this is a good idea, what do you think?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1408,373 +1421,357 @@ def zonal_statistics(
...
+    # If the aggregate vector already contains a field called 'original_fid',
+    # this will break and GDAL will emit a warning.
+    temp_working_dir = tempfile.mkdtemp(dir=working_dir)
+    copy_path = os.path.join(temp_working_dir, 'vector_copy.gpkg')
+    target_vector = ogr.GetDriverByName('GPKG').CreateDataSource(copy_path)
+    target_layer_id = 'disjoint_vector'
+    target_layer = target_vector.CreateLayer(
+        name=target_layer_id,
+        srs=aggregate_layer.GetSpatialRef(),
+        geom_type=ogr.wkbPolygon)
```
phargogh, line 1517: Is the source vector required to be a polygon for us to compute statistics?  A rasterization technique like we have in this function should technically support varying types of geometries.

We should be able to get the geometry type from the source layer, or if that cannot be known for some reason or if there are multiple geometry types in a single layer, `ogr.wkbUnknown` can be used instead.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1408,373 +1421,357 @@ def zonal_statistics(
...
+            LOGGER.warning(
+                f'Value counts requested on a floating-point raster: {path}. '
+                'This can cause excessive memory usage if the raster has '
+                'continuous values.')
+        target_path = os.path.join(temp_working_dir, f'{i}.tif')
+        target_raster_path_band_list.append((target_path, band))
+        gdal.Warp(
+            destNameOrDestDS=target_path,
+            srcDSOrSrcDSTab=base_path,
+            format='GTIFF',
```
phargogh: Could I suggest we use the `VRT` format here?  With warping the file to a geotiff, we will always expect that the new file will be clipped to the subset, but that necessarily involves a bunch of disk I/O, compression and the like.  If we use a VRT, we're subsetting the raster without actually writing all that stuff out ... which is great if all we need to do is read the raster data!  I haven't gotten to the part of the function where these are read yet, so please forgive me if we do need to write to these layers.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1408,373 +1421,357 @@ def zonal_statistics(
...
+            destNameOrDestDS=fid_raster_path,
+            srcDS=target_vector,
+            allTouched=False,
+            attribute=fid_field_name,
+            noData=fid_nodata,
+            outputBounds=bbox_intersection,
+            xRes=raster_info['pixel_size'][0],
+            yRes=raster_info['pixel_size'][1],
+            format='GTIFF',
+            outputType=gdal.GDT_Int32,
```
phargogh: Since we're rasterizing FIDs (positive integers), we could use `gdal.GDT_Uint16` to have the same range of possible values with half the required memory/disk space.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1408,373 +1421,357 @@ def zonal_statistics(
...
+
+        for set_index, fid_raster_path in enumerate(fid_raster_paths):
+            LOGGER.info(
+                f'disjoint polygon set {set_index} of {len(fid_raster_paths)}')
+            fid_raster = gdal.OpenEx(fid_raster_path, gdal.OF_RASTER)
+            fid_band = fid_raster.GetRasterBand(1)
+
+            for offset in offset_list:
+                fid_block = fid_band.ReadAsArray(**offset)
+                data_block = data_band.ReadAsArray(**offset)
```
phargogh: I like that we have logging at these higher levels (for each raster, then for each polygon set)!  Would it be worth adding timed logging to the offset iteration?  I think there's an object in this file for it if that's something that you think would add value here.

src/pygeoprocessing/geoprocessing.py
phargogh: Come to think of it, we can easily discover the max values of each polygon set (`max(disjoint_fid_set)`), so if we _really_ wanted to minimize memory, disk I/O, etc. used by these rasters, we _could_ determine the GDAL datatype from the max values in the disjoint FID sets!  It's extra bookkeeping and might not have any tangible benefit, and without knowing how it'd profile I'd suspect that Uint16 would be just fine.  I leave that up to you if you think this would add value.

phargogh (changes requested):
Very exciting, thanks for putting this together!  I had a couple minor things that I'd love to hear your thoughts on, and very happy to chat about anything as would be helpful.

src/pygeoprocessing/geoprocessing.py
emlys: I wasn't sure about `_list` because it can be a list or a single tuple. I'll change it to `base_raster_path_band_list` for now.

src/pygeoprocessing/geoprocessing.py
emlys: Sure, that will work if I change the nodata value to `numpy.iinfo(numpy.uint16).max`

emlys: The optimization you suggest would be really neat to have, but it feels like it could be a whole separate PR. Maybe that could be abstracted away in a separate pygeoprocessing function?

src/pygeoprocessing/geoprocessing.py
emlys: Good idea! It seems to work just fine with VRT.

src/pygeoprocessing/geoprocessing.py
emlys: Could we maybe move this to a separate issue? It makes sense, but historically the zonal_statistics docstring and implementation have specifically required polygons (for instance, using `calculate_disjoint_polygon_set`). I would want to do some further testing to make sure that works with other geometries.

src/pygeoprocessing/geoprocessing.py
emlys: Oh that's an interesting point... I just used `str` to make the values hashable, so they could go in a set. I think in most cases, the rasters would have previously been run through `align_and_resize_raster_stack` or something equivalent, so I expect they'd be exactly the same. Maybe this could wait until we encounter a use case for it? That would also help inform what precision is needed.

src/pygeoprocessing/geoprocessing.py
emlys: Done!

src/pygeoprocessing/geoprocessing.py
phargogh: Yeah, that's a tougher one.  I suppose the technically-correct middle ground would be some kind of ordered `iterable`.  This is one of those things where when we redesign the API, we might consider naming that doesn't include the type, especially now that the type can be described with python's type hints.

For this parameter, I'll leave it up to you about `list` or not!

src/pygeoprocessing/geoprocessing.py
phargogh: >Maybe this could wait until we encounter a use case for it?

Yes, that sounds good to me.

>I just used str to make the values hashable, so they could go in a set.

Yep, this sounds like a good solution.

src/pygeoprocessing/geoprocessing.py
phargogh: I see this has already been moved to https://github.com/natcap/pygeoprocessing/issues/322, which is great.  Thanks for creating this!

src/pygeoprocessing/geoprocessing.py
phargogh: >The optimization you suggest would be really neat to have, but it feels like it could be a whole separate PR. Maybe that could be abstracted away in a separate pygeoprocessing function?

Eh, I'm not convinced it would really be worth it for the extra iteration it would impose given that [`max` is O(n)](https://wiki.python.org/moin/TimeComplexity) and the disjoint sets can be large.  In any case, let's leave it out of this PR.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1372,15 +1372,28 @@ def zonal_statistics(
     that are not aggregated use their bounding box to intersect with the
     raster for overlap statistics.
 
+    Statistics are calculated on the set of pixels that fall within each
+    feature polygon. If `ignore_nodata` is false, nodata pixels are considered
```
phargogh:
```suggestion
    feature polygon. If ``ignore_nodata`` is false, nodata pixels are considered
```

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1372,15 +1372,28 @@ def zonal_statistics(
...
+    'sum': sum of valid pixel values
+    'count': number of valid pixels
+    'nodata_count': number of nodata pixels
+    'value_counts': number of pixels having each unique value
+
     Note:
         There may be some degenerate cases where the bounding box vs. actual
         geometry intersection would be incorrect, but these are so unlikely as
         to be manually constructed. If you encounter one of these please email
         the description and dataset to jdouglass@stanford.edu.
```
phargogh:
```suggestion
        to be manually constructed. If you encounter one of these please create
        an issue at https://github.com/natcap/pygeoprocessing/issues with the
        datasets used.
```

phargogh (changes requested):
Thanks @emlys !  Everything looks good except for a handful of trivial changes that I made suggestions for.  I also left a comment about the parameter name with/without `_list` and I leave it up to you about what seems best here!  Leaving it as it was would preserve backwards compatibility, which might be nice.

Anyways, let me know what you think!

src/pygeoprocessing/geoprocessing.py
emlys: Good point about backwards compatibility. I reverted this to `base_raster_path_band`.

phargogh (approved):
Thanks, @emlys !

PR #314, opened 2023-03-28
Fixing another `shared_memory` leak in `raster_calculator`.

### LINKED ISSUE #247: Shared memory objects leaking and being cleaned up on process shutdown

### LINKED ISSUE #309: Change over to `mamba` for dependency installation in GHA
InVEST recently switched over to `mamba` for its GHA dependency installation, which has saved considerable time in actions.  It would be great to do the same here in this repo.

phargogh:
This PR fixes another shared memory leak in `raster_calculator` found during the NCI work.

Fixes #247 
Fixes #309

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,6 @@
 # --------------------
 # This file records the packages and requirements needed in order for
 # pygeoprocessing to work as expected.
-
 Cython
 GDAL>=3.0.4
 numpy>=1.10.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,4 +9,4 @@ numpy>=1.10.1
 Rtree>=0.8.3
 scipy>=0.14.1,!=0.19.1
 Shapely>=1.6.4
-importlib_metadata; python_version < '3.8'
+importlib_metadata  # technically only required on python < 3.8; easier to install with conda across all versions
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,3 +9,4 @@ numpy>=1.10.1
 Rtree>=0.8.3
 scipy>=0.14.1,!=0.19.1
 Shapely>=1.6.4
+importlib_metadata; python_version < '3.8'
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -589,13 +589,14 @@ def raster_calculator(
                     LOGGER.error("stats_worker_thread.join() timed out")
                     raise RuntimeError(
                         "stats_worker_thread.join() timed out")
-                if sys.version_info >= (3, 8) and use_shared_memory:
-                    LOGGER.debug(
-                        f'unlink shared memory for process {os.getpid()}')
-                    shared_memory.close()
-                    shared_memory.unlink()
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -9,3 +9,4 @@ numpy>=1.10.1
 Rtree>=0.8.3
 scipy>=0.14.1,!=0.19.1
 Shapely>=1.6.4
+importlib_metadata  # technically only required on python < 3.8; easier to install with conda across all versions
```
davemfish, line 11: Looks like this is not working in the `3.7` envs. But the failure is actually during environment creation, so maybe some other library is using this package and no longer getting the version it expected?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -589,13 +589,14 @@ def raster_calculator(
...
-                    LOGGER.debug(
-                        f'unlinked shared memory for process {os.getpid()}')
+
+            if sys.version_info >= (3, 8) and use_shared_memory:
+                LOGGER.debug(
+                    f'unlink shared memory for process {os.getpid()}')
+                shared_memory.close()
+                shared_memory.unlink()
+                LOGGER.debug(
+                    f'unlinked shared memory for process {os.getpid()}')
```
davemfish, line 599: Ahhh, whitespace. Github's side-by-side diff not doing me any favors here.

davemfish (changes requested):
Hey @phargogh , the change makes sense to me. But there's some sort of library incompatibility going in with python `3.7`

phargogh:
Thanks for pointing out the build failure there, @davemfish ! This turned out to be an unfortunate combination of an old version of `build` being installed and `setuptools` on its way out.  I upgraded to the latest version of the `build` package (distributed now on `conda-forge` as `python-build`) and also took this opportunity to roughly halve the GHA runtime by using `mamba` for environment management as we do in InVEST.  The python 3.7 tests appear to be passing so far, fingers crossed that it'll keep going for the rest of the tests.

This also means that this PR addresses #309

src/pygeoprocessing/geoprocessing.py
phargogh: Yeah, it's easy for the diff to hide this!  Isn't there even a whitespace-ignoring view?  So if the only changes were whitespace, you won't see the changes at all?

requirements.txt
phargogh: Yeah, this turned out to be an apparent incompatibility between the `build` and `importlib_metadata` packages that is resolved by upgrading to the latest version of `build`, which is now distributed on `conda-forge` as `python-build`.

davemfish (approved):
Thanks for figuring all that out!

PR #325, opened 2023-06-20
Expose `array_equals_nodata` function in api and add tests for it

emlys:
In preparation for https://github.com/natcap/invest/issues/1213, this PR adds `array_equals_nodata` to the `geoprocessing` API, and copies over its tests from invest.


phargogh (changes requested):
This all looks good!  Since we're officially exposing this in the `pygeoprocessing` namespace, could you add a note about it in `HISTORY`?  A quick search of the file doesn't turn up any hits.

phargogh (approved):

PR #327, opened 2023-07-19
Implement `raster_map`

### LINKED ISSUE #235: Proposal: `raster_calculator` wrapper

### LINKED ISSUE #328: Broken tests: `TypeError: issubclass() arg 1 must be a class`
Apparently caused by the Cython 3.0.0 release, though somehow it's not fixed by pinning cython<3.0.0. This test is failing:
https://github.com/natcap/pygeoprocessing/blob/1e92d5aaf16c478565599a682ea8b9e60445593f/tests/test_geoprocessing.py#L130-L144

because `isinstance(func, (types.FunctionType, types.BuiltinFunctionType))` is `False`, it goes on to the next line, which triggers the error. 

It's failing on the function `calculate_slope`, which happens to be the first attribute of `pygeoprocessing` that it checks. The type of `calculate_slope` is `<class '_cython_3_0_0.cython_function_or_method'>`, which is _not_ a subclass of `types.FunctionType`.

emlys:
Fixes #235, fixes #328

### DIFF
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -709,7 +709,13 @@ def raster_map(op, *rasters, target_path, target_nodata=None,
     Returns:
         ``None``
     """
-    nodatas = [get_raster_info(r)['nodata'][0] for r in rasters]
+    nodatas = []
+    for raster in rasters:
+        raster_info = get_raster_info(raster)
+        if raster_info['n_bands'] > 1:
+            LOGGER.warning(f'{raster} has more than one band. Only the first '
+                           'band will be used.')
+        nodatas.append(raster_info['nodata'][0])
 
     # choose an appropriate dtype if none was given
     if target_dtype is None:
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -14,7 +14,8 @@ import threading
 import time
 
 from . import geoprocessing_core
-from .geoprocessing_core import DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS
+from .geoprocessing_core import DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS, \
+    DEFAULT_CREATION_OPTIONS, INT8_CREATION_OPTIONS
 from .geoprocessing_core import DEFAULT_OSR_AXIS_MAPPING_STRATEGY
 from osgeo import gdal
 from osgeo import ogr
@@ -2075,10 +2076,10 @@ def warp_raster(
         warped_raster_path = target_raster_path
     base_raster = gdal.OpenEx(base_raster_path, gdal.OF_RASTER)
 
-    raster_creation_options = list(raster_driver_creation_tuple[1])
+    raster_creation_options = DEFAULT_CREATION_OPTIONS
     if (base_raster_info['numpy_type'] == numpy.int8 and
             'PIXELTYPE' not in ' '.join(raster_creation_options)):
-        raster_creation_options.append('PIXELTYPE=SIGNEDBYTE')
+        raster_creation_options = INT8_CREATION_OPTIONS
 
     # WarpOptions.this is None when an invalid option is passed, and it's a
     # truthy SWIG proxy object when it's given a valid resample arg.
@@ -4154,9 +4155,14 @@ def raster_op(op, *rasters, target_path='', target_nodata=None,
                 f'Target nodata value {target_nodata} is incompatible with '
                 f'the target dtype {target_dtype}')
 
+    creation_options = DEFAULT_CREATION_OPTIONS
+    if target_dtype == numpy.int8:
+        creation_options = INT8_CREATION_OPTIONS
+
     raster_calculator(
         [(path, 1) for path in rasters],  # assume the first band
         apply_op,
         target_path,
         NUMPY_TO_GDAL_TYPE[numpy.dtype(target_dtype)],
-        target_nodata)
+        target_nodata,
+        raster_driver_creation_tuple=('GTIFF', tuple(creation_options)))
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -31,6 +31,8 @@ from osgeo import osr
 
 from . import geoprocessing_core
 from .geoprocessing_core import DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS
+from .geoprocessing_core import DEFAULT_CREATION_OPTIONS
+from .geoprocessing_core import INT8_CREATION_OPTIONS
 from .geoprocessing_core import DEFAULT_OSR_AXIS_MAPPING_STRATEGY
 
 # This is used to efficiently pass data to the raster stats worker if available
@@ -4486,7 +4488,7 @@ def choose_nodata(dtype):
         return int(numpy.iinfo(dtype).max)
 
 
-def raster_op(op, *rasters, target_path='', target_nodata=None,
+def raster_map(op, *rasters, target_path='', target_nodata=None,
               target_dtype=None):
     """Apply a pixelwise function to a series of raster inputs.
 
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -648,6 +648,120 @@ def array_equals_nodata(array, nodata):
     return numpy.isclose(array, nodata, equal_nan=True)
 
 
+def choose_dtype(*raster_paths):
+    """
+    Choose an appropriate dtype for an output derived from the given inputs.
+
+    Returns the dtype with the greatest size/precision among the inputs, so
+    that information will not be lost.
+
+    Args:
+        *raster_paths: series of raster path strings
+
+    Returns:
+        numpy dtype
+    """
+    dtypes = []
+    for path in raster_paths:
+        raster = gdal.OpenEx(path, gdal.OF_RASTER)
+        band = raster.GetRasterBand(1)
+        dtypes.append(_gdal_to_numpy_type(band))
+    return numpy.result_type(*dtypes)
+
+
+def choose_nodata(dtype):
+    """
+    Choose an appropriate nodata value for data of a given dtype.
+
+    Args:
+        dtype (numpy.dtype): data type for which to choose nodata
+
+    Returns:
+        number to use as nodata value
+    """
+    try:
+        return float(numpy.finfo(dtype).max)
+    except ValueError:
+        return int(numpy.iinfo(dtype).max)
+
+
+def raster_map(op, *rasters, target_path='', target_nodata=None,
+              target_dtype=None):
+    """Apply a pixelwise function to a series of raster inputs.
+
+    The output raster will have nodata where any input raster has nodata.
+    Raster inputs are split into aligned blocks, and the function is
+    applied individually to each stack of blocks (as numpy arrays).
+
+    Args:
+        op (function): Function to apply to the inputs. It should accept a
+            number of arguments equal to the length of ``*inputs``. It should
+           return a numpy array with the same shape as its array input(s).
+        *rasters (strs): Paths to rasters to input to ``op``, in the order that
+           they will be passed to ``op``. All rasters should be aligned and
+           have the same dimensions.
+        target_path (str): path to write out the output raster.
+        target_nodata (number): Nodata value to use for the output raster.
+            Optional. If not provided, a suitable nodata value will be chosen.
+        target_dtype (numpy.dtype): dtype to use for the output. Optional. If
+            not provided, a suitable dtype will be chosen.
+
+    Returns:
+        ``None``
+    """
+    nodatas = [get_raster_info(r)['nodata'][0] for r in rasters]
+
+    def apply_op(*arrays):
+        """Apply the function `op` to the input arrays.
+
+        Args:
+            *arrays: numpy arrays with the same shape.
+
+        Returns:
+            numpy array
+        """
+        result = numpy.full(arrays[0].shape, target_nodata, dtype=target_dtype)
+
+        # make a mask that is True where all input arrays are valid,
+        # and False where any input array is invalid.
+        valid_mask = numpy.full(arrays[0].shape, True)
+        for array, nodata in zip(arrays, nodatas):
+            valid_mask &= ~array_equals_nodata(array, nodata)
+
+        # mask all arrays to the area where they all are valid
+        masked_arrays = [array[valid_mask] for array in arrays]
+        # apply op to the masked arrays in order
+        result[valid_mask] = op(*masked_arrays)
+        return result
+
+    # choose an appropriate dtype if none was given
+    if target_dtype is None:
+        target_dtype = choose_dtype(*rasters)
+
+    # choose an appropriate nodata value if none was given
+    # if the user provides a target nodata,
+    # check that it can fit in the target dtype
+    if target_nodata is None:
+        target_nodata = choose_nodata(target_dtype)
+    else:
+        if not numpy.can_cast(target_nodata, target_dtype):
+            raise ValueError(
+                f'Target nodata value {target_nodata} is incompatible with '
+                f'the target dtype {target_dtype}')
+
+    creation_options = DEFAULT_CREATION_OPTIONS
+    if target_dtype == numpy.int8:
+        creation_options = INT8_CREATION_OPTIONS
+
+    raster_calculator(
+        [(path, 1) for path in rasters],  # assume the first band
+        apply_op,
+        target_path,
+        NUMPY_TO_GDAL_TYPE[numpy.dtype(target_dtype)],
+        target_nodata,
+        raster_driver_creation_tuple=('GTIFF', tuple(creation_options)))
+
+
 def raster_reduce(function, raster_path_band, initializer, mask_nodata=True,
                   largest_block=_LARGEST_ITERBLOCK):
     """Cumulatively apply a reducing function to each block of a raster.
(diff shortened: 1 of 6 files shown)


src/pygeoprocessing/geoprocessing.py
```diff
@@ -632,6 +648,120 @@ def array_equals_nodata(array, nodata):
...
+            raise ValueError(
+                f'Target nodata value {target_nodata} is incompatible with '
+                f'the target dtype {target_dtype}')
+
+    creation_options = DEFAULT_CREATION_OPTIONS
+    if target_dtype == numpy.int8:
+        creation_options = INT8_CREATION_OPTIONS
+
+    raster_calculator(
+        [(path, 1) for path in rasters],  # assume the first band
```
phargogh, line 765: Since this is assuming that we will be using band 1, would it be worth logging a warning or something like that if more than 1 band is provided?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -632,6 +648,120 @@ def array_equals_nodata(array, nodata):
...
+    def apply_op(*arrays):
+        """Apply the function `op` to the input arrays.
+
+        Args:
+            *arrays: numpy arrays with the same shape.
+
+        Returns:
+            numpy array
+        """
+        result = numpy.full(arrays[0].shape, target_nodata, dtype=target_dtype)
```
phargogh, line 750: At this point in the function, `target_dtype` and `target_nodata`  might not yet have been defined, right?  What would happen here if either (or both)?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -632,6 +648,120 @@ def array_equals_nodata(array, nodata):
...
+        target_dtype (numpy.dtype): dtype to use for the output. Optional. If
+            not provided, a suitable dtype will be chosen.
+
+    Returns:
+        ``None``
+    """
+    nodatas = [get_raster_info(r)['nodata'][0] for r in rasters]
+
+    def apply_op(*arrays):
+        """Apply the function `op` to the input arrays.
```
phargogh:
```suggestion
        """Apply the function ``op`` to the input arrays.
```

src/pygeoprocessing/geoprocessing.py
```diff
@@ -632,6 +648,120 @@ def array_equals_nodata(array, nodata):
...
+        number to use as nodata value
+    """
+    try:
+        return float(numpy.finfo(dtype).max)
+    except ValueError:
+        return int(numpy.iinfo(dtype).max)
+
+
+def raster_map(op, *rasters, target_path='', target_nodata=None,
+              target_dtype=None):
```
phargogh: `target_path=''` is an interesting default value!  What happens if this is not overridden by the user?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -632,6 +648,120 @@ def array_equals_nodata(array, nodata):
...
+    if target_dtype is None:
+        target_dtype = choose_dtype(*rasters)
+
+    # choose an appropriate nodata value if none was given
+    # if the user provides a target nodata,
+    # check that it can fit in the target dtype
+    if target_nodata is None:
+        target_nodata = choose_nodata(target_dtype)
+    else:
+        if not numpy.can_cast(target_nodata, target_dtype):
```
phargogh: Oh that's a neat check.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -632,6 +648,120 @@ def array_equals_nodata(array, nodata):
...
+    if target_dtype == numpy.int8:
+        creation_options = INT8_CREATION_OPTIONS
+
+    raster_calculator(
+        [(path, 1) for path in rasters],  # assume the first band
+        apply_op,
+        target_path,
+        NUMPY_TO_GDAL_TYPE[numpy.dtype(target_dtype)],
+        target_nodata,
+        raster_driver_creation_tuple=('GTIFF', tuple(creation_options)))
```
phargogh: While creating a GeoTIFF is probably the most common use case, it is a little inconsistent with the existing raster_calculator function (and some other functions).  Would you be opposed to exposing an option for raster driver creation options, but defaulting to GTiff?  Of course we'd still need to handle the signed byte case for GTiffs as well.

tests/test_geoprocessing.py
```diff
@@ -5042,6 +5043,253 @@ def test_geometry_field_mismatch(self):
...
+        array = numpy.array([
+            [1, 2, 3],
+            [4, 5, 6],
+            [7, 8, 9]], dtype=numpy.float32)
+        path = os.path.join(self.workspace_dir, 'a.tif')
+        out_path = os.path.join(self.workspace_dir, 'out.tif')
+        _array_to_raster(array, -1, path)
+
+        pygeoprocessing.raster_map(
+            lambda a: a * 2, path, target_path=out_path)
```
phargogh: That does read nicely!

src/pygeoprocessing/geoprocessing.py
phargogh: Related, I wonder: would [keyword-only arguments](https://peps.python.org/pep-3102/#specification) be useful here?  I think that might be the intent with the way these are expressed here, but wanted to check!

phargogh (changes requested):
Nice!  I had a few questions about state under certain cases, but this is looking good!

src/pygeoprocessing/geoprocessing.py
emlys: Good point! Yep, I think I wanted it to be a keyword-only argument. In retrospect, I think it's cleaner to have `rasters` be a list. That more clearly separates what are the input paths vs the output path.

emlys: Well, I can't decide. Which do you like better?
```
raster_map(lambda a: a * 2, path, target_path=out_path)
raster_map(lambda a, b: a + b, path_a, path_b, target_path=out_path)

or

raster_map(lambda a: a * 2, [path], out_path)
raster_map(lambda a, b: a + b, [path_a, path_b], out_path)
```

src/pygeoprocessing/geoprocessing.py
emlys: Oh interesting, I agree that other formats should be allowed. In `warp_raster`, it looks like only the first element (the format) of `raster_driver_creation_tuple` is used - the other creation options are defined with a default like this.

src/pygeoprocessing/geoprocessing.py
emlys: It seems like this gets evaluated _after_ `target_dtype` and `target_nodata` are defined further down in `raster_map`, but it does make more sense to define them above!

src/pygeoprocessing/geoprocessing.py
emlys: Added a warning and test for this

src/pygeoprocessing/geoprocessing.py
emlys: I added a `target_driver` kwarg to `raster_map` to expose this option. The benefit of _not_ exposing the other creation options is that `raster_map` can handle setting the int8 `SIGNED_BYTE` flag if needed. Do you think that's sufficient for the `raster_map` use case, or should we expose the creation options tuple as well?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -632,6 +648,128 @@ def array_equals_nodata(array, nodata):
...
+        target_nodata = choose_nodata(target_dtype)
+    else:
+        if not numpy.can_cast(target_nodata, target_dtype):
+            raise ValueError(
+                f'Target nodata value {target_nodata} is incompatible with '
+                f'the target dtype {target_dtype}')
+
+    creation_options = DEFAULT_CREATION_OPTIONS
+    if target_dtype == numpy.int8:
+        creation_options = INT8_CREATION_OPTIONS
```
phargogh: Now that we're defining the filetype in the function signature, wouldn't we also need to allow the user to define creation options since GDAL has filetype-specific options?  These defaults are great for GTiffs but wouldn't they cause an error when used with other formats?

src/pygeoprocessing/geoprocessing.py
phargogh: Aha!  Yes, you're absolutely right ... since the variable exists, the value of the variable, even if changes after the function is defined, will be accessible to the function at the time the function is called.  And python has always worked this way, so I don't know why I thought it was any different!

src/pygeoprocessing/geoprocessing.py
phargogh: >In retrospect, I think it's cleaner to have rasters be a list. That more clearly separates what are the input paths vs the output path.

I also think it's cleaner to provide the rasters as a list.

Design-wise, it seems important to clearly separate the input paths from the output path.  With the existence of keyword-only arguments, though, I think that offers another way to separate things out so maybe we have 2 options here:

```python
# option 1: args are either positional or keyword
# raster_map(lambda a: a * 2, [path], out_path)
def raster_map(func, path_list, out_path):
    ...

# option 2: target path is a required keyword arg; TypeError when `target_path` is missing
# raster_map(lambda a: a * 2, path, target_path=out_path)
def raster_map(func, *input_paths, *, target_path):
    ...
```

Option 1 is certainly the most flexible in that all arguments can be either positional or keywords, and it is up to the user to decide what is best for them.  This sort of function signature is the most consistent with the current pygeoprocessing API.

Option 2 on the other hand is more restrictive in how it is called (`func` and `input_paths` _must_ be positional, `target_path` _must_ be a keyword), but it is also more like what you see in a lot of jupyter-based libraries.  It's a strong stylistic choice, but if we think it'd be useful I think that'd be OK!  Especially if we had an example of how to call it.

Anyways, given all of the above, I think I agree with the direction you're leaning, which would be option 1.

src/pygeoprocessing/geoprocessing.py
phargogh: Yeah, I think we need to expose the `creation_options` tuple as well.  As implemented, the defaults look like this in `geoprocessing_core.pyx`:

```python
DEFAULT_CREATION_OPTIONS = ('TILED=YES', 'BIGTIFF=YES', 'COMPRESS=LZW',
                             'BLOCKXSIZE=256', 'BLOCKYSIZE=256')
INT8_CREATION_OPTIONS = DEFAULT_CREATION_OPTIONS + (
    'PIXELTYPE=SIGNEDBYTE',)
```

The likely conflict here is that `DEFAULT_CREATION_OPTIONS` are GTiff-specific (here's the [ERS format](https://gdal.org/drivers/raster/ers.html), for example, with many fewer creation options).

`PIXELTYPE=SIGNEDBYTE` is a GDAL implementation detail that happens to be consistent across the data formats that support signed bytes, which are [EEDAI, ERS, GTA, GTiff, HFA, netCDF, PostGISRaster, EHDR, RRaster, Rasterlite2 and Zarr.](https://gdal.org/development/rfc/rfc87_signed_int8.html#details).

So, it makes sense that `PIXELTYPE=SIGNEDBYTE` would be added to the needed creation options in certain cases, but the base set of creation options should be specific to the target datatype, and since the user can change the datatype they should also be able to change the creation options.

phargogh (changes requested):
Requesting changes only because of the creation options question.  The API question is an interesting one!  But I agree with you, I think the list separates things most cleanly.

src/pygeoprocessing/geoprocessing.py
emlys: @phargogh I'm not sure I understand the difference between `raster_map(func, *input_paths, *, target_path)` and `raster_map(func, *input_paths, target_path)` - they seem to work the same? I do like that in this format, you don't have to use a list when passing in just one raster.

src/pygeoprocessing/geoprocessing.py
emlys: Oh, thanks for pointing that out. I added a `raster_driver_creation_tuple` argument consistent with other pygeoprocessing functions, but `PIXELTYPE=SIGNEDBYTE` will be added if necessary.

src/pygeoprocessing/geoprocessing.py
emlys: This is the case for `warp_raster` - it always uses the GTIFF-specific options

phargogh: To my eyes, `warp_raster` does use the options provided by `raster_driver_creation_tuple[1]` (which defaults to a GTiff with GTiff creation options), and then it adds the signed int8 creation option as well if it needs to.  Are you seeing the same thing?

```python
    raster_creation_options = list(raster_driver_creation_tuple[1])
    if (base_raster_info['numpy_type'] == numpy.int8 and
            'PIXELTYPE' not in ' '.join(raster_creation_options)):
        raster_creation_options.append('PIXELTYPE=SIGNEDBYTE')
```

emlys: Oh you're right. I must have accidentally changed that on my branch.

src/pygeoprocessing/geoprocessing.py
phargogh: Uffda, I'm sorry I confused myself here, @emlys !  I had the function signatures wrong in my snippet above.  The corrected options would be 
```python
def raster_map_1(func, input_paths, target_path):
    pass


def raster_map_2(func, *input_paths, target_path):
    pass


raster_map_1(lambda x: x**2, input_paths=['foo.tif', 'bar.tif'], target_path='baz.tif')
raster_map_1(lambda x: x**2, ['foo.tif', 'bar.tif'], 'baz.tif')

# This function can only be called this way; target_path must be a kwarg.
# Input_paths cannot be provided as a kwarg.
raster_map_2(lambda x: x**2, 'foo.tif', 'bar.tif', target_path='baz.tif')
```

src/pygeoprocessing/geoprocessing.py
```diff
@@ -734,9 +738,9 @@ def raster_map(op, *rasters, target_path, target_nodata=None,
                 f'Target nodata value {target_nodata} is incompatible with '
                 f'the target dtype {target_dtype}')
 
-    creation_options = DEFAULT_CREATION_OPTIONS
-    if target_dtype == numpy.int8:
-        creation_options = INT8_CREATION_OPTIONS
+    driver, options = raster_driver_creation_tuple
```
phargogh, line 737: That's a really nice, terse way to express these two things!

src/pygeoprocessing/geoprocessing.py
```diff
@@ -632,6 +648,132 @@ def array_equals_nodata(array, nodata):
...
+        *raster_paths: series of raster path strings
+
+    Returns:
+        numpy dtype
+    """
+    dtypes = []
+    for path in raster_paths:
+        raster = gdal.OpenEx(path, gdal.OF_RASTER)
+        band = raster.GetRasterBand(1)
+        dtypes.append(_gdal_to_numpy_type(band))
```
dcdenu4: Would it be worth using `get_raster_info` which returns `numpy_type`?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -632,6 +648,132 @@ def array_equals_nodata(array, nodata):
...
+    return numpy.result_type(*dtypes)
+
+
+def choose_nodata(dtype):
```
dcdenu4, line 668: Should this be exposed in `__init__.py`? If not, should we follow the convention of starting the function with an underscore?

dcdenu4 (commented):
Hey @emlys , sorry for jumping in here so late but once I was looking at the function definition options I couldn't resist looking through the rest! I'll reply separately about the open question, but also had a few quick comments / questions generally. This looks great!

src/pygeoprocessing/geoprocessing.py
dcdenu4: @emlys and @phargogh, my instinct is to go with option 1 because it feels most at home in the current API design. In another programming language we might be able to easily overload the function based on data type and allow for a single path OR a list of paths easily.

I also find it mildly annoying when entering a single path into a list or even a list of (path, band) tuples in some API instances. But in my opinion if I was using `raster_map` with multiple inputs I'd want to pass a list.

davemfish: I prefer option 1, though mostly from the perspective of someone reading code, rather than writing it. If I read code written with option 2, and I'm not yet super familiar with the function signature, it won't be immediately obvious that `foo.tif` and `bar.tif` are both inputs for the function. Whereas with option 1, there are always 3 args, so it's easy to intuit that the 1st is the function, 2nd is the inputs, 3rd is the output.

src/pygeoprocessing/geoprocessing.py
emlys: Sure, I'll add it to `__init__.py`.

src/pygeoprocessing/geoprocessing.py
emlys: Neat!

emlys:
@phargogh this is finally ready! I switched the signature to use a list of paths.

phargogh (approved):
Looks great and looking forward to using these new functions!

PR #333, opened 2023-08-16
`zonal_statistics`: expand bbox intersection to align with grid

### LINKED ISSUE #332: `zonal_statistics`: error when intersection between raster and vector is small
When the intersection between the input raster and vector is small (maybe smaller than a pixel, or not overlapping a pixel centerpoint?), `zonal_statistics` logs a couple of GDAL errors and then exits with an unhelpful error:
```
ERROR 1: Attempt to create 0x0 dataset is illegal,sizes must be larger than zero.
INFO:pygeoprocessing.geoprocessing:calculating disjoint polygon sets
INFO:pygeoprocessing.geoprocessing:build shapely polygon list
INFO:pygeoprocessing.geoprocessing:build shapely rtree index
INFO:pygeoprocessing.geoprocessing:poly feature lookup 100.0% complete on test_poly.gpkg
INFO:pygeoprocessing.geoprocessing:build poly intersection lookup
INFO:pygeoprocessing.geoprocessing:poly intersection feature lookup 100.0% complete on test_poly.gpkg
INFO:pygeoprocessing.geoprocessing:maximal subset build 100.0% complete on test_poly.gpkg
ERROR 1: Attempt to create 0x0 dataset is illegal,sizes must be larger than zero.
ERROR 1: Cannot create /var/folders/zh/dy02ljyd4mj5_nfm158hwcr00000gn/T/tmp7qys962o/fid_set_0.tif
INFO:pygeoprocessing.geoprocessing:calculating stats on raster 0 of 1
  File "/Users/emily/natural-capital-footprint-impact/test.py", line 94, in <module>
    pygeoprocessing.zonal_statistics(
  File "/Users/emily/miniconda3/envs/morgan-stanley/lib/python3.11/site-packages/pygeoprocessing/geoprocessing.py", line 1632, in zonal_statistics
    offset_list = list(iterblocks((raster_path, band), offset_only=True))
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/emily/miniconda3/envs/morgan-stanley/lib/python3.11/site-packages/pygeoprocessing/geoprocessing.py", line 3210, in iterblocks
    raise ValueError(
ValueError: Raster at /var/folders/zh/dy02ljyd4mj5_nfm158hwcr00000gn/T/tmp7qys962o/0.tif could not be opened.
```

The errors are coming from `gdal.Warp` - it's trying and failing to create a 0x0 raster (no pixels in the intersection). We should handle this case.

emlys:
This PR updates `zonal_statistics` to make sure that the bounding box used is aligned with the input raster stack. Coincidentally fixes #332.

The original bounding box is the intersection between the input raster(s) and vector. This box doesn't necessarily align with the input raster's grid. Having the intermediate rasters' grid shifted by a fraction of a pixel can change the results, because it changes which pixels' centerpoints fall within a polygon.

Keeping the grid aligned with the input will keep results consistent, independent of the disjoint polygon sets' bounding boxes. This also resolves most of the places where invest test results were changed by pygeoprocessing 2.4.1.


src/pygeoprocessing/geoprocessing.py
```diff
@@ -1357,6 +1357,44 @@ def interpolate_points(
...
+        (pixel_width < 0 and bbox[0] > x_origin) or
+        (pixel_height < 0 and bbox[1] > y_origin)):
+        raise ValueError('bbox must fall within geotransform grid')
+    return [
+        x_origin + abs(pixel_width) * math.floor(
+            (bbox[0] - x_origin) / pixel_width * numpy.sign(pixel_width)),
+        y_origin + abs(pixel_height) * math.floor(
+            (bbox[1] - y_origin) / pixel_height * numpy.sign(pixel_height)),
+        x_origin + abs(pixel_width) * math.ceil(
+            (bbox[2] - x_origin) / pixel_width * numpy.sign(pixel_width)),
```
emlys, line 1595: It took me a long time to figure out these equations and I'm still not sure I understand them intuitively. Happy to discuss

emlys: This would be interesting as a tough interview question

tests/test_geoprocessing.py
```diff
@@ -1044,8 +1044,8 @@ def test_zonal_stats_all_outside(self):
         new_feature.SetGeometry(new_geometry)
         layer.CreateFeature(new_feature)
 
-        x_pos = origin_x + (n+.99)
-        y_pos = origin_y - (n+.99)
+        x_pos = origin_x + (n-.01)
```
emlys, line 1050: This test was failing after the changes to zonal stats in this PR, and I'm not sure why it was passing in the first place. The original polygon is in purple and the new polygon is in brown. According to a comment above, this is meant to test "a polygon that barely intersects the upper left hand corner but is nodata." The original polygon actually intersects 3 valid pixels, with a sum result of `254 + 239 + 238 = 731`, not 0. I moved it up and left by one pixel to be in line with the intent of the test.
Screenshot 2023-09-19 at 11 03 46
.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1357,6 +1357,44 @@ def interpolate_points(
         band.WriteArray(raster_out_array, offset['xoff'], offset['yoff'])
 
 
+def align_bbox(geotransform, bbox):
+    """Pad a bounding box so that it aligns with the grid of a given geotransform.
+
+    Ignores row and column rotation.
```
phargogh, line 1559: If the user does provide rotation, could we log a warning or something like that?  I'm not entirely sure it's a failure state to provide rotation here and the output bbox will be close-ish to what it should be anyways, so it'd be great if we could at least have something appear in the logfile to help us debug if it were to happen.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1357,6 +1357,44 @@ def interpolate_points(
...
+    # NOTE: x_origin and y_origin do not necessarily equal minx and miny.
+    # If pixel_width is positive, x_origin = minx
+    # If pixel_width is negative, x_origin = maxx
+    # If pixel height is positive, y_origin = miny
+    # If pixel height is negative, y_origin = maxy
+    x_origin, pixel_width, _, y_origin, _, pixel_height = geotransform
+    if pixel_width == 0 or pixel_height == 0:
+        raise ValueError('Pixel width and height must not be 0')
+    if bbox[2] < bbox[0] or bbox[3] < bbox[1]:
+        raise ValueError('bbox must be in order [minX, minY, maxX, maxY]')
```
phargogh, line 1583: This is helpful!  I've been bitten by the ordering of bbox coordinates many times.

src/pygeoprocessing/geoprocessing.py
phargogh: I agree, this adjustment would be a neat puzzle-style interview question!  Relevant to the domain, useful and also a little bit obscure.

tests/test_geoprocessing.py
phargogh: Sounds great, and thanks for the graphic!

phargogh (changes requested):
This looks great @emlys !  Could you add a note to HISTORY, since we're exposing a new function at the top module level?

src/pygeoprocessing/geoprocessing.py
emlys: Sure!

emlys:
@phargogh I added a log warning if the geotransform includes row/column rotation, and also fixed an edge case that came up when running the invest tests.

emlys:
After looking more at the similar functionality in `create_raster_from_bounding_box`, I'm not seeing enough overlap that would make sense to factor out, but let me know if you disagree!

phargogh:
Yeah, let's leave them separate for now.  I think there might be a way for us to merge the functionality (especially now that we have 3 semi-related use cases!) but we should design the solution and implement/refactor in another PR.  I just created #348 for this, but if it does make sense to do, I think that'd make sense to do in a separate PR!

phargogh (approved):

PR #334, opened 2023-08-18
testing imports: check if `pygeoprocessing.__all__` members are callable rather than checking specific types

### LINKED ISSUE #328: Broken tests: `TypeError: issubclass() arg 1 must be a class`

davemfish:
Fixes #328 

Here's a proposed solution, anyway. I'm not sure if this is exactly equivalent to the prior test, but it's agnostic to Cython versions & types, and seems to support the `__init__` construction of `__all__`.

### DIFF
--- a/tests/test_geoprocessing.py
+++ b/tests/test_geoprocessing.py
@@ -133,15 +133,14 @@ class TestGeoprocessing(unittest.TestCase):
         # raises a SyntaxWarning.  Instead, I'll just ensure that every
         # attribute in pygeoprocessing.__all__ is a function that is available
         # at the pygeoprocessing level.
-        import inspect
         for attrname in pygeoprocessing.__all__:
             try:
                 func = getattr(pygeoprocessing, attrname)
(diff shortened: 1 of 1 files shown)


phargogh (approved):
Looks good to me!

emlys (approved):
👍

PR #335, opened 2023-08-18
standardizing approaches across `warp_raster` and `create_raster_from_bounding_box`

### LINKED ISSUE #321: `warp_raster` and `create_raster_from_bounding_box` sometimes create different raster dimensions on the same input
@davemfish describes a case in https://github.com/natcap/invest/issues/1312 where, given the same input rasters, the functions `pygeoprocessing.warp_raster` and `pygeoprocessing.create_raster_from_bounding_box` create different outputs from the same inputs.  The key difference is that `create_raster_from_bounding_box` will always round up (`numpy.ceil`), while `warp_raster` will use `numpy.isclose(residual, 0.0)` to see if the bounding box is numerically aligned to pixels.

Given the same target bounding box, pixel size, etc., these two functions should produce the same output.

davemfish:
Fixes #321 

This PR takes the approach we already use in `warp_raster` for determining a target raster size and applies it in `create_raster_from_bounding_box`. I thought about pulling this out to a function, but so far it only appears twice and is generally more readable when it's all inline, in my opinion.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -16,6 +16,10 @@ Unreleased Changes
 * Added a new function, ``pygeoprocessing.array_equals_nodata``, which returns
   a boolean array indicating which elements have nodata. It handles integer,
   float, and ``nan`` comparison, and the case where the nodata value is `None`.
+* Standardized the approach used in ``warp_raster`` and 
+  ``create_raster_from_bounding_box`` for determining the dimensions of the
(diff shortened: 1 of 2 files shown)


phargogh (approved):
Yep, I agree it makes sense to copy-paste this between these two functions.  When we have a third instance of this logic, we can see about abstracting it in a way that makes sense.

Thanks for addressing this!

PR #336, opened 2023-08-22
zonal stats: look up features by FID, not attribute filter #331

### LINKED ISSUE #331: `zonal_statistics` slow on polygons that overlap nodata
I introduced a bug in #311 that causes poor performance on polygons that only overlap nodata pixels in the base raster(s). Those polygons are getting skipped and then handled later on in the set of `unset_fids`, which is less efficient.

emlys:
Fixes #331 

Looking up features by FID, rather than `GetNextFeature()` on a layer with an attribute filter, improves the performance a lot. I'm guessing that the attribute filter feature is not optimized like I'd hoped. `GetNextFeature()` probably just goes through all features in order until it finds one matching the attribute filter, which explains my observation that it became slower with each iteration - the matching feature was further along each time.

Also dereferencing a couple of GDAL objects that I missed before.

### DIFF
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -1492,6 +1492,7 @@ def zonal_statistics(
         raise RuntimeError(
             f"Could not open layer {aggregate_layer_name} of {aggregate_vector_path}")
 
+
     # Define the default/empty statistics values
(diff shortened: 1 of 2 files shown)


phargogh (approved):
Looks like a nice, direct solution to me!

PR #337, opened 2023-08-27
Expose overview level to use in `warp_raster`

### LINKED ISSUE #326: `warp_raster` is missing an option to specify the overview level
If overviews are present in a raster, GDAL will automatically pick an overview level to use that it thinks is reasonable.  This leads to unexpected results.

`gdal.Warp` has an option to specify the `overviewLevel`, and if `overviewLevel=0`, it uses the base raster.  We should expose this in the `warp_raster` API.

phargogh:
`gdal.Warp` has a parameter that allows the user to specify the over view level to use when warping pixels.  The default behavior, from GDAL's perspective, is that if overviews are present, it will select the overview that is closest to the target resolution and then warp those pixels into the target raster.  This would save computational time.  Unfortunately, this also can lead to unexpected results in some cases, as we found in the NCI water quality pipeline.

The solution here is to expose the overview level to use as a parameter to `pygeoprocessing.warp_raster`, which defaults to `-1`, indicating the base raster.

Fixes #326

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -69,7 +69,7 @@ jobs:
           condarc: |
               channels:
                 - conda-forge
-          create-args: |
+          create-args: >-
               python=${{ matrix.python-version }}
               gdal=${{ matrix.gdal }}
               setuptools
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -62,7 +62,7 @@ jobs:
           fetch-depth: 0
 
       - name: setup-micromamba
-        uses: mamba-org/provision-with-micromamba@main
+        uses: mamba-org/setup-micromamba@v1
         with:
           # Grab requirements from pip-compatible requirements.txt
           environment-file: requirements.txt
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -66,6 +66,9 @@ jobs:
         with:
           # Grab requirements from pip-compatible requirements.txt
           environment-file: requirements.txt
+          condarc: |
+              channels:
+                - conda-forge
           create-args: |
               python=${{ matrix.python-version }}
               gdal=${{ matrix.gdal }}
@@ -74,7 +77,6 @@ jobs:
               flake8
               pytest
           environment-name: pyenv
-          channels: conda-forge
 
       - name: Lint with flake8
         run: |
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -66,7 +66,7 @@ jobs:
         with:
           # Grab requirements from pip-compatible requirements.txt
           environment-file: requirements.txt
-          extra-specs: |
+          create-args: |
               python=${{ matrix.python-version }}
               gdal=${{ matrix.gdal }}
               setuptools
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -2196,7 +2196,7 @@ def warp_raster(
         base_raster_path, target_pixel_size, target_raster_path,
         resample_method, target_bb=None, base_projection_wkt=None,
         target_projection_wkt=None, n_threads=None, vector_mask_options=None,
-        gdal_warp_options=None, working_dir=None, use_overview_level=0,
+        gdal_warp_options=None, working_dir=None, use_overview_level=-1,
         raster_driver_creation_tuple=DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS,
         osr_axis_mapping_strategy=DEFAULT_OSR_AXIS_MAPPING_STRATEGY):
     f"""Resize/resample raster to desired pixel size, bbox and projection.
@@ -2242,10 +2242,15 @@ def warp_raster(
         working_dir (string): if defined uses this directory to make
             temporary working files for calculation. Otherwise uses system's
             temp directory.
-        use_overview_level=0 (int): The overview level to use for warping.
-            A value of ``0`` (the default) indicates that the base raster
-            should be used.
-            # TODO: what about other overview levels?
+        use_overview_level=-1 (int/str): The overview level to use for warping.
+            A value of ``-1`` (the default) indicates that the base raster
+            should be used for the source pixels. A value of ``'AUTO'``
+            will make GDAL select the overview with the resolution that is
+            closest to the target pixel size and warp using that overview's
+            pixel values.  Any other integer indicates that that overview index
+            should be used.  For example, suppose the raster has overviews at
+            levels 2, 4 and 8.  To use level 2, set ``use_overview_level=0``.
+            To use level 8, set ``use_overview_level=2``.
         raster_driver_creation_tuple (tuple): a tuple containing a GDAL driver
             name string as the first element and a GDAL creation options
             tuple/list as the second. Defaults to a GTiff driver tuple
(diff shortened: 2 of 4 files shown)


src/pygeoprocessing/geoprocessing.py
```diff
@@ -97,19 +98,48 @@ def __init__(self, missing_values, raster_path, value_map):
...
+    #
+    # In GDAL < 3.4, an error would be logged if GDAL couldn't recognize the
+    # option.  Pushing a null error handler avoids this and cleans up the
+    # logging.
+    gdal.PushErrorHandler(lambda *args: None)
+    _options = []
+    # GDAL's python bindings only offer this one way of translating gdal.GRA_*
+    # options to their string names (but compatibility is limited in different
+    # GDAL versions, see notes above)
+    warp_opts = gdal.WarpOptions(
```
davemfish, line 120: Do you mind explaining to me what this call does? It doesn't seem like we're using the return value anywhere. And we weren't using it before these changes either. So does it have some side-effect? GDAL docs suggest the return of `gdal.WarpOptions` can be passed to `gdal.Warp`. https://gdal.org/api/python/osgeo.gdal.html#osgeo.gdal.Warp

src/pygeoprocessing/geoprocessing.py
```diff
@@ -97,19 +98,48 @@ def __init__(self, missing_values, raster_path, value_map):
...
+        options=_options,  # Add rendered flags to this list.
+        resampleAlg=getattr(gdalconst, _warp_algo),  # use GRA_* name.
+        overviewLevel=None)  # Don't add overview parameters.
+    gdal.PopErrorHandler()
+
+    for _item in _options:
```
davemfish, line 127: Nothing was ever pushed to `_options`, right? Is it still an empty list?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2402,6 +2432,15 @@ def warp_raster(
         working_dir (string): if defined uses this directory to make
             temporary working files for calculation. Otherwise uses system's
             temp directory.
+        use_overview_level=-1 (int/str): The overview level to use for warping.
+            A value of ``-1`` (the default) indicates that the base raster
```
davemfish, line 2437: I like defaulting to `-1` instead of `AUTO`

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2402,6 +2432,15 @@ def warp_raster(
...
+            should be used for the source pixels. A value of ``'AUTO'``
+            will make GDAL select the overview with the resolution that is
+            closest to the target pixel size and warp using that overview's
+            pixel values.  Any other integer indicates that that overview index
+            should be used.  For example, suppose the raster has overviews at
+            levels 2, 4 and 8.  To use level 2, set ``use_overview_level=0``.
+            To use level 8, set ``use_overview_level=2``.
```
davemfish, line 2444: Thanks for documenting this.

davemfish (commented):
@phargogh , overview stuff looks good. I just have questions about the `WarpOptions` stuff. Thanks!

.github/workflows/pythonpackage.yml
```diff
@@ -62,7 +62,7 @@ jobs:
           fetch-depth: 0
 
       - name: setup-micromamba
-        uses: mamba-org/provision-with-micromamba@main
+        uses: mamba-org/setup-micromamba@v1
```
davemfish, line 65: It looks like the `extra-specs` property was renamed to `create-args`. https://github.com/mamba-org/setup-micromamba#create-environment-from-environment-specs

This should resolve the failed checks

davemfish (changes requested):

src/pygeoprocessing/geoprocessing.py
phargogh: Yes, `gdal.WarpOptions()` returns a SWIG object representing the selected warp options that we have provided.  Unfortunately, it isn't at all clear how to access the returned object.  The only attribute I can actually read is `this`, which isn't iterable, doesn't have a `__dict__` object and has no attributes that apparently do anything that I can identify.  So my best alternative is to use the `options=` kwarg to `WarpOptions`, which has a side effect of having the parameters written to the list we pass in.

Here's the `dir(warp_opts)` readout ... maybe you'll see something I've missed?

```python
>>> warp_opts
(<osgeo.gdal.GDALWarpAppOptions; proxy of <Swig Object of type 'GDALWarpAppOptions *' at 0x105b33060> >, None, None)
>>> dir(warp_opts[0])
['__class__', '__del__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattr__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__swig_destroy__', '__swig_getmethods__', '__swig_setmethods__', '__weakref__', 'this']
>>> warp_opts[0].__dict__
{'this': <Swig Object of type 'GDALWarpAppOptions *' at 0x105b33060>}
>>> dir(warp_opts[0].this)
['__class__', '__delattr__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__int__', '__le__', '__lt__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', 'acquire', 'append', 'disown', 'next', 'own']
```

src/pygeoprocessing/geoprocessing.py
phargogh: `_options` has `gdalwarp` CLI options added to it by the `gdal.WarpOptions` call above.

src/pygeoprocessing/geoprocessing.py
phargogh: Definitely!  `AUTO` is also the GDAL default, which also means that you may end up with unexpected results in some warping cases if you don't know that GDAL does this.

.github/workflows/pythonpackage.yml
phargogh: Thanks!  Patched in https://github.com/natcap/pygeoprocessing/pull/337/commits/6c3cae7d54e42576cbedada5dfa5c3956777fd87

src/pygeoprocessing/geoprocessing.py
phargogh: I added an inline comment about this in https://github.com/natcap/pygeoprocessing/pull/337/commits/7dc6d57c3a414a7c1dce1f0a33e5908c996ba0b7

src/pygeoprocessing/geoprocessing.py
phargogh: Clarified the side effect in a comment in https://github.com/natcap/pygeoprocessing/pull/337/commits/7dc6d57c3a414a7c1dce1f0a33e5908c996ba0b7

src/pygeoprocessing/geoprocessing.py
```diff
@@ -118,11 +118,12 @@ def __init__(self, missing_values, raster_path, value_map):
     # options to their string names (but compatibility is limited in different
     # GDAL versions, see notes above)
     warp_opts = gdal.WarpOptions(
-        options=_options,  # Add rendered flags to this list.
+        options=_options,  # SIDE EFFECT: Adds flags to this list.
```
davemfish, line 121: Thank you!

src/pygeoprocessing/geoprocessing.py
davemfish: >So my best alternative is to use the options= kwarg to WarpOptions, which has a side effect of having the parameters written to the list we pass in.

Thank you! This is what I was missing.

davemfish (approved):

PR #338, opened 2023-08-27
Watershed delineation: handling the case where flow_dir doesn't have an SRS.

### LINKED ISSUE #254: delineate_watersheds_d8 does not handle the case where SRS is undefined
D8 watershed delineation implicitly requires that the flow direction raster has a defined spatial reference and doesn't handle the case where the SRS might not be defined.  The projection is not actually used in the delineation, it's only assigned to the output datasets.  So, we should handle this case and allow watershed delineation to proceed or throw an error and make this really explicit.

We should also make sure that the flow direction raster and outflow vector have the same projections or that both have no projection.

phargogh:
Also noting this change in HISTORY and testing.

Fixes #254

### DIFF
--- a/src/pygeoprocessing/routing/watershed.pyx
+++ b/src/pygeoprocessing/routing/watershed.pyx
@@ -628,10 +628,16 @@ def delineate_watersheds_d8(
         target_layer_name='watersheds'):
     """Delineate watersheds for a vector of geometries using D8 flow dir.
 
+    Note:
+        The ``d8_flow_dir_raster_path_band`` and ``outflow_vector_path`` files
+        must have the same spatial reference system.  The output watersheds
+        vector will use this same spatial reference system.
+
     Args:
         d8_flow_dir_raster_path_band (tuple): A (path, band_id) tuple
             to a D8 flow direction raster.  This raster must be a tiled raster
-            with block sizes being a power of 2.
+            with block sizes being a power of 2.  The output watersheds vector
+            will have its spatial reference copied from this raster.
         outflow_vector_path (str): The path to a vector on disk containing
             features with valid geometries from which watersheds will be
             delineated.  Only those parts of the geometry that overlap valid
--- a/src/pygeoprocessing/routing/watershed.pyx
+++ b/src/pygeoprocessing/routing/watershed.pyx
@@ -701,7 +701,8 @@ def delineate_watersheds_d8(
 
     gtiff_driver = gdal.GetDriverByName('GTiff')
     flow_dir_srs = osr.SpatialReference()
-    flow_dir_srs.ImportFromWkt(flow_dir_info['projection_wkt'])
+    if flow_dir_info['projection_wkt']:
+        flow_dir_srs.ImportFromWkt(flow_dir_info['projection_wkt'])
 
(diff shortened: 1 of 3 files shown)


emlys (commented):
@phargogh this looks good, but I'm not sure I understand when it would be OK to have a missing SRS? How do we know that the flow direction raster intersects with the other inputs?

phargogh:
> when it would be OK to have a missing SRS? How do we know that the flow direction raster intersects with the other inputs?

Well, we don't know for sure!  But if the units of the geotransform and geometries are all the same, it may be reasonable to assume that the user knows what they are doing and continue running.  So we would be assuming that the inputs are in some projected coordinate system and valid for delineating watersheds.  I don't think that's a reasonable assumption to make in InVEST, specifically, but I think it could be OK for a geoprocessing routine like this.

What do you think?

emlys:
I wonder if it makes sense to have a larger conversation about this and decide when/if this makes sense in general, so we can be consistent within pygeoprocessing? For instance, maybe it's also ok for the outflow vector to have no SRS? But if both inputs have no SRS, is that a problem?

phargogh:
The consistency question is a good one, and probably worth a discussion!  I'll add it to the agenda for an upcoming software team call.

In the shorter term, it turns out that we have a valid use case for where there might not be a defined spatial reference: in the various [supported raster formats](https://gdal.org/drivers/raster/index.html) that don't support georeferencing.  Taking this [CEOS](https://gdal.org/drivers/raster/ceos.html#raster-ceos) example [file from the GDAL tests](https://github.com/OSGeo/gdal/blob/master/autotest/gdrivers/data/ceos/IMAGERY-75K.L-3), the SRS's WKT is falsey:

```python
>>> from osgeo import gdal
>>> ds = gdal.Open('/Users/jdouglass/Downloads/IMAGERY-75K.L-3')
>>> ds.GetProjection()
''
```

phargogh:
Thanks for your thoughts and the review @emlys !  I created a new issue #341 for the general pygeoprocessing question in line with the discussion this morning, and I improved the docstring in the watershed delineation routine.

emlys (changes requested):
Thanks for adding that note to the docstring! It just looks like some extra carraige returns got in there, otherwise good!

phargogh:
Huh, I'm so used to my editor chomping them that I missed them in the diff!  Thanks @emlys I think they should be clear now

src/pygeoprocessing/routing/watershed.pyx
```diff
@@ -628,10 +628,16 @@ def delineate_watersheds_d8(
         target_layer_name='watersheds'):
     """Delineate watersheds for a vector of geometries using D8 flow dir.
 
+    Note:
```
emlys, line 631:
```suggestion
    Note:
```

emlys (changes requested):
Sorry @phargogh but I'm seeing one more `^M` on line 631, those line ending characters are sneaky!

phargogh:
Ugh, the lesson for me here is to be _very_ careful when copying code out of a git diff!  Thanks @emlys and I'm sorry for missing these!

Patched and pushed in https://github.com/natcap/pygeoprocessing/pull/338/commits/5ab4a460889e3c1480595028b5bdded033bdc0ba

emlys (approved):

PR #339, opened 2023-08-27
Improving references to TauDEM in routing

### LINKED ISSUE #255: Fix documentation about TauDEM flow directions
In our documentation, we sometimes say that we use TauDEM's flow direction conventions.  Maybe this was the case once, but it is no longer the case.  [According to the TauDEM docs, they now use indicies 1-8 instead of 0-7](https://hydrology.usu.edu/taudem/taudem5/help53/D8ContributingArea.html).

We should update the docs, and maybe provide an example of how to convert a TauDEM flow direction raster to pygeoprocessing-style flow direction, or else maybe include a function ([example](https://gist.github.com/phargogh/435db78e062bcb53baca5cdd8929e4bd))  somewhere to convert TauDEM flow direction to PGP.

This came up in the NCI Nitrate Concentration work.

phargogh:
Fixes #255

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -16,10 +16,14 @@ Unreleased Changes
 * Added a new function, ``pygeoprocessing.array_equals_nodata``, which returns
   a boolean array indicating which elements have nodata. It handles integer,
   float, and ``nan`` comparison, and the case where the nodata value is `None`.
-* Standardized the approach used in ``warp_raster`` and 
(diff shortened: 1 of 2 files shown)


emlys (approved):
Looks good, thanks @phargogh!

emlys:
There is a conflict in HISTORY.rst, feel free to merge once that's resolved

PR #343, opened 2023-09-01
Handling complex stddev values from `raster_calculator`'s stats worker

### LINKED ISSUE #342: Stats worker may return a complex when pygeoprocessing is built with `Cython >= 3.0.0`
I came across this exception after rebuilding my container for NCI (forgive the line numbers):

```python
8163 Traceback (most recent call last):
8164   File "/opt/conda/lib/python3.9/multiprocessing/pool.py", line 125, in worker
8165     result = (True, func(*args, **kwds))
8166   File "/home/users/jadoug06/workspace/2021-nci-ndr-sherlock/prep-ndr-inputs-pipeline.py", line 245, in intensified_rainfed_n_app
8167     pygeoprocessing.raster_calculator(
8168   File "/opt/conda/lib/python3.9/site-packages/pygeoprocessing/geoprocessing.py", line 569, in raster_calculator
8169     float(target_stddev))
8170 TypeError: can't convert complex to float
```

It turns out that in Cython 3.0.0, they included the following change:
> The ** power operator now behaves more like in Python by returning the correct complex result if required by math. A new cpow directive was added to turn on the previous C-like behaviour. (Github issue [#4936](https://github.com/cython/cython/issues/4936))

When the `raster_calculator` stats worker does its final computation for the standard deviation, it does the following, taking the square root by way of raising to the 0.5 power:

```python
    if n > 0:
        stats_work_queue.put(
            (min_value, max_value, M_local,
                (S_local / <double>n) ** 0.5))
```

phargogh:
I wasn't sure how to test this effectively since this is an environment issue and pygeoprocessing currently requires Cython < 3.0.

Fixes #342

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -37,6 +37,10 @@ Unreleased Changes
   where the function would raise an Exception when the target raster path was
   provided as a filename only, not within a directory, even though the parent
   directory could be inferred. https://github.com/natcap/pygeoprocessing/issues/313
+* Fixing a bug where the statistics worker in
+  ``pygeoprocessing.raster_calculator`` may return a complex value. This is
+  only an issue when pygeoprocessing is compiled against Cython 3.0.0 and
+  later. https://github.com/natcap/pygeoprocessing/issues/342
 
 2.4.0 (2023-03-03)
 ------------------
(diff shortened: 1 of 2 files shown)


davemfish (approved):
@phargogh I don't think this needs to be tested

PR #347, opened 2023-09-19
DOCS: cleanup TOC and various docstring format issues

### LINKED ISSUE #346: API Docs: some missing docs; noisy table of contents
Sphinx is not rendering docs for `warp_raster`, `stitch_raster`, and `build_overviews`. Those docstrings are constructed with an `f-string`, which is not allowed by sphinx.

I'm also going to try to reduce the redundancy in the API reference table of contents:

API REFERENCE

[pygeoprocessing](https://pygeoprocessing.readthedocs.io/en/latest/api/modules.html)
[pygeoprocessing package](https://pygeoprocessing.readthedocs.io/en/latest/api/pygeoprocessing.html)
[pygeoprocessing.geoprocessing module](https://pygeoprocessing.readthedocs.io/en/latest/api/pygeoprocessing.geoprocessing.html#)
[pygeoprocessing.geoprocessing_core module](https://pygeoprocessing.readthedocs.io/en/latest/api/pygeoprocessing.geoprocessing_core.html)
[pygeoprocessing.kernels module](https://pygeoprocessing.readthedocs.io/en/latest/api/pygeoprocessing.kernels.html)
[pygeoprocessing.multiprocessing package](https://pygeoprocessing.readthedocs.io/en/latest/api/pygeoprocessing.multiprocessing.html)
[pygeoprocessing.multiprocessing.raster_calculator module](https://pygeoprocessing.readthedocs.io/en/latest/api/pygeoprocessing.multiprocessing.raster_calculator.html)
[pygeoprocessing.routing package](https://pygeoprocessing.readthedocs.io/en/latest/api/pygeoprocessing.routing.html)
[pygeoprocessing.routing.helper_functions module](https://pygeoprocessing.readthedocs.io/en/latest/api/pygeoprocessing.routing.helper_functions.html)
[pygeoprocessing.routing.routing module](https://pygeoprocessing.readthedocs.io/en/latest/api/pygeoprocessing.routing.routing.html)
[pygeoprocessing.routing.watershed module](https://pygeoprocessing.readthedocs.io/en/latest/api/pygeoprocessing.routing.watershed.html)
[pygeoprocessing.symbolic module](https://pygeoprocessing.readthedocs.io/en/latest/api/pygeoprocessing.symbolic.html)

I find it confusing since there are multiple pathways to get to the same content. And depending on the path you take, the namespace could look different. For example, these entries appear on two different pages:

`pygeoprocessing.geoprocessing.align_and_resize_raster_stack(...)`
`pygeoprocessing.align_and_resize_raster_stack(...)`

They are both correct and both exist in the API, but I don't think it's helpful to be so "complete" in the documentation. I would rather only show people second entry.

davemfish:
Fixes #346 

I'm interested in feedback & opinions about the table of contents before merging. 

Edit: everybody was on board with the TOC change.

I also added some custom css to increase the font-size of the nested TOC items.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -10,8 +10,7 @@ sphinx:
 conda:
     environment: docs/environment-rtd.yml
 
-python:
-    version: 3.8
-    install:
-        - method: setuptools
-          path: .
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "mambaforge-22.9"
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -14,3 +14,6 @@ build:
   os: ubuntu-22.04
   tools:
     python: "mambaforge-22.9"
+  jobs:
+    post_install:
+      - pip install .
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -109,6 +109,7 @@ class TimedLoggingAdapter(logging.LoggerAdapter):
     This object is helpful for creating consistency in logging callbacks and is
     derived from the python stdlib ``logging.LoggerAdapter``.
     """
+
     def __init__(self, interval_s=_LOGGING_PERIOD):
         """Initialize the timed logging adapter.
 
@@ -1376,12 +1377,12 @@ def zonal_statistics(
     feature polygon. If ``ignore_nodata`` is false, nodata pixels are considered
     valid when calculating the statistics:
 
-    'min': minimum valid pixel value
-    'max': maximum valid pixel value
-    'sum': sum of valid pixel values
-    'count': number of valid pixels
-    'nodata_count': number of nodata pixels
-    'value_counts': number of pixels having each unique value
+    - 'min': minimum valid pixel value
+    - 'max': maximum valid pixel value
+    - 'sum': sum of valid pixel values
+    - 'count': number of valid pixels
+    - 'nodata_count': number of nodata pixels
+    - 'value_counts': number of pixels having each unique value
 
     Note:
         There may be some degenerate cases where the bounding box vs. actual
@@ -2199,7 +2200,7 @@ def warp_raster(
         gdal_warp_options=None, working_dir=None,
         raster_driver_creation_tuple=DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS,
         osr_axis_mapping_strategy=DEFAULT_OSR_AXIS_MAPPING_STRATEGY):
-    f"""Resize/resample raster to desired pixel size, bbox and projection.
+    """Resize/resample raster to desired pixel size, bbox and projection.
 
     Args:
         base_raster_path (string): path to base raster.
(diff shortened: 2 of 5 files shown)


.readthedocs.yml
```diff
@@ -10,8 +10,10 @@ sphinx:
...
 
-python:
-    version: 3.8
-    install:
-        - method: setuptools
-          path: .
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "mambaforge-22.9"
```
phargogh, line 16: I love that this takes the build time down from [1265s](https://readthedocs.org/projects/pygeoprocessing/builds/21832922/) to [276s](https://readthedocs.org/projects/pygeoprocessing/builds/21970683/)

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2199,16 +2200,17 @@ def warp_raster(
...
         base_raster_path (string): path to base raster.
         target_pixel_size (list/tuple): a two element sequence indicating
             the x and y pixel size in projected units.
         target_raster_path (string): the location of the resized and
             resampled raster.
-        resample_method (string): the resampling technique, one of
-            ``{_GDAL_WARP_ALGOS_FOR_HUMAN_EYES}``
+        resample_method (string): the resampling technique, one of,
+            'rms | mode | sum | q1 | near | q3 | average | cubicspline |
+            bilinear | max | med | min | cubic | lanczos'
```
phargogh, line 2213: Could you say more about this change?

My thinking behind making this an f-string was that different GDAL versions have different resample methods, and having a way to detect them programmatically and render them to a string for insertion into the docstring meant that we don't need to maintain it quite as much when these options change.  Having said that, I don't think I verified that the strings would render in the API docs, so maybe this is the only way to have them display?

phargogh (commented):
Thanks @davemfish !  Overall this looks great, but I was curious about reverting from f-string insertion of the warp algorithms back to statically defining the options.

src/pygeoprocessing/geoprocessing.py
davemfish: So I changed it because sphinx was not rendering docstrings for these functions at all. It didn't raise any errors/warnings about it, but `pydocstyle` couldn't handle it either: https://github.com/PyCQA/pydocstyle/issues/368

I didn't realize that different GDAL versions have different resample methods, though it makes sense. Unfortunately I don't think the variable in the f-string ever gets expanded, whether you're reading the docs on readthedocs, or in the source code, you never get to see the values. 

Alternatively, we could just say, "must be one of the options supported by `gdal.Warp`", or something like that. What do you think? @phargogh

phargogh: Wow!  That is both neat and highly unexpected.  Sphinx must just be flexing/parsing (but not interpreting), and/or using an old version of the language.  Either way, good to know!

So, for the current docstring, let's just leave it as-is.  I have only ever seen gdal add resample methods (except for the one time they renamed 'nearest' to 'near'), so this is likely to be a subset in future releases.

phargogh (approved):

PR #351, opened 2023-09-25
Improve `int` memory usage throughout routing

### LINKED ISSUE #246: Pitfilling and D8 flow direction progress logging displays integer overflow on large rasters
This came up on the forums recently (https://community.naturalcapitalproject.org/t/questions-about-running-delineatelt/2577/1).  You can see from the screenshots that on very large DEMs, we're getting an integer overflow when tracking the total number of pixels that should be able to be resolved with a higher-memory counter int, maybe even an unsigned int.

We should take a look at all of the progress logging in `routing.pyx` to see if there are any other functions where this should be updated as well.

### LINKED ISSUE #350: MFD flow direction producing nodata holes on super large rasters
For a NatCap project, we have a 30m DEM of Colombia that is resulting in disjointed streams.  When a single watershed within Colombia is used instead, the streams are contiguous, as we expect.

I dug into this a bit more and the root cause appears to be that the flow direction routine is producing nodata holes where the streams are becoming disjointed, but only on the super large raster.  I have the data.

I'd guess there's an integer overflow issue happening somewhere here because of the size of these rasters.

phargogh:
This PR addresses two issues in `routing.pyx`:

1. Fixes #246 : progress logging throughout `routing.py` is improved.  Specifically:
   * The overflow issue has been addressed (progress logging should work on rasters with up to 2^64 pixels now)
   * All progress logging now states what operation is running. No more plain `40.2% complete` statements.
2. Fixes #350 : I believe the key fix was that the plateau distance nodata value was suffering an integer overflow due to being the product of two signed integers, which meant the compiler most likely inferred that the result should also be a signed integer.  Forcing this into an unsigned long (which will then be cast to a `double` by GDAL when written out to a Float64 raster) resolves the issue.

I ended up adjusting most integers that represent unsigned integer data to unsigned integers to increase the range of available values possible before we run into this issue again.  In most cases, `unsigned int` is plenty large for our purposes.  If we had a DEM of 2^32 pixels wide at a 1m resolution, that would be 10 times larger than the circumference of the earth!

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,6 +1,16 @@
 Release History
 ===============
 
+Unreleased Changes
+------------------
+* Fixed an issue where MFD flow direction was producing many nodata holes given
+  a large-enough DEM.  These nodata holes would then propagate to flow
+  accumulation and stream extraction, producing very disjointed stream
+  networks. https://github.com/natcap/pygeoprocessing/issues/350
+* Improved progress logging in MFD flow direction, MFD flow accumulation, MFD
+  stream extraction to prevent integer overflows in percentages and improve
+  the readability of log messages. https://github.com/natcap/pygeoprocessing/issues/246
+
 2.4.1 (2023-09-05)
 ------------------
 * The ``pygeoprocessing`` package metadata has been updated to use
--- a/src/pygeoprocessing/routing/routing.pyx
+++ b/src/pygeoprocessing/routing/routing.pyx
@@ -676,15 +676,15 @@ def fill_pits(
     """
     # These variables are used to iterate over the DEM using `iterblock`
     # indexes
-    cdef int win_ysize, win_xsize, xoff, yoff
+    cdef unsigned int win_ysize, win_xsize, xoff, yoff
 
     # the _root variables remembers the pixel index where the plateau/pit
     # region was first detected when iterating over the DEM.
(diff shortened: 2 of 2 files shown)


emlys (approved):
This makes sense to me @phargogh! It looks like all of these variables only need to hold positive numbers - none of them store DEM values, right?

Regarding the logging improvements, I wonder if we might also change the log formatter to automatically display the function name? It looks like there is a `funcName` attribute which could do that: https://docs.python.org/3/library/logging.html#logrecord-attributes

phargogh:
I believe yes, I only changed the datatype of values that related to raster row/col indices, stack depth indices, and things like that where we should never have negative values.

> Regarding the logging improvements, I wonder if we might also change the log formatter to automatically display the function name? It looks like there is a funcName attribute which could do that: https://docs.python.org/3/library/logging.html#logrecord-attributes

I totally agree with you, but there's been an issue with cython functions where the function's name isn't reflected in logging that comes from cython, it only displays the module.  So displaying the function name in the message itself gets around that.  It's not ideal, but it works!

phargogh:
@emlys here's a relevant issue from the cython tracker: https://github.com/cython/cython/issues/2203, which refers over to the open issue https://github.com/cython/cython/issues/1373.

PR #353, opened 2023-09-26
Add support for int64, uint64, and int8

### LINKED ISSUE #329: GDAL 3.7 changes behavior for signed bytes
[Per the GDAL docs](https://gdal.org/user/raster_data_model.html#image-structure-domain):

> PIXELTYPE: This may appear on a GDT_Byte band (or the corresponding dataset) and have the value SIGNEDBYTE to indicate the unsigned byte values between 128 and 255 should be interpreted as being values between -128 and -1 for applications that recognise the SIGNEDBYTE type. Starting with GDAL 3.7, this metadata item is no longer used, as the Int8 data type is used for signed bytes.

It is not immediately clear to me if that means that the SIGNEDBYTE flag will be ignored entirely or if we should expect pygeoprocessing to handle both cases.  We should investigate this behavior and make a decision on how to handle this.  Until we do, signed integer raster operations in GDAL 3.7+ may have unexpected behavior.

### LINKED ISSUE #352: Add support for int64 and uint64
GDAL 3.5 added the types `GDT_Int64` and `GDT_UInt64`. Using these types causes a `KeyError` in pygeoprocessing because they're missing from the `_GDAL_TYPE_TO_NUMPY_LOOKUP` dictionary. A user ran into this issue and reported it: https://community.naturalcapitalproject.org/t/keyerror-13-in-pygeoprocessing/4113

I think all we have to do is add the new types to the dictionary.

emlys:
Fixes #352 
Fixes #329 
- Removed our mappings between numpy and gdal dtypes, using the functions gdal provides for this instead. That way we don't have to explicitly include different dtypes depending on the gdal version.
- Bump up the GDAL and python versions that we test in github actions
- Skip a couple of tests related to signed bytes on GDAL>=3.7. Since the new `gdal.GDT_Int8` is just like any other GDAL type, I don't think we need to add tests specifically for it.
- Fixed what seems like a bug in `create_raster_from_base`, where dtype-related metadata (`PIXELTYPE=SIGNEDBYTE`) was copied over from the base raster, even though the target dtype is not copied from the base raster. Now that metadata has to be set in the `raster_creation_tuple`.

## behavior of GDAL 3.7+ with signed bytes

A warning is raised if you create a raster using the deprecated flag:
```
>>> driver = gdal.GetDriverByName('GTIFF')
>>> raster = driver.Create('foo.tif', 1, 1, 1, gdal.GDT_Byte, options=['PIXELTYPE=SIGNEDBYTE'])
Warning 1: Using PIXELTYPE=SIGNEDBYTE with Byte data type is deprecated (but still works). Using Int8 data type instead is now recommended.
```
It seems that drivers will still be able to read signed byte rasters. They just treat them as a new data type (Int8) and don't return the `PIXELTYPE=SIGNEDBYTE` metadata element:
```
>>> band = raster.GetRasterBand(1)
>>> band.DataType
14  # gdal.GDT_Int8
>>> band.GetMetadata('IMAGE_STRUCTURE')
{}
```
From the [migration guide](https://github.com/OSGeo/gdal/blob/534f1e85071e20b042f458c374782842bc3dfc2f/MIGRATION_GUIDE.TXT#L7),
>  Following RFC 87, PIXELTYPE=SIGNEDBYTE in IMAGE_STRUCTURE metadata domain is
  no longer reported by drivers that used to do it. The new GDT_Int8 data type
  is now reported.
  On writing, the PIXELTYPE=SIGNEDBYTE creation option is preserved in drivers
  that used to support it, but is deprecated and external code should rather use
  the GDT_Int8 data type.


pyproject.toml
```diff
@@ -13,8 +13,7 @@ build-backend = "setuptools.build_meta"
 
 [tool.pytest.ini_options]
 # Raise warnings to exceptions.
-filterwarnings = 'error'
-
+filterwarnings = ["error", "default::DeprecationWarning", "default::FutureWarning"]
```
phargogh, line 16: Nice.  This is a great idea and I'm glad they're supporting this in `pyproject.toml` now.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -3703,7 +3665,7 @@ def mask_op(base_array, mask_array):
     os.remove(mask_raster_path)
 
 
-def _gdal_to_numpy_type(band):
+def _gdal_to_numpy_type(gdal_type, metadata):
```
phargogh, line 3668: Since the function interface is changing, could you update the docstring here as well?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -3716,27 +3678,27 @@ def _gdal_to_numpy_type(band):
...
-        gdal.GDT_UInt32: numpy.uint32,
-        gdal.GDT_Float32: numpy.float32,
-        gdal.GDT_Float64: numpy.float64,
-    }
+    if (GDAL_VERSION < (3, 7, 0) and gdal_type == gdal.GDT_Byte and
+        (('PIXELTYPE=SIGNEDBYTE' in metadata) or
+         ('PIXELTYPE' in metadata and metadata['PIXELTYPE'] == 'SIGNEDBYTE'))):
+            return numpy.int8
+
+    numpy_type = gdal_array.GDALTypeCodeToNumericTypeCode(gdal_type)
```
phargogh, line 3687: Well how about that!  This is a nice function to know about.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -3716,27 +3678,27 @@ def _gdal_to_numpy_type(band):
...
+    if numpy_type is None:
+        raise ValueError(f"Unsupported DataType: {gdal_type}")
+    return numpy_type
 
-    if band.DataType in base_gdal_type_to_numpy:
-        return base_gdal_type_to_numpy[band.DataType]
 
-    if band.DataType != gdal.GDT_Byte:
-        raise ValueError("Unsupported DataType: %s" % str(band.DataType))
+def _numpy_to_gdal_type(numpy_type):
```
phargogh, line 3693: Could you add a docstring here that includes the input and output types?  Thanks!

src/pygeoprocessing/symbolic.py
```diff
@@ -105,35 +105,16 @@ def evaluate_raster_calculator_expression(
...
-        # it's a signed byte
-        target_gdal_type = gdal.GDT_Byte
-        target_raster_driver_creation_tuple = (
-            raster_driver_creation_tuple[0],
-            tuple(raster_driver_creation_tuple[1])+('PIXELTYPE=SIGNEDBYTE',))
+    target_gdal_type, type_creation_options = (
+        geoprocessing._numpy_to_gdal_type(target_numpy_type))
+
+    driver, creation_options = raster_driver_creation_tuple
+    creation_options = list(creation_options) + type_creation_options
```
phargogh, line 112: Do we need to handle the case where a creation option is defined both in `creation_options` and in `type_creation_options`?  Will GDAL work as we hope if an option is defined in both?

src/pygeoprocessing/geoprocessing_core.pyx
```diff
@@ -656,22 +656,20 @@ def raster_band_percentile(
...
 
     """
-    raster_type = pygeoprocessing.get_raster_info(
-        base_raster_path_band[0])['datatype']
-    if raster_type in (
-            gdal.GDT_Byte, gdal.GDT_Int16, gdal.GDT_UInt16, gdal.GDT_Int32,
-            gdal.GDT_UInt32):
+    numpy_type = pygeoprocessing.get_raster_info(
+        base_raster_path_band[0])['numpy_type']
+    if numpy.issubdtype(numpy_type, numpy.integer):
```
phargogh, line 661: This is a really nice update - one less thing to have to worry about updating if GDAL adds more int or float types!

phargogh (changes requested):
This is great!  Thanks @emlys .  I just had a couple comments on docstrings in the new functions and a question about how GDAL handles duplicate creation options.

src/pygeoprocessing/symbolic.py
emlys: It seems to work as expected if `PIXELTYPE=SIGNEDBYTE` is defined twice. And it seems that `SIGNEDBYTE` is the only recognized value for `PIXELTYPE`, so it wouldn't overwrite a different option.

src/pygeoprocessing/geoprocessing.py
emlys: Added a docstring!

phargogh (approved):

phargogh:
Looks great! Just waiting for the tests to finish before merging.

PR #356, opened 2023-10-20
Handle nan nodata in calculate_slope

### LINKED ISSUE #352: Add support for int64 and uint64

emlys:
Fixes #352

### DIFF
--- a/src/pygeoprocessing/geoprocessing_core.pyx
+++ b/src/pygeoprocessing/geoprocessing_core.pyx
@@ -403,6 +403,12 @@ def calculate_slope(
             buf_obj=dem_array[y_start:y_end, x_start:x_end],
             **block_offset_copy)
 
+        def _eq(value, nodata):
+            """Compare value against nodata, handling NaN"""
+            if numpy.isnan(nodata):
+                return numpy.isnan(value)
+            return value == nodata
+
         for row_index in range(1, win_ysize+1):
             for col_index in range(1, win_xsize+1):
                 # Notation of the cell below comes from the algorithm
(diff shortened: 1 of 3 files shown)


src/pygeoprocessing/geoprocessing_core.pyx
```diff
@@ -403,6 +403,12 @@ def calculate_slope(
             buf_obj=dem_array[y_start:y_end, x_start:x_end],
             **block_offset_copy)
 
+        def _eq(value, nodata):
+            """Compare value against nodata, handling NaN"""
+            if numpy.isnan(nodata):
+                return numpy.isnan(value)
+            return value == nodata
+
```
phargogh: Refactoring this to a separate function makes a whole lot of sense to me!

Since this function is called 9 times on every pixel value, could I suggest a small refactor?  Statically typing this should yield a big speedup:

```pyrex
cimport libc.math as cmath

# This will need to be a top-level function since it's a `cdef`
# Using doubles because the pixels are already cast to numpy.float64
cdef inline bool _eq(double value, double nodata):
    if cmath.isnan(nodata):
        return cmath.isnan(value)
    return value == nodata
```

With this change Cython will compile this down to pure C, eliminating the need to make multiple python-land function calls to `_eq` and `numpy.isnan` with all the associated overhead and type conversions.

phargogh (changes requested):
Thanks @emlys!  The refactor looks great and I just had one suggestion about an opportunity to speed up the new comparison function.

src/pygeoprocessing/geoprocessing_core.pyx
emlys: Ooh good idea!

phargogh (approved):
🚀

PR #362, opened 2024-01-17
Log or warn if GDAL's cache exceeds SLURM's allocated node memory

### LINKED ISSUE #361: Warn about setting a GDAL cache max if probably on a SLURM node
After some time away from Sherlock, I wanted to run a simple pygeoprocessing function call on a relatively small raster but found myself exhausting memory.  Of course, I had forgotten that I was running this within a SLURM job, which may have some extra identifying features available, such as environment variables.

It would be really great to be able to, at import time, log a warning or something that we are running in a slurm job and that gdal's cache max is unset, defaulting to a very large memory limit.  Even better would be if we could detect the slurm job memory limit and compare it to the GDAL cache limit, but even a simple warning would be great.

phargogh:
This PR introduces a check of environment variables (to determine whether we're operating within a SLURM node) and GDAL's cache max setting, and if the GDAL cache is set to exceed the amount of memory allocated to the SLURM node, a warning is logged or sent through the warning system, depending on whether you're using logging or not.

This check executes at the import of the package, but is available as a function call in a new `slurm_utils` module.

Fixes #361

### DIFF
--- a/src/pygeoprocessing/slurm_utils.py
+++ b/src/pygeoprocessing/slurm_utils.py
@@ -25,8 +25,7 @@ def log_warning_if_gdal_will_exhaust_slurm_memory():
     warning is logged with the logging system.  Otherwise, the warnings system
     is used directly.
     """
-    slurm_env_vars = set(k for k in os.environ.keys() if k.startswith('SLURM'))
-    if slurm_env_vars:
+    if {'SLURM_MEM_PER_NODE'}.issubset(set(os.environ.keys())):
         gdal_cache_size = gdal.GetCacheMax()
         if gdal_cache_size < 100000:
             # If the cache size is 100,000 or greater, it's assumed to be in
@@ -38,7 +37,7 @@ def log_warning_if_gdal_will_exhaust_slurm_memory():
             gdal_cache_size_mb = gdal_cache_size / 1024 / 1024
 
         slurm_mem_per_node = os.environ['SLURM_MEM_PER_NODE']
-        if gdal_cache_size_mb > int(os.environ['SLURM_MEM_PER_NODE']):
+        if gdal_cache_size_mb > int(slurm_mem_per_node):
             message = (
                 "GDAL's cache max exceeds the memory SLURM has "
                 "allocated for this node. The process will probably be "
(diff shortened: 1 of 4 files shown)


src/pygeoprocessing/slurm_utils.py
```diff
@@ -0,0 +1,58 @@
...
+        if gdal_cache_size < 100000:
+            # If the cache size is 100,000 or greater, it's assumed to be in
+            # bytes.  Otherwise, units are interpreted as megabytes.
+            # See gcore/gdalrasterblock.cpp for reference.
+            gdal_cache_size_mb = gdal_cache_size
+        else:
+            # Convert from bytes to megabytes
+            gdal_cache_size_mb = gdal_cache_size / 1024 / 1024
+
+        slurm_mem_per_node = os.environ['SLURM_MEM_PER_NODE']
```
dcdenu4, line 39: If we're in a SLURM system is it guaranteed this environment variable is set? Is it possible we get a KeyError here? Just wondering if we should check for this specific ENV var instead of `if slurm_env_vars`.

src/pygeoprocessing/slurm_utils.py
```diff
@@ -0,0 +1,58 @@
...
+        if gdal_cache_size_mb > int(os.environ['SLURM_MEM_PER_NODE']):
```
dcdenu4: Since we have `slurm_mem_per_node` above, can we use that variable here?

dcdenu4 (changes requested):
Cool, thanks @phargogh! Just a few minor comments.

src/pygeoprocessing/slurm_utils.py
phargogh: > If we're in a SLURM system is it guaranteed this environment variable is set?
I have not been able to find a direct answer to this question, but it is my understanding that a series of `SLURM_*` 
environment variables are expected to be set on every job run.  There also doesn't appear to be a better way to identify if we're in a SLURM job other than to inspect the env vars.

> Is it possible we get a KeyError here?
Definitely seems like a thing we should protect against!  I've patched this in 29346336

src/pygeoprocessing/slurm_utils.py
phargogh: Definitely.  Also patched in 29346336

dcdenu4 (approved):
Looks good, thanks!

PR #367, opened 2024-01-31
Only rasterize once in `align_and_resize_raster_stack`

phargogh:
This PR refactors `align_and_resize_raster_stack` and `warp_raster` to only rasterize a mask vector once.  This results in a large speedup in cases where mask vector geometries are expensive to rasterize, such as when geometries are complex or there are many features in the vector.

I have tried using this approach in a dev build of `natcap.invest` on a big DEM (InVEST SDR, all of Colombia at 30m resolution) and this optimization cuts the total runtime of the model nearly in _half_.

The approach here is:
1. Use a Warped VRT to quickly determine the shape and projection of the mask raster even before the actual warping takes place
2. Convert the warped VRT to a GTiff
3. Rasterize the vector onto the GTiff
4. Pass the GTiff down to each `warp_raster` call
5. `warp_raster` has been modified to accept either a vector (plus parameters) or a pre-masked raster.

### DIFF
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -120,7 +120,8 @@ for _warp_algo in (_attrname for _attrname in dir(gdalconst)
 _GDAL_WARP_ALGORITHMS = set(_GDAL_WARP_ALGORITHMS)
 _GDAL_WARP_ALGOS_FOR_HUMAN_EYES = "|".join(_GDAL_WARP_ALGORITHMS)
 LOGGER.debug(
-    f'Detected warp algorithms: {", ".join(_GDAL_WARP_ALGOS_FOR_HUMAN_EYES)}')
+    'Detected warp algorithms: '
+    f'{_GDAL_WARP_ALGOS_FOR_HUMAN_EYES.replace("|", ", ")}')
 
 
 class TimedLoggingAdapter(logging.LoggerAdapter):
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -2687,11 +2687,9 @@ def warp_raster(
                 _mask_values, target_raster_path,
                 source_raster_info['datatype'], source_nodata)
         else:
-            # If the user did not provide a mask raster, we need to create one.
-            # mask_raster() does the rasterization for us.
-            #
-            # Make sure the raster creation options passed to ``mask_raster``
-            # reflect any metadata updates
+            # If the user did not provide a mask in raster form, we can just
+            # call down to ``mask_raster``, which will rasterize the vector and
+            # then mask out pixels in ``warped_raster_path`` for us.
             updated_raster_driver_creation_tuple = (
                 raster_driver_creation_tuple[0],
                 tuple(raster_creation_options))
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -1172,9 +1172,6 @@ def align_and_resize_raster_stack(
                       vector_mask_options['mask_vector_where_filter']
                       if 'mask_vector_where_filter' in vector_mask_options
                       else None))
-    else:
-        # No masking, set to None
-        mask_raster_path = None
 
     for index, (base_path, target_path, resample_method) in enumerate(zip(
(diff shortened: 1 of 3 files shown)


src/pygeoprocessing/geoprocessing.py
```diff
@@ -120,7 +120,8 @@ def __init__(self, missing_values, raster_path, value_map):
 _GDAL_WARP_ALGORITHMS = set(_GDAL_WARP_ALGORITHMS)
 _GDAL_WARP_ALGOS_FOR_HUMAN_EYES = "|".join(_GDAL_WARP_ALGORITHMS)
 LOGGER.debug(
-    f'Detected warp algorithms: {", ".join(_GDAL_WARP_ALGOS_FOR_HUMAN_EYES)}')
+    'Detected warp algorithms: '
+    f'{_GDAL_WARP_ALGOS_FOR_HUMAN_EYES.replace("|", ", ")}')
```
phargogh, line 125: This fix was not actually part of the feature implementation; it's just been bugging me for a while and was so minor that I snuck it in here :)

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2433,10 +2484,13 @@ def warp_raster(
             in Well Known Text format.
         n_threads (int): optional, if not None this sets the ``N_THREADS``
             option for ``gdal.Warp``.
-        vector_mask_options (dict): optional, if not None, this is a
-            dictionary of options to use an existing vector's geometry to
-            mask out pixels in the target raster that do not overlap the
-            vector's geometry. Keys to this dictionary are:
+        vector_mask_options (dict or str): optional. If None, no masking will
+            be done.  If a string, it must be a path to a raster that is in the
```
davemfish: Small discrepancy here. Later on in this docstring `mask_raster_path` is mentioned as a key in the dict, but here it is suggested to pass directly as a string. I don't think the str option is actually supported.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2609,20 +2671,44 @@ def warp_raster(
...
+                return output
+
+            raster_calculator(
+                [(warped_raster_path, 1),
+                 (vector_mask_options['mask_raster_path'], 1)],
+                _mask_values, target_raster_path,
+                source_raster_info['datatype'], source_nodata)
+        else:
+            # If the user did not provide a mask raster, we need to create one.
+            # mask_raster() does the rasterization for us.
```
davemfish: I'm a little confused by this comment, maybe it's a leftover? If we're here then we're masking the warped target by a vector, right? No need to create a mask raster first, just use `mask_raster` as you have done?

davemfish (changes requested):
Hey @phargogh , this looks great. I just a couple minor comments about comments & docstrings. 

And I wonder if you think a future API change might be warranted where we rename `warp_raster`'s `vector_mask_options` to `mask_options`?

src/pygeoprocessing/geoprocessing.py
phargogh: Whoops! Yes, this is left over from an earlier draft and patched in 52524630

src/pygeoprocessing/geoprocessing.py
phargogh: This comment could be improved, so I tried to do so in e5c8557f.

If we've gotten to this point, the user provided a mask vector, which means that we need to 1) create a new raster from base for the rasterization, 2) rasterize the vector and 3) use the rasterized vector mask to determine which pixels are masked out.  `mask_raster` handles all 3 of these steps for us.

Does that help clarify things?

phargogh:
Thanks @davemfish ! 

> And I wonder if you think a future API change might be warranted where we rename warp_raster's vector_mask_options to mask_options?

I think this is an awesome idea so I went ahead and implemented it.  So if `vector_mask_options` is provided, `DeprecationWarning` is raised and the variable is renamed to `mask_options` internally.  That way when we decide to actually make the API change (pygeoprocessing 3.0 maybe?  maybe sooner?), we just have to quickly update things and not rename a pile of variables.   c55e469f

davemfish (approved):
Thanks for these updates @phargogh . Great idea to change the API now rather than later and issue the deprecation warning. Thanks for doing that!

PR #369, opened 2024-02-06
Fix minor logging issue in `zonal_statistics`

### LINKED ISSUE #368: Logging for `zonal_statistics` slightly incorrect
I have a case where I'm running `pygeoprocessing.zonal_statistics` with:
* A stack of pre-aligned rasters
* Non-overlapping geometries, indicated to `zonal_stats` through `polygons_might_overlap=False`

I can confirm that there's only one rasterization happening, which is great!  But there's also some logging that makes it seem like there's going to be more processing involved:

```
02/01/2024 19:55:41  (pygeoprocessing.geoprocessing) geoprocessing.logger_callback(3929) INFO rasterizing disjoint polygon set 1 of 40 set 1.0% complete ()
```

It turns out that there's never a set 2 of 40, so I think the logging is just using the wrong variable name.

phargogh:
This PR fixes a minor bug in the logging of `zonal_statistics` where we were displaying an incorrect count of the number of disjoint polygon sets to rasterize.

I also threw in a couple linting fixes about line length and added the function name to a logging message.

Fixes #368

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -20,6 +20,9 @@ Unreleased Changes
   provided instead of a vector.  If both are provided, the mask raster alone is
   used.  The new mask raster must have the same dimensions and geotransform as
   the output warped raster. https://github.com/natcap/pygeoprocessing/issues/366
+* Fixed a bug in ``zonal_statistics`` where the wrong number of disjoint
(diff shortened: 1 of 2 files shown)


emlys (approved):
Sorry for the delay! Looks good

PR #376, opened 2024-03-04
`get_gis_type` remote file access

### LINKED ISSUE #375: `get_gis_type` should work on remote datasets
GDAL can open a connection to a remote file given a URL the same as for a local file, and easily retrieve metadata as from `gdalinfo`.

In `pygeoprocessing`, `get_raster_info` and `get_vector_info` take advantage of this, but `get_gis_type` does not, simply because it checks if the file exists on the local filesystem before trying to open it.

davemfish:
Fixes #375 

The cost of this change is a slightly less informative error message when a file does not exist. Instead of telling the user,
`"{path} does not exist"`, we tell them `"{path} could not be opened as a GDAL raster or vector"`.

That's somewhat unsatisfying, but as it is the much more commonly used `get_raster_info` and `get_vector_info` already give the `"Could not open as a gdal.OF_RASTER"` message when the file does not exist.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,10 +3,14 @@ Release History
 
 Unreleased Changes
 ------------------
+* ``get_gis_type`` can accept a path to a remote file, allowing the GDAL driver
+  to open it if the driver supports the protocol.
+  https://github.com/natcap/pygeoprocessing/issues/375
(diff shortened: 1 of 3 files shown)


phargogh (approved):
Sounds great to me!

PR #388, opened 2024-04-09
proposal/bugfix: Remove checks for gdal resampling algorithm values

### LINKED ISSUE #387: GDAL warp algorithm names missing with GDAL>=3.8.5
GDAL 3.8.5 [fixed a bug](https://github.com/OSGeo/gdal/pull/9260) where the options list provided to `gdal.WarpOptions` was modified by the function call. pygeoprocessing was relying on this bug to get the warp algorithm name. We'll need to find another way to get `_GDAL_WARP_ALGORITHMS`.
https://github.com/natcap/pygeoprocessing/blob/83e3ffd37a5489482b475a2079b1c53ee5262d3a/src/pygeoprocessing/geoprocessing.py#L79-L119

emlys:
Fixes #387 

This was pretty simple so I went ahead with it, but happy to consider alternatives too!

We were relying on a gdal bug to build a list of human-readable names for resampling algorithms. In gdal 3.8.5, this no longer works. We could implement a different way to get the names, but I propose that we remove the list entirely.

The lists in question (`_GDAL_WARP_ALGORITHMS` and `_GDAL_WARP_ALGOS_FOR_HUMAN_EYES`) are used in `warp_raster` and `build_overviews`. They're only used to check that the provided resampling algorithm is a valid option, and raise a helpful error message if not. I think we can do away with that because GDAL raises its own (slightly less helpful) error message if given an invalid algorithm name:

`gdal.WarpRaster` raises a `TypeError: in method 'wrapper_GDALWarpDestName', argument 4 of type 'GDALWarpAppOptions *'` and prints out `ERROR 5: Unknown resampling method: <resampling method>.` With `gdal.UseExceptions()`, you get a nicer `RuntimeError: Unknown resampling method: <resampling method>.`

`gdal.BuildOverviews` raises a `RuntimeError: Building overviews failed or was interrupted for <raster path>.` and prints out `ERROR 1: GDALGetResampleFunction: Unsupported resampling method "<resampling_method>".` With `gdal.UseExceptions()`, you get `RuntimeError: GDALGetResampleFunction: Unsupported resampling method "<resampling method>"`.

These error messages aren't quite as nice as what `pygeoprocessing` had, but they're adequate to identify the issue. Since we pass the resample method argument directly to the underlying gdal function, I think it makes sense to let GDAL handle it.

A couple of behavior changes:
- The `gdal.WarpRaster` resample algorithm argument can be a `gdal.GRA_*` constant as well as a string name. By removing the pygeoprocessing check, we can accept `gdal.GRA_*` constants too.
- `gdal.WarpRaster` and `gdal.BuildOverviews` use the nearest neighbor algorithm if the resampling argument _starts with_ 'near'. So 'near', 'nearest', 'nearest_neighbor', and 'near_foo' all work.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,6 +6,10 @@ Unreleased Changes
 * Our github actions for building python distributions now use
   `actions/setup-python@v5`, which uses node 20.
   https://github.com/natcap/pygeoprocessing/issues/384
+* ``warp_raster`` and ``build_overviews`` no longer raise ``ValueError``s if
+  called with an invalid resampling algorithm. We now fall back to the
+  underlying GDAL functions' error messages.
+  https://github.com/natcap/pygeoprocessing/issues/387
 
 2.4.3 (2024-03-06)
 ------------------
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -76,55 +76,6 @@ _VALID_GDAL_TYPES = (
 _LOGGING_PERIOD = 5.0  # min 5.0 seconds per update log message for the module
 _LARGEST_ITERBLOCK = 2**16  # largest block for iterblocks to read in cells
 
-_GDAL_WARP_ALGORITHMS = []
-for _warp_algo in (_attrname for _attrname in dir(gdalconst)
-                   if _attrname.startswith('GRA_')):
-    # In GDAL < 3.4, the options used were actually gdal.GRIORA_*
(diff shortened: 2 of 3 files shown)


phargogh (approved):
Yeah, this makes sense as a solution.  In GDAL 4.0 the switch to exceptions by default will make it more readable anyways, and this solution is a more general one since, as you point out, allows GDAL to decide how to interpret a warp mode and whether it is invalid.  Thanks @emlys !

Just waiting to confirm that the tests pass before merging.

PR #389, opened 2024-04-10
Implement decaying D8 flow accumulation

### LINKED ISSUE #386: Contribute Pollutant Decay to Flow Accumulation
For the NCI Water Quality work, I implemented a modification to D8 Flow Accumulation that allowed for the on-pixel load (`1` per pixel, or else the pixel value of a weights raster, if used) to decay by some known proportion as it traveled downstream.  Rafa said there could be some applications in InVEST for modeling pollutant decay in this way.

For example, suppose we have the following flowpath, weights and decay factors:

```
Flowpath: a --> b --> c --> d
Weights:  1     2     3     4
Decay:    0.5   0.75  0.3   0.1
```

On pixel `a`, the flow accumulation would be `1`.

On pixel `b`, the flow accumulation would be the decayed contribution from upstream, plus the on-pixel load from `b`, `1*0.5**1 + 2*0.75**0`.

On pixel `c`, the flow accumulation would follow the same pattern, `1*0.5**2 + 2*0.75**1 + 3*0.3**0`

Pixel `d` would again follow the same pattern: `1*0.5**3 + 2*0.75**2 + 3*0.3**1 + 4*0.1**0`.

That's the idea anyways, at least for a simple flowpath.  For real flowpaths, the best way to do this is actually to track the upstream contributions so that we can sum them correctly in cases of convergent flow and non-linear flowpaths.

phargogh:
This PR copies the modifications to D8 flow accumulation implemented for the NCI Water Quality project that allow for pixel loads to decay as they pass over the landscape.  You could think of like a pollutant flowing across the landscape, where some fraction is absorbed due to landcover or some other factor.

Tests have been implemented and runtime should only trivially be affected when decaying accumulation is not in use.  When decaying accumulation is used, there is an extra cost in memory and in iteration over the pollutant contributions to the pixel.

Fixes #386

### DIFF
--- a/src/pygeoprocessing/routing/routing.pyx
+++ b/src/pygeoprocessing/routing/routing.pyx
@@ -138,6 +138,18 @@ cdef struct FlowPixelType:
     int last_flow_dir
     double value
 
+cdef struct DecayingValue:
+    double decayed_value
+    double min_value
+
+cdef struct WeightedFlowPixelType:
+    int xi
+    int yi
+    int last_flow_dir
+    double value
+    queue[DecayingValue] decaying_values
+
+
 # this struct is used in distance_to_channel_mfd to add up each pixel's
 # weighted distances and flow weights
 cdef struct MFDFlowPixelType:
@@ -1442,7 +1454,8 @@ def flow_dir_d8(
 
 def flow_accumulation_d8(
         flow_dir_raster_path_band, target_flow_accum_raster_path,
-        weight_raster_path_band=None,
+        weight_raster_path_band=None, custom_decay_factor=None,
+        double min_decay_proportion=0.001,
         raster_driver_creation_tuple=DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS):
     """D8 flow accumulation.
 
@@ -1468,6 +1481,14 @@ def flow_accumulation_d8(
             weight. If ``None``, 1 is the default flow accumulation weight.
             This raster must be the same dimensions as
             ``flow_dir_mfd_raster_path_band``.
+        custom_decay_factor=None (float or str): a custom decay factor, either
+            represented as a float (where the same decay factor is applied to
+            all valid pixels) or a raster where pixel values represent
+            spatially-explicit decay values.
+        min_decay_proportion=0.001 (float): A value representing the minimum
+            decayed value that should continue to be tracked along the flow
+            path when using a custom decay factor.  If the upstream decayed
+            contribution falls below this value, it is not included.
         raster_driver_creation_tuple (tuple): a tuple containing a GDAL driver
             name string as the first element and a GDAL creation options
             tuple/list as the second. Defaults to a GTiff driver tuple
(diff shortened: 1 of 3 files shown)


src/pygeoprocessing/routing/routing.pyx
```diff
@@ -138,6 +138,18 @@ cdef struct FlowPixelType:
     int last_flow_dir
     double value
 
+cdef struct DecayingValue:
+    double decayed_value  # The value, which will be progressively updated as it decays
+    double min_value  # The minimum value before the Decaying Value should be ignored.
+
+cdef struct WeightedFlowPixelType:
```
dcdenu4: I noticed the other `struct` definitions have a few comments above to describe them. Should these follow that same convention?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -138,6 +138,18 @@ cdef struct FlowPixelType:
...
     double value
 
+cdef struct DecayingValue:
+    double decayed_value  # The value, which will be progressively updated as it decays
+    double min_value  # The minimum value before the Decaying Value should be ignored.
+

+cdef struct WeightedFlowPixelType:
+    int xi
+    int yi
```
dcdenu4: The other `struct` definitions use `unsigned int` for pixel location, any reason not to do that here as well for consistency?

src/pygeoprocessing/routing/routing.pyx
```diff
@@ -1593,7 +1643,9 @@ def flow_accumulation_d8(
         xi_n = -1
         yi_n = -1
 
-        # search block for to set flow direction
+        # Search the block for root pixels.
+        # A root pixel is a flow direction pixel that is nodata, which means it
+        # may be a drain just off the edge.
```
dcdenu4, line 1654: Thank you!

src/pygeoprocessing/routing/routing.pyx
dcdenu4, line 152: I think that this replaces ALL instances of `FlowPixelType`. Should we remove that struct definition?

dcdenu4 (changes requested):
Hey @phargogh, sorry for the LONG delay here. This looks solid to me, I just had a few questions related to consistency.

src/pygeoprocessing/routing/routing.pyx
phargogh: Yep, great catch!  Patched in 4f0e062d

phargogh: Interestingly, I'm still seeing `FlowPixelType` used in `flow_accumulation_mfd`, so I think it makes sense to leave it here for now.

src/pygeoprocessing/routing/routing.pyx
phargogh: Nope!  In fact they really should be `unsigned int`s for consistency with the way indexes are used elsewhere in the function.  Patched in d645b70a

dcdenu4 (approved):
Thanks!

PR #390, opened 2024-05-07
update to cython 3

emlys:
Update to Cython 3. Also fixes the github actions workflow for the new mac runners.

### DIFF
--- a/.github/workflows/build-py-dists.yml
+++ b/.github/workflows/build-py-dists.yml
@@ -9,7 +9,6 @@ jobs:
             matrix:
                 os: [windows-latest, macos-latest]
                 python-version: [3.7, 3.8, 3.9, "3.10", "3.11", "3.12"]
-                python-arch: [x64]
 
         steps:
             - uses: actions/checkout@v4
@@ -20,11 +19,10 @@ jobs:
             - name: Fetch git tags
               run: git fetch origin +refs/tags/*:refs/tags/*
 
-            - name: Set up python ${{ matrix.python-version }} ${{ matrix.python-arch }}
+            - name: Set up python ${{ matrix.python-version }}
               uses: actions/setup-python@v5
               with:
                   python-version: ${{ matrix.python-version }}
-                  architecture: ${{ matrix.python-arch }}
 
             - name: Install dependencies
               run: python -m pip install build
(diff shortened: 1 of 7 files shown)


.github/workflows/build-py-dists.yml
```diff
@@ -8,8 +8,7 @@ jobs:
             fail-fast: false
             matrix:
                 os: [windows-latest, macos-latest]
-                python-version: [3.7, 3.8, 3.9, "3.10", "3.11", "3.12"]
-                python-arch: [x64]
```
emlys, line 12: `python-arch` isn't needed here, and it caused an issue with the new ARM64 mac runners.

.github/workflows/build-py-dists.yml
```diff
@@ -8,8 +8,7 @@ jobs:
             fail-fast: false
             matrix:
                 os: [windows-latest, macos-latest]
-                python-version: [3.7, 3.8, 3.9, "3.10", "3.11", "3.12"]
```
emlys, line 11: python 3.7 is end-of-life, we already removed it from the test workflow. And `setup-python` couldn't find an ARM64 python 3.7 distribution.

.github/workflows/build-py-dists.yml
phargogh: I bet this is a holdover from when we were still doing x86 builds.  Makes sense to remove!

HISTORY.rst
```diff
@@ -10,6 +10,7 @@ Unreleased Changes
   called with an invalid resampling algorithm. We now fall back to the
   underlying GDAL functions' error messages.
   https://github.com/natcap/pygeoprocessing/issues/387
+* Updated to Cython 3.
```
phargogh, line 13: If we are removing testing/wheel-building support for python 3.7, should that also be noted here in HISTORY and also removed from the `setup.py` trove classifiers?

phargogh (changes requested):
Looks great to me, I just had one comment about dropping explicit support for python 3.7.

HISTORY.rst
emlys: Ooh you're right

phargogh (approved):
Code changes look good to me, thanks @emlys !  Just waiting for the tests to pass.

PR #392, opened 2024-05-09
enable gdal.UseExceptions throughout pygeoprocessing

### LINKED ISSUE #391: Enable `gdal.UseExceptions()`
See discussion: https://github.com/natcap/invest/discussions/1299

emlys:
Fixes #391 

This PR adds a context manager called `GDALUseExceptions` that turns on `gdal.UseExceptions` within the context and restores the original state afterward. Also adds a decorator that applies the context manager to a function.

I first tried to use the context manager GDAL provides, `gdal.ExceptionMgr`, but it didn't do what I needed. It doesn't silence the `FutureWarning`, and I noticed it behaved weirdly and didn't seem to always restore state correctly.

I applied either the decorator or the context manager everywhere that we use GDAL, to each function individually so that our `gdal.UseExceptions` state doesn't affect, and isn't affected by, other code. Since we aren't importing any libraries that might also change the state, I think it's guaranteed that exceptions will be enabled where we expect. I used the decorator as much as possible, and the context manager in a few places where cython or multiprocessing complained about the decorator.

I removed some error handling that's made redundant by GDAL's exceptions. We have some tests that tested that error handling, and I thought about removing those tests, but decided to update them to test the GDAL exceptions instead. Just to make sure those show up right.

Also replaced all `assertTrue(x in y)` with `assertIn(x, y)` because they were driving me crazy 🙃

### DIFF
--- a/src/pygeoprocessing/geoprocessing_core.pyx
+++ b/src/pygeoprocessing/geoprocessing_core.pyx
@@ -46,7 +46,8 @@ DEFAULT_OSR_AXIS_MAPPING_STRATEGY = osr.OAMS_TRADITIONAL_GIS_ORDER
 LOGGER = logging.getLogger('pygeoprocessing.geoprocessing_core')
 
 
-class GDALUseExceptions(object):
+class GDALUseExceptions:
+    """Context manager that enables GDAL exceptions and restores state after."""
 
     def __init__(self):
         pass
@@ -55,12 +56,20 @@ class GDALUseExceptions(object):
         self.currentUseExceptions = gdal.GetUseExceptions()
         gdal.UseExceptions()
 
-    def __exit__(self, exc_type, exc_val, exc_tb):
+    def __exit__(self, _, _, _):
         if self.currentUseExceptions == 0:
             gdal.DontUseExceptions()
 
 
 def gdal_use_exceptions(func):
+    """Decorator that enables GDAL exceptions and restores state after.
+
+    Args:
+        func (callable): function to call with GDAL exceptions enabled
+
+    Returns:
+        Wrapper function that calls ``func`` with GDAL exceptions enabled
+    """
     def wrapper(*args, **kwargs):
         with GDALUseExceptions():
             return func(*args, **kwargs)
--- a/src/pygeoprocessing/geoprocessing_core.pyx
+++ b/src/pygeoprocessing/geoprocessing_core.pyx
@@ -56,7 +56,7 @@ class GDALUseExceptions:
         self.currentUseExceptions = gdal.GetUseExceptions()
         gdal.UseExceptions()
 
-    def __exit__(self, _, _, _):
+    def __exit__(self, exc_type, exc_val, exc_tb):
         if self.currentUseExceptions == 0:
             gdal.DontUseExceptions()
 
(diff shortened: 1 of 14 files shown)


src/pygeoprocessing/symbolic.py
```diff
@@ -5,7 +5,6 @@
 
 from . import geoprocessing
 from .geoprocessing import DEFAULT_GTIFF_CREATION_TUPLE_OPTIONS
-from osgeo import gdal
```
emlys, line 8: Unused import

src/pygeoprocessing/routing/watershed.pyx
```diff
@@ -470,100 +473,105 @@ cdef cset[CoordinatePair] _c_split_geometry_into_seeds(
...
+        local_geotransform = [
+            local_origin_x, flow_dir_geotransform[1], flow_dir_geotransform[2],
+            local_origin_y, flow_dir_geotransform[4], flow_dir_geotransform[5]]
+        gtiff_driver = gdal.GetDriverByName('GTiff')
+        # Raster is sparse, no need to fill.
+        raster = gtiff_driver.Create(
+            target_raster_path, int(local_n_cols), int(local_n_rows), 1,
+            gdal.GDT_Byte, options=GTIFF_CREATION_OPTIONS)
+
+        raster.SetSpatialRef(flow_dir_srs)
```
emlys, line 516: It seems like `SetSpatialRef` is equivalent to `SetProjection` but takes in a spatial reference object rather than a WKT string. Calling `ExportToWKT` on an empty spatial reference raises an unhelpful error, so I changed this to avoid having to get the WKT string at all.

src/pygeoprocessing/geoprocessing_core.pyx
```diff
@@ -45,6 +45,28 @@ DEFAULT_OSR_AXIS_MAPPING_STRATEGY = osr.OAMS_TRADITIONAL_GIS_ORDER
...
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if self.currentUseExceptions == 0:
+            gdal.DontUseExceptions()
+
+
+def gdal_use_exceptions(func):
+    def wrapper(*args, **kwargs):
+        with GDALUseExceptions():
+            return func(*args, **kwargs)
+    return wrapper
```
phargogh: Since these two callables are at the top level, would it be worth adding a docstring to indicate their purpose?

phargogh (changes requested):
This looks great, @emlys thanks so much for going through everything so carefully!

My one request here is that we add docstrings to the decorator and the context manager so that they can be referred to more easily in the API documentation.

And I had one additional question that didn't fit in the line-by-line review: do we need to change anything with our minimum GDAL version stated in `requirements.txt`?  All the tests are passing so we should at least be good to GDAL 3.2.2, but I just wanted to make sure that our current minimum GDAL requirement holds true.

Thanks!

emlys:
@phargogh I added those docstrings, and I tested locally with gdal 3.0.4, our minimum supported version. All tests passed but it's worth noting that GDAL's error messages vary across versions and operating systems, hence some of the tests allowing multiple error types/messages, but they seem to be generally improving. So users of the oldest GDAL versions might encounter less helpful errors.

emlys:
I also had to remove a deprecated import and pin numpy due to some new dependency releases.

phargogh (commented):
Thanks @emlys !  Everything looks great here w/r/t `gdal.UseExceptions`.  I just remembered: since we are changing behavior here (changes in exceptions raised especially), should there be a note about it in HISTORY?  Aside from that, I think this is ready to be merged!

emlys:
@phargogh I added a note in history and also removed the numpy < 2 pin, since your PR is now merged

phargogh (approved):

PR #394, opened 2024-05-21
fix RST syntax error

emlys:


HISTORY.rst
```diff
@@ -6,7 +6,7 @@ Release History
 * Our github actions for building python distributions now use
   `actions/setup-python@v5`, which uses node 20.
   https://github.com/natcap/pygeoprocessing/issues/384
-* ``warp_raster`` and ``build_overviews`` no longer raise ``ValueError``s if
+* ``warp_raster`` and ``build_overviews`` no longer raise a ``ValueError`` if
```
emlys, line 9: There is a way to keep the `s` by escaping whitespace after the backticks: "\``ValueError``\ s". But I didn't like how that is less human-readable.

phargogh (approved):
Yep, make sense to me!

PR #397, opened 2024-06-18
Minor updates for numpy 2.0 compatibility

### LINKED ISSUE #396: Ensure compatibility with numpy 2.0
Numpy 2.0 was just released, and when I install the latest version (`numpy=2.0.0`) into an environment and run `python -m pytest tests`, I'm getting a few failures on my mac:

```
FAILED tests/test_geoprocessing.py::TestGeoprocessing::test_raster_map_nodata - TypeError: can_cast() does not support Python ints, floats, and complex because the result used to depend on the value.
FAILED tests/test_geoprocessing.py::TestGeoprocessing::test_raster_map_nodata_and_dtype - TypeError: can_cast() does not support Python ints, floats, and complex because the result used to depend on the value.
FAILED tests/test_geoprocessing.py::TestGeoprocessing::test_raster_map_nodata_propagates - TypeError: can_cast() does not support Python ints, floats, and complex because the result used to depend on the value.
FAILED tests/test_geoprocessing.py::TestGeoprocessing::test_reclassify_raster_reclass_max_nodata - TypeError: in method 'Band_SetNoDataValue', argument 2 of type 'double'
```

Since this is a major api change, maybe we should expand our test matrix to test against numpy 2.0 if we aren't already?

phargogh:
This PR adds a few minor updates for numpy 2.0 API compatibility.  Pygeoprocessing was almost entirely compatible already, there were just a few minor changes to the compilation and to the behavior of a couple of functions.

Fixes #396

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -58,7 +58,7 @@ jobs:
               python-build
               flake8
               pytest
-              numpy<2
+              numpy<2  # temporarily capping numpy<2 until we figure out build issues with gdal>=3.8
           environment-name: pyenv
 
       - name: Lint with flake8
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -71,7 +71,7 @@ jobs:
 
       - name: Build and Install PyGeoprocessing
         run: |
-            python -m build --wheel --verbose
+            python -m build --wheel --verbose --no-isolation
             python -m pip install $(find dist -name "*.whl")
 
       - name: Test with pytest
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -71,7 +71,7 @@ jobs:
 
       - name: Build and Install PyGeoprocessing
         run: |
-            python -m build --wheel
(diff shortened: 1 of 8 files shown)


.github/workflows/pythonpackage.yml
```diff
@@ -58,6 +58,7 @@ jobs:
               python-build
               flake8
               pytest
+              numpy<2
```
emlys, line 62: Do we still want this pinned?

phargogh: Unfortunately, the combination of `numpy<2, gdal>=3.8` appears necessary with the latest changes to conda-forge builds.  I have been trying to figure out the right way around this, but it is elusive.  So, as a workaround, I have added `numpy<2` to the requirements here.

emlys: Ok! I would just suggest adding a comment explaining that so we know if we can remove it in the future

phargogh: Great idea!  I added a brief comment to cf711f08, but please let me know if you'd like some additional detail!  I also created an issue for this over at https://github.com/natcap/pygeoprocessing/issues/399

emlys (approved):

PR #401, opened 2024-06-27
Fix compilation/testing issues against numpy 2.0

phargogh:
This PR fixes an issue with our build dependencies.  We used to have to use `oldest-supported-numpy` when building to ensure compatibility with older versions of numpy.  It now appears that building against `numpy>=2` is sufficient for allowing pygeoprocessing to run against earlier versions of numpy.

The only catch appears to be that numpy 2 was only released for python>=3.9, so we are still building against `oldest-supported-numpy` for python 3.8 builds.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -51,7 +51,6 @@ jobs:
           condarc: |
               channels:
                 - conda-forge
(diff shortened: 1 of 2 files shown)


emlys (approved):
Thanks @phargogh!

PR #402, opened 2024-06-28
use functools.wraps with gdal exceptions decorator

emlys:
Thanks for the suggestion @phargogh! That was an easy fix. I confirmed that the wrapped functions now have docstrings and can be pickled.


phargogh (approved):
Easy change and easy review!  I'm glad it's working as we hoped :)

PR #405, opened 2024-10-09
Remove the numpy version cap

### LINKED ISSUE #403: Do the pygeoprocessing 2.4.5 release for numpy 2.0 compatibility
Numpy 2.0 compatibility is a big-enough thing that it makes sense to just release on its own.  So let's ship it!

phargogh:
This should have been included in the 2.4.5 release (#404 ), but I forgot about it.  Nothing should break with the 2.4.5 release, it's just that the next release will allow numpy 2 to be installed at pip-install time.

Fixes #403

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,7 +1,10 @@
 Release History
 ===============
 
-.. Unreleased Changes
+Unreleased Changes
+------------------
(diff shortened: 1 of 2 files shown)


dcdenu4 (approved):
Thanks @phargogh, sorry I missed this as well!

PR #406, opened 2024-10-09
Improve exception on `convolve_2d` worker timeout

phargogh:
This PR makes 2 concrete improvements to `convolve_2d`:

1. Instead of raising `queue.Empty` on timeout, `RuntimeError` with a more helpful message is raised
2. We were not using the function signature's `max_timeout` parameter, so we are now using it.

Also, I shortened the text we're asserting in a couple other tests because recent versions of GDAL made some adjustments to the text of their exceptions.  Shortening the string we're asserting allows us to have the test remain valid for all the GDAL versions we're checking.

Fixes #360

PR #408, opened 2024-10-10
Handle `raster_map` case with float32 array and float64 nodata values

### LINKED ISSUE #358: `raster_map` errors on valid combinations of dtype and nodata value
If `target_dtype` and `target_nodata` are provided to `raster_map`, it tries to confirm that the nodata value fits into the dtype using `numpy.can_cast`. This does not work as I expected:
```
>>> float32_min = float(numpy.finfo(numpy.float32).min)
>>> numpy.can_cast(float32_min, numpy.float32)
False
```
It's usually not necessary to set the `target_dtype` and `target_nodata`, but this bug prevents us from doing so in some cases.

phargogh:
This PR fixes #358, which describes a type-casting issue between several widths of the same type (e.g. float32 vs. float64).  I think @dcdenu4 's suggestion of just adding a kwarg to `numpy.can_cast` is a sufficient solution for what we're trying to do.

This PR's tests will fail until #406 is merged, which is why I'm leaving it as a draft for now.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,15 +1,16 @@
 Release History
 ===============
 
-.. Unreleased Changes
+Unreleased Changes
+------------------
+* Handling a case in ``raster_map`` where an exception would be raised when a
+  float32 array was passed along with a float64 nodata value.
+  https://github.com/natcap/pygeoprocessing/issues/358
 
 2.4.5 (2024-10-08)
 ------------------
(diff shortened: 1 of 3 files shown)


emlys (approved):
Thanks @phargogh, and sorry for the delay!

phargogh:
Looks like tests are passing now after resolving the merge conflict and @emlys approved, so I'll go ahead and merge.  Thanks, all!

PR #416, opened 2024-10-28
test and build against Python 3.13 and GDAL 3.9

### LINKED ISSUE #415: Support Python 3.13 and GDAL 3.9
Build and test against these versions

davemfish:
For the list of test exclusions, the ones I added are the cases where `libmamba` failed to solve for an environment with those combinations. And the ones removed are based on our decision to refer to the Debian Stable package list for the minimum version to test against.

In addition, an exception message appears to have changed slightly in GDAL, so I updated a test.

Fixes #415


emlys (approved):
Thanks @davemfish!

PR #420, opened 2024-12-16
Check for non-numeric keys in value_map in reclassify raster function #419

### LINKED ISSUE #419: Provide better error handling for NA/NaN keys in value map in reclassify raster function
Currently `reclassify_raster` will fail if there is a `NA` key in the value map. This is because we are using `np.isclose` to check if any of the `value_map` keys are close to the `nodata` value, and `isclose` cannot handle `NA` values. 

The current error message produced will be either `TypeError: boolean value of NA is ambiguous` (if trying to check `pandas.NA` against another value, as occurs when value map is created from a csv with an empty row value), or `TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''` (if comparing a value (including `NA`) to `NA`; possibly in other situations).

claire-simpson:
Check that the `value_map` does not contain any non-numeric keys in the raster reclassify function. This function will fail if there are any None/NA keys in the `value_map` dictionary so instead we're explicitly checking for them upfront to provide the user with a clearer error message. Note that this still allows `numpy.nan` keys to pass. I've also created a test to confirm this check works as anticipated.

Fixes #419

### DIFF
--- a/tests/test_geoprocessing.py
+++ b/tests/test_geoprocessing.py
@@ -177,7 +177,6 @@ class TestGeoprocessing(unittest.TestCase):
         value_map = {
             test_value: 100,
         }
-        target_nodata = -1
         with self.assertRaises(
                 pygeoprocessing.ReclassificationMissingValuesError) as cm:
             pygeoprocessing.reclassify_raster(
@@ -222,7 +221,6 @@ class TestGeoprocessing(unittest.TestCase):
         value_map = {
             test_value: 100,
         }
-        target_nodata = -1
         pygeoprocessing.reclassify_raster(
             (raster_path, 1), value_map, target_path, gdal.GDT_Float32,
             target_nodata, values_required=True)
@@ -245,7 +243,6 @@ class TestGeoprocessing(unittest.TestCase):
         value_map = {
             test_value: 100,
         }
-        target_nodata = -1
         # we expect a value error because we didn't pass a (path, band)
         # for the first argument
         with self.assertRaises(ValueError):
@@ -267,7 +264,6 @@ class TestGeoprocessing(unittest.TestCase):
 
         empty_value_map = {
         }
-        target_nodata = -1
         with self.assertRaises(ValueError):
             pygeoprocessing.reclassify_raster(
                 (raster_path, 1), empty_value_map, target_path,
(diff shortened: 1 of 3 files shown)


tests/test_geoprocessing.py
```diff
@@ -187,6 +187,27 @@ def test_reclassify_raster_missing_pixel_value(self):
...
+        n_pixels = 9
+        pixel_matrix = numpy.ones((n_pixels, n_pixels), numpy.float32)
+        target_nodata = -1
+        raster_path = os.path.join(self.workspace_dir, 'raster.tif')
+        target_path = os.path.join(self.workspace_dir, 'target.tif')
+        _array_to_raster(
+            pixel_matrix, target_nodata, raster_path)
+
+        value_map = {1: 2, None: 3, "s": 4, numpy.nan: 5, numpy.float32(99): 6}
+        target_nodata = -1
```
davemfish: `target_nodata` is defined twice

davemfish (changes requested):
Thanks, @claire-simpson , this is very succinct and effective. Would you please add a note in HISTORY about the new `TypeError` being raised? And I found one little redundant line in the test.

claire-simpson:
Thanks @davemfish! I noticed that many of the geoprocessing tests define the ``target_nodata`` two times, may be something to look into?

davemfish:
> Thanks @davemfish! I noticed that many of the geoprocessing tests define the `target_nodata` two times, may be something to look into?

Interesting, feel free to clean that up for us if you'd like to! It could happen in this PR if you want, or a separate one.

claire-simpson:
> Interesting, feel free to clean that up for us if you'd like to! It could happen in this PR if you want, or a separate one.

Went ahead and did it. I suspect target_nodata was redefined explicitly as the function called in between the duplicate assignments did set target_nodata but that should be out of scope and therefore unnecessary to reassign target_nodata in the test script

davemfish:
> Went ahead and did it. I suspect target_nodata was redefined explicitly as the function called in between the duplicate assignments did set target_nodata but that should be out of scope and therefore unnecessary to reassign target_nodata in the test script

Thank you!

davemfish (approved):

Issue #421, opened 2025-01-07
`warp_raster` doesn't raise a helpful error if the source raster path isn't a string


phargogh:
It is trivially easy to accidentally pass in a tuple as the source raster:

```python 
>>> '/vsicurl/my/raster/path.tif',
('/vsicurl/my/raster/path.tif',)
```

And when this happens, GDAL throws a cryptic error that would be near-trivial to avoid if we just did a little checking up front:

```python
Traceback (most recent call last):
  File "/home/users/jadoug06/workspace/2024-12-19-build-esa-overviews/test-clip-esa2021.py", line 22, in <module>
    pygeoprocessing.warp_raster(
  File "src/pygeoprocessing/geoprocessing_core.pyx", line 77, in pygeoprocessing.geoprocessing_core.gdal_use_exceptions.wrapper
  File "src/pygeoprocessing/geoprocessing_core.pyx", line 78, in pygeoprocessing.geoprocessing_core.gdal_use_exceptions.wrapper
  File "/opt/conda/lib/python3.12/site-packages/pygeoprocessing/geoprocessing.py", line 2607, in warp_raster
    gdal.Warp(
  File "/opt/conda/lib/python3.12/site-packages/osgeo/gdal.py", line 790, in Warp
    return wrapper_GDALWarpDestDS(destNameOrDestDS, srcDSTab, opts, callback, callback_data)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.12/site-packages/osgeo/gdal.py", line 4811, in wrapper_GDALWarpDestDS
    return _gdal.wrapper_GDALWarpDestDS(*args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: in method 'wrapper_GDALWarpDestDS', argument 1 of type 'GDALDatasetShadow *'
```

PR #422, opened 2025-01-07
Adding type checking for base, target path inputs to `warp_raster`

phargogh:
This PR is a simple fix to #421, where we'd get to a cryptic GDAL error if the developer accidentally provided a source or target path that wasn't a string.  In my case, I passed a tuple.  Now the function will just raise `ValueError` with a short but hopefully more helpful error message if the path provided is not a string.

Fixes:#421


claire-simpson (approved):
Looks good!

PR #428, opened 2025-04-08
Support for multipolygons in `zonal_statistics`

### LINKED ISSUE #322: Confirm and document whether or not `zonal_statistics` works for other geometries
The docstring and implementation of `zonal_statistics` suggest it's meant for polygons, but it might work on other geometries too. I accidentally ran it on a point veector and got results, but I don't know if they're correct. We should confirm whether or not `zonal_statistics` produces valid results on other geometries. If so, update the documentation and tests to support this. If not, assert that the input geometries are indeed polygons.

emlys:
Fixes #322 
- Check the input geometry type, and raise an error if it's not Polygon or MultiPolygon
- When creating the temporary vector used in `zonal_statistics`, use the appropriate geometry type according to the input type (Polygon or MultiPolygon) to prevent the warnings seen sometimes in invest.

A PR in invest (natcap/invest#1882) goes along with this, but the order of merging doesn't matter.

### DIFF
--- a/tests/test_geoprocessing.py
+++ b/tests/test_geoprocessing.py
@@ -956,7 +956,7 @@ class TestGeoprocessing(unittest.TestCase):
 
         srs = osr.SpatialReference()
         srs.ImportFromEPSG(4326)
-        layer = vector.CreateLayer('small_vector', srs=srs)
+        layer = vector.CreateLayer('small_vector', srs=srs, geom_type=ogr.wkbPolygon)
         layer.CreateField(ogr.FieldDefn('expected_value', ogr.OFTInteger))
         layer_defn = layer.GetLayerDefn()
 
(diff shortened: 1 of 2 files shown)


tests/test_geoprocessing.py
```diff
@@ -1011,38 +1011,22 @@ def test_zonal_stats_for_small_polygons(self):
...
             (x_pos, y_pos)])
-        new_feature = ogr.Feature(layer_defn)
-        new_geometry = ogr.CreateGeometryFromWkb(shapely_feature.wkb)
-        new_feature.SetGeometry(new_geometry)
-        layer.CreateFeature(new_feature)
-        layer.CommitTransaction()
-        layer.SyncToDisk()
-        layer = None
-        vector = None
+        _geometry_to_vector([geom], vector_path, projection_epsg=4326)
```
emlys, line 1029: Updating a few unrelated tests to use `_geometry_to_vector` - better to avoid duplicating all the layer creation code, and it sets the geometry type correctly, which is now required.

emilyanndavis (approved):
Thanks, @emlys! Looking forward to seeing fewer warnings in the InVEST logs. :)

Issue #288, opened 2022-12-13
`raster_calculator`'s raster dimensions checking doesn't list filepaths that differ


phargogh:
When running `raster_calculator` with rasters of 2 different resolutions, the exception lists the raster sizes but not which sizes belong to which rasters.  This information is available to us, so it'd be nice to make a more helpful error message and make it easier to debug.

megannissel:
Addressed in #430

PR #430, opened 2025-04-14
Include filepaths in `raster_calculator` mistmatched dimensions error

megannissel:
Addresses #288 

`raster_calculator` issues an error if dimensions of the input rasters don't match. The error message only included the differing dimensions, however, without the associated filepaths. Including filepaths _and_ dimensions will make it easier for a user to debug / fix the offending raster(s). 

I also noticed we could be making use of the `base_raster_path_band_list`, which was already filtered on `_is_raster_path_band_formatted`, rather than re-running this check against the `base_raster_path_band_const_list`.


dcdenu4 (approved):
Looks good to me!

PR #434, opened 2025-06-18
raster_calculator stats to include percentage of valid pixels

### LINKED ISSUE #431: Feature: `raster_calculator` to calculate `STATISTICS_VALID_PERCENT`
See the proposed change to `raster_calculator` below, but first some background on GDAL stats.

#### GDAL Stats overview
Statistics are stored as [band metadata](https://gdal.org/en/stable/user/raster_data_model.html#raster-band). Key:value pairs with a `STATISTICS_` prefixed key.

6 metadata attributes are potentially defined when `band.ComputeStatistics()` is called:

>Optional statistics stored in metadata:
> - STATISTICS_MEAN: mean
> - STATISTICS_MINIMUM: minimum
> - STATISTICS_MAXIMUM: maximum
> - STATISTICS_STDDEV: standard deviation
> - STATISTICS_APPROXIMATE: only present if GDAL has computed approximate statistics
> - STATISTICS_VALID_PERCENT: percentage of valid (not nodata) pixel

An example, here's a band that has no statistics pre-calculated.
```
>>> band.GetMetadata()
{}
>>> band.GetStatistics(0, 0)  # (do not approximate, do not force calculations if stats don't already exist)
>>> band.ComputeStatistics(0)  # (do not approximate)
[1638.0, 3121.0, 2132.9436720268745, 326.68508520188595]
>>> band.GetMetadata()
{'STATISTICS_MAXIMUM': '3121', 'STATISTICS_MEAN': '2132.9436720269', 'STATISTICS_MINIMUM': '1638', 'STATISTICS_STDDEV': '326.68508520189', 'STATISTICS_VALID_PERCENT': '41.09'}
>>>
```

While `GetStatistics()` and `ComputeStatistics()` only ever return four stats (min, max, mean, stddev), `ComputeStatistics()` does also calculate and set `STATISTICS_VALID_PERCENT` as band metadata.

### Current behavior:
`calc_raster_stats=True` is a default argument. If true, it calculates min, max, mean, stddev while it's already iterating over the raster blocks. At the end, it sets those values using `band.SetStatistics()`

### Proposed change:
Also calculate `STATISTICS_VALID_PERCENT` as the percentage (0 - 100) of pixels that are not nodata. Set this as band metadata using `band.SetMetadata()`

## Other pygeoprocessing functions
Other functions create rasters but do not calculate statistics. I'm **not proposing these as changes**, but it's open for discussion.

Some functions besides `raster_calculator` use `iterblocks` and could theoretically calculate stats using the same approach as `raster_calculator`.
* `interpolate_points`
* `convolve_2d`

Other functions call `gdal.Warp` or `gdal.RasterizeLayer` to create rasters. If these functions were to calculat stats, it would probably happen _after_ the raster is created, using `band.ComputeStatistics()`, and so would be less optimal than the `raster_calculator` approach.
* `warp_raster`
* `rasterize`
* `align_and_resize_raster_stack`
* `mask_raster`

davemfish:
Fixes #431 

This follows GDAL conventions & behavior for what gets written to the band metadata when `band.ComputeStatistics(0)` is called.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,6 +1,12 @@
 Release History
 ===============
 
+.. Unreleased Changes
+* ``raster_calculator`` ``calc_raster_stats=True`` will count the number
+  of non-nodata pixels in the target raster and write
+  'STATISTICS_VALID_PERCENT' metadata along with other stats.
+  https://github.com/natcap/pygeoprocessing/issues/431
+
 2.4.8 (2025-05-02)
 ------------------
 * ``zonal_statistics`` raises an error if the vector is not a Polygon
--- a/docs/environment-rtd.yml
+++ b/docs/environment-rtd.yml
@@ -8,7 +8,7 @@ name: env-readthedocs
 channels:
     - conda-forge
 dependencies:
-    - python=3.8
+    - python=3.12
     - gdal>=3
     - numpy
     - rtree
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -435,6 +435,7 @@ def raster_calculator(
 
             pixels_processed = 0
             n_pixels = n_cols * n_rows
+            valid_pixel_count = 0
 
             # iterate over each block and calculate local_op
             for block_offset in block_offset_list:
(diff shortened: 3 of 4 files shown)


davemfish:
The Windows 3.13 wheel build is probably failing for this reason: https://github.com/python/cpython/issues/135151

phargogh (approved):
Thanks @davemfish !  Everything looks good to me.

Reading through the thread in https://github.com/python/cpython/issues/135151, the build failure should be addressed in 3.13.5, which is already out by now but clearly hasn't made its way into github's `setup-python` action.  The build issue here should be resolved once that happens.

PR #443, opened 2025-07-11
Replace cython _ManagedRaster with shared C++ ManagedRaster implementation

### LINKED ISSUE #439: Migrate natcap.invest.managed_raster to pygeoprocessing
natcap.invest extensions now use a shared C++ implementation of `ManagedRaster`, but pygeoprocessing still has its own Cython implementation. We should move the C++ implementation from natcap.invest to pygeoprocessing, and natcap.invest should reference it from pygeoprocessing.

This will involve adding some complexity to the package build due to compiling C++.

emlys:
Fixes #439 

- Copied over C++ ManagedRaster extension from natcap.invest
- Created a new cython module `pygeoprocessing.extensions` which exposes the C++ extensions `ManagedRaster`, `LRUCache`, `FastFileIterator`, and associated utilities
- Updated the setup.py configuration to support the new C++ extension (largely copied from natcap.invest's setup.py). This includes adding a new required environment variable when building on Windows, `PYGEOPROCESSING_GDAL_LIB_PATH`.
- Replace Cython implementations of `ManagedRaster` with the new C++ implementation
- Replace float comparison functions with the new shared function `is_close`
- Use shared definitions of some module-level variables rather than re-defining in each module
- Ensure that all `ManagedRaster`s are closed using the `.close()` method

When these changes are released, we can remove the `ManagedRaster` implementation from natcap.invest and instead import it from pygeoprocessing.


src/pygeoprocessing/geoprocessing_core.pyx
```diff
@@ -454,7 +442,7 @@ def calculate_slope(
                 # def
                 # ghi
                 e = dem_array[row_index, col_index]
-                if _eq(e, dem_nodata):
+                if is_close(e, dem_nodata):
```
emlys, line 445: `_eq` did not allow for float imprecision, but it seems like it should have.

.github/workflows/build-py-dists.yml
```diff
@@ -1,5 +1,20 @@
...
+
+concurrency:
+  # make sure only one run of this workflow for a given PR or a given branch
+  # can happen at one time. previous queued or started runs will be cancelled.
+  # github.workflow is the workflow name
+  # github.ref is the ref that triggered the workflow run
+  # on push, this is refs/heads/<branch name>
+  # on pull request, this is refs/pull/<pull request number>/merge
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
emlys, line 12: Copied this over from natcap.invest. Keeps queued jobs from building up.

phargogh (commented):
Makes sense to me, thanks @emlys !

By the way, I know this isn't a user-facing change, but since it is a significant change, do you think we should make a note in HISTORY about it?

emlys:
Sure! Added a history note

phargogh (approved):
Thanks @emlys !

PR #444, opened 2025-07-16
Test and document using GDAL virtual file system handlers

### LINKED ISSUE #441: Officially support GDAL virtual file system inputs
Pygeoprocessing should support reading data from remote sources. GDAL supports this very nicely with [virtual file system handlers](https://gdal.org/en/stable/user/virtual_file_systems.html#vsicurl-http-https-ftp-files-random-access), and it looks like we always interact with input data via GDAL, so I would expect input data on virtual file systems to “just work” (though we should test this more).

Action items:
- Add a few tests using `/vsicurl/` to make sure that basic functionality is working (`get_raster_info`, `get_vector_info`, `raster_map` seem like good candidates)
- Update all docstrings to highlight that GDAL virtual file system handlers are supported on input paths
- Update API docs to point out that GDAL virtual file system handlers are supported on input paths, and that they are NOT officially supported on outputs, but may work.

More context from slack:
> Given that this has likely worked for quite some time without being explicitly supported, I’m not inclined to add any restrictions on where virtual file inputs are accepted. We could update the pygeoprocessing documentation, or perhaps every docstring, to indicate that GDAL VSI paths are supported for inputs. And we should be aware of this expectation when writing new code.
As far as outputs, I’m inclined to say we don’t need to restrict that either. At the level of pygeoprocessing, I’d say it’s up to the user - they are welcome to try using a VSI output, and it MAY work, but we do not guarantee it. It’s up to the user to understand memory limitations etc. of VSI outputs. They should also realize that using a remote output path does not mean that all data will be written there - there are many intermediate files produced on the local filesystem.
The contract is clearly different for invest, where I think it would make sense to support specific VSIs for inputs only.
As far as authentication and other features, I’m inclined to start simple (e.g. using GDAL’s env variables, credential files) and only add abstractions around that as we find a need for them.

emlys:
Fixes #441 
- Add a few tests using vsicurl (testing against new test data included in this repo, so that we can reference it at the github hosted URL)
- Add a section to the documentation about using virtual file system handlers
- Update all docstrings to indicate that GDAL vsi paths are supported for inputs


src/pygeoprocessing/geoprocessing.py
```diff
@@ -876,8 +882,8 @@ def align_and_resize_raster_stack(
               > 10'`` would use all features whose field value of 'id' is >
               10.
             * ``'mask_raster_path'`` (str): Optional. the string path to where
-              the mask raster should be written on disk.  If not provided, a
-              temporary file will be created within ``working_dir``.
+              the mask raster should be written/  If not provided, a temporary
```
dcdenu4:
```suggestion
              the mask raster should be written. If not provided, a temporary
```

dcdenu4: Should `mask_vector_path` above get the blurb too?

src/pygeoprocessing/geoprocessing.py
```diff
@@ -2450,7 +2467,8 @@ def warp_raster(
     """Resize/resample raster to desired pixel size, bbox and projection.
 
     Args:
-        base_raster_path (string): path to base raster.
+        base_raster_path (string): path to base raster. Paths may use any
+            GDAL-supported scheme, including virtual file system /vsi schemes.
```
dcdenu4, line 2471: I couldn't leave a comment on the appropriate line, but should we have the blurb for `mask_vector_path` and `mask_raster_path` under `mask_options` below? Starting on line 2494.

src/pygeoprocessing/geoprocessing.py
```diff
@@ -3138,7 +3159,8 @@ def convolve_2d(
 
     Args:
         signal_path_band (tuple): a 2 tuple of the form
-            (filepath to signal raster, band index).
+            (filepath to signal raster, band index). Paths may use any
+            GDAL-supported scheme, including virtual file system /vsi schemes.
         kernel_path_band (tuple): a 2 tuple of the form
```
dcdenu4, line 3167: Should we have a blurb for `kernel_path_band`?

tests/test_geoprocessing.py
```diff
@@ -5682,6 +5682,31 @@ def test_get_raster_info_overviews(self):
...
+    def test_raster_map_vsicurl(self):
+        """PGP: raster_map with vsicurl."""
+        # Access test data hosted on github
+        input_path = '/vsicurl/https://raw.githubusercontent.com/emlys/pygeoprocessing/feature/441/tests/test_data/raster.tif'
+        target_path = os.path.join(self.workspace_dir, 'out.tif')
+        pygeoprocessing.raster_map(
+            lambda block: block * 2, [input_path], target_path)
+        numpy.testing.assert_array_equal(
+            pygeoprocessing.raster_to_numpy_array(target_path),
+            numpy.ones((9, 9), dtype=numpy.float32))
```
dcdenu4, line 5708: This got me thinking that it'd be nice to document what `test_data/raster.tif` is and it's properties. Maybe a more descriptive name for that test file could help at the very least. But if I wanted to use it in another test function I wouldn't know what to expect for values.

tests/test_geoprocessing.py
```diff
@@ -5682,6 +5682,31 @@ def test_get_raster_info_overviews(self):
...
+        """PGP: vector info with vsicurl."""
+        # Access test data hosted on github
+        path = '/vsicurl/https://raw.githubusercontent.com/emlys/pygeoprocessing/feature/441/tests/test_data/vector.gpkg'
+        vector_info = pygeoprocessing.get_vector_info(path)
+        self.assertEqual(vector_info['bounding_box'], [-15, 14.8, -14.8, 15])
+
+    def test_raster_map_vsicurl(self):
+        """PGP: raster_map with vsicurl."""
+        # Access test data hosted on github
+        input_path = '/vsicurl/https://raw.githubusercontent.com/emlys/pygeoprocessing/feature/441/tests/test_data/raster.tif'
```
dcdenu4: In playing around with this, do you have a sense for where GDAL downloads the data it needs from a hosted file? Are we sure we're cleaning that up in our tests?

dcdenu4 (changes requested):
Thanks @emlys , I found a couple other places we might want to add the vsi blurb too. One I couldn't really comment on is under the `def build_overviews` functions in `geoprocessing.py`, line 4670.

When thinking about testing a few questions came to mind that I was curious if you'd come across or had thought about.
1) Where does GDAL save the needed downloaded data from URLs and how / when does GDAL garbage collect it? Is that something we need to be mindful of?
2) I'd love to have some more comprehensive testing that could help us spot issues further in advance, but I know we don't want to burden our GHA runners all the time. It might be nice to start thinking about having a more comprehensive test runner / suite that we can manually run every once in awhile or that runs periodically during off hours.

src/pygeoprocessing/geoprocessing.py
emlys: Good catch, added those

src/pygeoprocessing/geoprocessing.py
emlys: Added

tests/test_geoprocessing.py
emlys: I don't know that it downloads it to anywhere - it may just be in memory. I wouldn't expect it to leave any extra files around afterward

tests/test_geoprocessing.py
emlys: Hmm, I'm not sure exactly how to handle that. I renamed the files to `small_raster.tif` and `small_vector.gpkg` so hopefully that helps a bit? But I don't want to be too specific e.g. `small_raster_for_vsi_tests.tif` would then not make sense if we did want to use it in another test.

emlys:
Thanks @dcdenu4! Here are my thoughts -

_Where does GDAL save the needed downloaded data from URLs and how / when does GDAL garbage collect it? Is that something we need to be mindful of?_

In the [docs](https://gdal.org/en/stable/user/virtual_file_systems.html) I do not see any mention of downloaded data being permanently stored. I think it's safe to assume that this is abstracted away by the drivers. The documentation suggests that everything is cached and there is a reasonable limit on the cache size:
> Partial downloads (requires the HTTP server to support random reading) are done with a 16 KB granularity by default. Starting with GDAL 2.3, the chunk size can be configured with the [CPL_VSIL_CURL_CHUNK_SIZE](https://gdal.org/en/stable/user/configoptions.html#config-CPL_VSIL_CURL_CHUNK_SIZE) configuration option, with a value in bytes. If the driver detects sequential reading, it will progressively increase the chunk size up to 128 times [CPL_VSIL_CURL_CHUNK_SIZE](https://gdal.org/en/stable/user/configoptions.html#config-CPL_VSIL_CURL_CHUNK_SIZE) (so 2 MB by default) to improve download performance.
In addition, a global least-recently-used cache of 16 MB shared among all downloaded content is used, and content in it may be reused after a file handle has been closed and reopen, during the life-time of the process or until [VSICurlClearCache()](https://gdal.org/en/stable/api/cpl.html#_CPPv417VSICurlClearCachev) is called. Starting with GDAL 2.3, the size of this global LRU cache can be modified by setting the configuration option [CPL_VSIL_CURL_CACHE_SIZE](https://gdal.org/en/stable/user/configoptions.html#config-CPL_VSIL_CURL_CACHE_SIZE) (in bytes).

_I'd love to have some more comprehensive testing that could help us spot issues further in advance, but I know we don't want to burden our GHA runners all the time. It might be nice to start thinking about having a more comprehensive test runner / suite that we can manually run every once in awhile or that runs periodically during off hours._

I think this could be addressed separately... I don't want to fall into the trap of testing GDAL functionality rather than pygeoprocessing functionality. VSI handlers are new to us, but not new to GDAL, so I think we can assume that they work as described and not need to write our own comprehensive tests using that feature.

I did also test locally running `raster_map` on a huge global dataset from the data hub, and it worked as expected and I observed that memory use did not climb during the run time.

tests/test_geoprocessing.py
dcdenu4: Would it be overkill to have a small README in the folder that describes the data? I'm mostly thinking that I don't know what the values are and therefore what I'd expect for some kind of output. Maybe we can address this in the future if we do more of this.

dcdenu4 (approved):
Thanks @emlys for taking the time to reply to my questions.

PR #447, opened 2025-09-26
Move `GDALUseExceptions` and `gdal_use_exceptions` into new utils module

### LINKED ISSUE #446: Move `gdal_use_exceptions` decorator out of cython
`gdal_use_exceptions` is currently defined in the `geoprocessing_core` cython module. Even though the function doesn't use any cython features, its type is `<cyfunction>` rather than `<function>`. This has a strange side effect: when I tried to use the `gdal_use_exceptions` decorator on a method of a Pydantic class in `natcap.invest`, Pydantic doesn't recognize it as a function, and raises the error:
```
pydantic.errors.PydanticUserError: A non-annotated attribute was detected: 
`do_execute = <cyfunction ModelSpec.do_execute at 0x166c4eb80>`. All model 
fields require a type annotation; if `do_execute` is not meant to be a field, 
you may be able to resolve this error by annotating it as a `ClassVar` or 
updating `model_config['ignored_types']`
```

The suggested solutions are not viable: the method should not be a `ClassVar`, and adding `cyfunction` to `model_config['ignored_types']` would require adding `cython` as a runtime dependency of `natcap.invest`. Much easier to move `gdal_use_exceptions` into a regular python module.

emlys:
Fixes #446 

Moving these functions out of cython into a regular python module to prevent the issue described in #446. To avoid circular imports, I had to make a new module to contain them, which I named `utils`.

### DIFF
--- a/src/pygeoprocessing/__init__.py
+++ b/src/pygeoprocessing/__init__.py
@@ -50,6 +50,8 @@ from .geoprocessing import zonal_statistics
 from .geoprocessing_core import calculate_slope
 from .geoprocessing_core import raster_band_percentile
 from .slurm_utils import log_warning_if_gdal_will_exhaust_slurm_memory
+from .utils import GDALUseExceptions
+from .utils import gdal_use_exceptions
 
 try:
     __version__ = version('pygeoprocessing')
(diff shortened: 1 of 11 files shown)


claire-simpson (changes requested):
Looks good! My only note is that `test_pygeoprocessing.py` still contains the line `from pygeoprocessing.geoprocessing_core import gdal_use_exceptions`

emlys:
Thanks for catching that! Back to you

claire-simpson (approved):

PR #450, opened 2025-12-10
Expand `zonal_statistics` supported geometry types

### LINKED ISSUE #448: Be less strict about polygon geometry types in zonal statistics
`wkbPolygon` and `wkbMultiPolygon` are currently supported. But it's possible we could support some other polygon types too, perhaps by converting them to one of these base types if necessary. https://github.com/OSGeo/gdal/blob/master/ogr/ogr_core.h

For example, here's a 3D Polygon that got rejected. All its z-values are `0`. I'm guessing this is the most common case when we encounter a non-traditional polygon geometry type. It could be worth accommodating it. 
https://community.naturalcapitalproject.org/t/summary-statistic-issue/6223/4

megannissel:
Fixes #448 

`zonal_statistics` previously only supported `wkbPolygon` and `wkbMultiPolygon` geometries. However, it would be reasonable to support 3D and measured (multi)polygons as well, as the z coordinates shouldn't have an affect / should be ignored for the purposes of the zonal stats calculations. 

This change also brings `zonal_statistics` in line with InVEST validation of input vectors, where the model spec input `geometry_types` maps `'POLYGON'` and `'MULTIPOLYGON'` to several ogr types:
```
'POLYGON': [ogr.wkbPolygon, ogr.wkbPolygonM,
            ogr.wkbPolygonZM, ogr.wkbPolygon25D],
'MULTIPOLYGON': [ogr.wkbMultiPolygon, ogr.wkbMultiPolygonM,
                 ogr.wkbMultiPolygonZM, ogr.wkbMultiPolygon25D]
```
Allowing for the same list of valid geometries will prevent InVEST models that calculate `zonal_statistics` from failing on vectors that would otherwise be valid InVEST inputs.

### DIFF
--- a/tests/test_geoprocessing.py
+++ b/tests/test_geoprocessing.py
@@ -1353,23 +1353,23 @@ class TestGeoprocessing(unittest.TestCase):
 
         polygon_a = [
             (origin[0], origin[1]),
-            (origin[0], -pixel_size * n_pixels+origin[1]),
-            (origin[0]+pixel_size * n_pixels,
-             -pixel_size * n_pixels+origin[1]),
-            (origin[0]+pixel_size * n_pixels, origin[1]),
+            (origin[0], -pixel_size * n_pixels + origin[1]),
+            (origin[0] + pixel_size * n_pixels,
+             -pixel_size * n_pixels + origin[1]),
+            (origin[0] + pixel_size * n_pixels, origin[1]),
             (origin[0], origin[1])]
         polygon_b = [
             (origin[0], origin[1]),
-            (origin[0], -pixel_size+origin[1]),
-            (origin[0]+pixel_size, -pixel_size+origin[1]),
-            (origin[0]+pixel_size, origin[1]),
+            (origin[0], -pixel_size + origin[1]),
+            (origin[0] + pixel_size, -pixel_size + origin[1]),
+            (origin[0] + pixel_size, origin[1]),
             (origin[0], origin[1])]
         polygon_c = [
             (origin[1]*2, origin[1]*3),
-            (origin[1]*2, -pixel_size+origin[1]*3),
-            (origin[1]*2+pixel_size,
-             -pixel_size+origin[1]*3),
-            (origin[1]*2+pixel_size, origin[1]*3),
+            (origin[1]*2, -pixel_size + origin[1]*3),
+            (origin[1]*2 + pixel_size,
+             -pixel_size + origin[1]*3),
+            (origin[1]*2 + pixel_size, origin[1]*3),
             (origin[1]*2, origin[1]*3)]
 
         layer.StartTransaction()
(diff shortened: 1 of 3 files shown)


tests/test_geoprocessing.py
```diff
@@ -1333,6 +1333,96 @@ def test_zonal_statistics(self):
...
+
+        projection = osr.SpatialReference()
+        projection.ImportFromEPSG(3116)
+        layer = vector.CreateLayer('aggregate', srs=projection,
+                                   geom_type=ogr.wkbPolygon25D)
+        layer_defn = layer.GetLayerDefn()
+
+        polygon_a = [
+            (origin[0], origin[1]),
+            (origin[0], -pixel_size * n_pixels+origin[1]),
```
davemfish: It's a really minor thing, but I think these would be more readable with spaces around the `+` operator. Especially since that operator is lower priority than the `*` which does have spaces. Alternatively, no spaces around `*` and spaces around `+` would also be more readable, I think.

tests/test_geoprocessing.py
```diff
@@ -1375,6 +1465,80 @@ def test_zonal_statistics_multipolygon(self):
...
+        vector = driver.CreateDataSource(aggregating_vector_path)
+
+        projection = osr.SpatialReference()
+        projection.ImportFromEPSG(3116)
+        layer = vector.CreateLayer('aggregate', srs=projection,
+                                   geom_type=ogr.wkbMultiPolygonZM)
+        layer_defn = layer.GetLayerDefn()
+
+        origin = (444720, 3751320)
+        polygon_a = [
```
davemfish: I wonder if all these zonal stats tests use the same primitive geometries, and just build them up in different ways to get the different high-order types? If so, do you think it makes sense to define these polygons, and maybe other related variables, someplace where they can be shared? Okay, looking at the rest of the test suite, there are many zonal stats tests and lots of duplicated code, so I won't suggest you do a refactor of all of that here. I might make an issue and do it separately though.

davemfish (commented):
@megannissel this looks good to me.

As far as the tests, I think testing all the different geometry types is probably not necessary. It certainly doesn't hurt, and if it was useful to write the tests in order to convince ourselves that all the different types are treated the same way, then that's a good enough reason to write the tests. But our function here does not do anything different with the different types, so I think we're ultimately testing things like `gdal.Rasterize` more than we're testing `pygeoprocessing.zonal_statistics`. And in that sense, I don't think we need to test gdal.

megannissel:
Writing these tests and confirming that nothing unexpected happened did make me feel more confident that I wasn't overlooking anything. But completely agree that we don't really need to be testing gdal itself.

davemfish (changes requested):
Thanks for the update, @megannissel . I forgot to suggest that we update HISTORY, but I think we should do that too.

davemfish (approved):

PR #452, opened 2025-12-11
Use bool rather than int in ManagedRaster cython constructor

### LINKED ISSUE #451: Routing tests failing with `TypeError: Expected bool, got int`
Almost all routing-related tests have begun failing, e.g.:

```
>   dem_raster_path_band[0].encode('utf-8'), dem_raster_path_band[1], 0)
E   TypeError: Expected bool, got int

src/pygeoprocessing/routing/routing.pyx:4103: TypeError
```

The third parameter to the `ManagedRaster` constructor is a bool and the type casting no longer works. This is likely the result of a change in cython 3.2.0, which released on November 5. @emlys [fixed the same error in InVEST](https://github.com/natcap/invest/pull/2240) last month.

megannissel:
Fixes #451 

The third parameter to the `ManagedRaster` constructor is a bool and the type casting that had previously worked no longer does. This is likely the result of a change in cython 3.2.0, which released on November 5. The same fix was implemented in InVEST in November.

### DIFF
--- a/src/pygeoprocessing/routing/routing.pyx
+++ b/src/pygeoprocessing/routing/routing.pyx
@@ -372,7 +372,7 @@ def fill_pits(
         [mask_nodata],
         raster_driver_creation_tuple=raster_driver_creation_tuple)
     flat_region_mask_managed_raster = ManagedRaster(
-        flat_region_mask_path.encode('utf-8'), 1, 1)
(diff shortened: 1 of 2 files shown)


emlys (approved):
Thanks @megannissel!

PR #457, opened 2026-01-13
Restore building wheels against each supported python version

### LINKED ISSUE #456: All wheels are being built against python 3.13
While working through the 2.4.10 release, I realized all of the wheels for each OS have the same name. I dug into the logs and it looks like they're all being built against python 3.13. I think this is the result of changes made in [this commit](https://github.com/natcap/pygeoprocessing/commit/0ad9e22a65f4c969c86180a3564430f736542feb), which dropped the reference to `${{ matrix.python-version }}` in the `build-wheels` job.

megannissel:
Fixes #456 

Brings back the reference to `${{ matrix.python-version }}` when creating the micromamba environment, so that we can build wheels against each supported Python version.

### DIFF
--- a/.github/workflows/build-py-dists.yml
+++ b/.github/workflows/build-py-dists.yml
@@ -41,7 +41,7 @@ jobs:
                   channels:
                     - conda-forge
                 create-args: >-
-                  python=3.13
+                  python=${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


emlys (approved):
Thanks!

Issue #454, opened 2026-01-09
`reclassify_raster` does not handle `numpy.nan` nodata


davemfish:
It's somewhat common for `nan` to be used as a raster nodata value. `reclassify_raster` can either reclassify  nodata pixels -- if it is listed in the value map -- or propagate them as nodata in the output. Neither of these will work for `nan` because we use `numpy.isclose()` without passing `equal_nan`.

I feel like we've discussed this in the past, but I don't remember if there's a reason we did not address it. It came up in #419 . @claire-simpson do you remember?

claire-simpson:
I think we could use `equal_nan`! I believe that the issue I noted in #419:

> a `numpy.nan` key will cause an issue later in the function

is related to these 2 lines:
```python
keys = sorted(numpy.array(list(value_map_copy.keys())))
values = numpy.array([value_map_copy[x] for x in keys])
```

There is some weirdness when trying to sort a `numpy.nan` key (in `value_map_copy`) in this line: `keys = sorted(numpy.array(list(value_map_copy.keys())))` that ends up causing a `KeyError` when trying to create the `values` array.I think this is because the dictionary key is initially `numpy.nan` but then gets converted to `numpy.float64(nan)` which is different and therefore not a valid key in `value_map_copy`.

e.g.
```python
value_map_copy = {1:11.0, 3:33, numpy.nan: -999}
print(value_map_copy) # {1: 11.0, 3: 33, nan: -999}
keys = sorted(numpy.array(list(value_map_copy.keys())))
print(keys) # [np.float64(1.0), np.float64(3.0), np.float64(nan)]
values = numpy.array([value_map_copy[x] for x in keys]) # raises KeyError: np.float64(nan)
```

However I think using `equal_nan` could actually solve this issue _if the input raster's nodata value is numpy.nan_, because, if `nan` is initially in the `value_map`, it would get deleted before `values = numpy.array([value_map_copy[x] for x in keys])`. Or maybe we should just be detecting `nan`s anyway and handling them separately. TLDR; Using `equal_nan` seems conceptually correct to me and we should handle `nan` values in the `value_map` so they don't cause a `KeyError`.

davemfish:
This makes sense; thanks for the refresher.

phargogh:
This was resolved in #465

PR #465, opened 2026-02-27
Handling NaN nodata values in reclassify_raster

phargogh:
This small PR handles the case where a raster passed to `pygeoprocessing.reclassify_raster` has a NaN nodata value.  I chose to use `pygeoprocessing.array_equals_nodata` instead of simply adding `equals_nan=True` to `numpy.isclose` so that we could take advantage of the special case speedups that are handled in that helper function.

Dave and Claire, I requested a review from both of you because you were discussing this over in #454.

Fixes:#454

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -7,6 +7,9 @@ Unreleased Changes
   References to the old name and website domain have been updated to reflect
   this change.
   https://github.com/natcap/pygeoprocessing/issues/458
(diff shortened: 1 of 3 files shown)


claire-simpson (approved):
Looks good to me, thanks @phargogh!

davemfish (approved):
Thanks, @phargogh

PR #467, opened 2026-02-27
Adding average to zonal_statistics output

### LINKED ISSUE #370: Add "average" metric to zonal statistics
`zonal_statistics` would benefit from returning the `average` value under a polygon. It already is aware of `sum` and `count` and would be a no cost value add that would benefit users, who currently would need to iterate through the returned object to compute themselves. Seems like a nice quality of life enhancement.

phargogh:
This implements an 'average' output in zonal_statistics. 

Fixes: #370


megannissel (approved):
Thanks, @phargogh, this will be a very helpful addition to zonal statistics!

PR #468, opened 2026-02-27
Correcting classifier usage

### LINKED ISSUE #466: Use SPDX license expression instead of trove classifiers
When running the pygeoprocessing conda-forge build, I noticed this big error message in the logs:

```
          ********************************************************************************
          Please consider removing the following classifiers in favor of a SPDX license expression:

          License :: OSI Approved :: BSD License

          See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
          ********************************************************************************
```

I think we did something similar for InVEST, so this should be very straightforward.

phargogh:
Just updating to the latest standard keywords in pyproject.toml and removing the license classification from the Trove classifiers.  Fixes: #466

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -7,6 +7,9 @@ Unreleased Changes
   References to the old name and website domain have been updated to reflect
   this change.
   https://github.com/natcap/pygeoprocessing/issues/458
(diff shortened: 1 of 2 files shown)


emlys (approved):
👍

PR #472, opened 2026-03-06
Fixing alignment side effect

### LINKED ISSUE #471: `align_and_resize_raster_stack` mutates input bounding box list in-place
`align_and_resize_raster_stack` mutates the bounding box object passed to `bounding_box_mode` when it is provided as a list of bounding box coordinates. This mutation happens in-place, which led to subtle bugs when the same bounding box object was reused elsewhere in my workflow.

I discovered this while developing Urban Mental Health model, and experienced the following issue:
 
Ran model for the first time in a given workspace, during which the following steps occurred:
(1) create `bounding_box` by getting the extent of the AOI
(2) use that `bounding_box` in `align_and_resize_raster_stack` `bounding_box_mode` param --> `align_and_resize_raster_stack` mutates bounding box (in a way that is actually helpful for my downstream tasks...)
(3) use `bounding_box` in a separate task (to resample my population raster); `bounding_box` has been (accidentally) mutated by PGP during the previous task, so my population raster is clipped to this new bounding box which is different than the one I initially calculated

I run the model again in the same workspace. Taskgraph recognizes that the aligned rasters from the prior model run do not need to be recreated so `align_and_resize_raster_stack` is not called. The population resample task is re-run (because of the different bounding box). _However in this model run, since `align_and_resize_raster_stack` was not called and therefore didn't mutate `bounding_box`, `bounding_box` is different and so the population raster is clipped to a different bounding box than in the first model run._ Later on in my workflow, I get a `ValueError: Input Rasters are not the same dimensions.` 

I think this issue is here in line 990: `target_bounding_box = bounding_box_mode`
And then later in line 1084 `target_bounding_box` is modified in place:
```python
for index in [0, 1]:
            n_pixels = int(
                (target_bounding_box[index] - align_bounding_box[index]) /
                float(align_pixel_size[index]))
            target_bounding_box[index] = (
                n_pixels * align_pixel_size[index] +
                align_bounding_box[index])
```

Could be fixed by just changing to `target_bounding_box = list(bounding_box_mode)`

phargogh:
Just a quick PR to fix the side effect, and testing for it as well.

Fixes #471

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -16,6 +16,10 @@ Unreleased Changes
 * Updating pyproject.toml to use the standard ``license-files`` key and
   replacing the license-related Trove classifier with the approved SPDX string.
   https://github.com/natcap/pygeoprocessing/issues/466
+* Fixing a side effect in ``pygeoprocessing.align_and_resize_raster_stack``
+  where aligning a raster stack using the bounding box mode would result in the
+  input bounding box being modified in-place.
+  https://github.com/natcap/pygeoprocessing/issues/471
 
 2.4.10 (2026-01-13)
 -------------------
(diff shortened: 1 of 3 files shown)


claire-simpson (approved):
Thanks!

PR #474, opened 2026-03-11
Fix `NaN` NoData handling in `convolve_2d`

### LINKED ISSUE #473: `convolve_2d` doesn't handle `nan` NoData values correctly
If you run `convolve_2d` on a raster that has any `nan` values, even if `nan` is set as the raster's `NoData` value, the resulting output raster will have big holes in it (or even be entirely `nan`). This is because we're using `numpy.isclose` to check if the pixel values are NoData, however `numpy.isclose(numpy.nan, numpy.nan)` is False. At minimum, we should be using `equal_nan=True` flag in `numpy.isclose`. We also might considering using `pygeoprocessing.array_equals_nodata`? If `nan` is (correctly) set as the NoData value for a raster, I would expect it to be treated the same as any other NoData value.

Reproducible example:
```python

# Create raster where one pixel = NaN
input_array = numpy.arange(100).reshape((10, 10)).astype(numpy.float32)
input_array[5, 5] = numpy.nan

srs = osr.SpatialReference()
srs.ImportFromEPSG(26910)
projection_wkt = srs.ExportToWkt()
input_raster_path='test_nan_raster.tif'
pygeoprocessing.numpy_array_to_raster(input_array,  target_nodata=numpy.nan, pixel_size=(1, -1), origin=(0, 0), projection_wkt=projection_wkt, target_path=input_raster_path)

# Create kernel raster
kernel_path = 'kernel.tif'
kernels.dichotomous_kernel(
    target_kernel_path=kernel_path,
    max_distance=1,
    normalize=False)

# Convolve
target_path = 'test_nan_raster_convolved.tif'
pygeoprocessing.convolve_2d(
        (input_raster_path, 1), (kernel_path, 1), target_path,
        ignore_nodata_and_edges=True, mask_nodata=True,
        normalize_kernel=True, target_datatype=6,
        target_nodata=numpy.nan)

```

Image

claire-simpson:
Fixes a bug where, if a raster with `NoData=nan` was passed to `convolve_2d`, a large part of/the entire resulting raster would become `nan`. 

I followed #465 and opted to use `pygeoprocessing.array_equals_nodata` (instead of adding `equal_nan=True` to `numpy.isclose`) to take advantage of the special case speedups that are handled in that helper function.

Fixes #473

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -16,13 +16,13 @@ Unreleased Changes
 * Updating pyproject.toml to use the standard ``license-files`` key and
   replacing the license-related Trove classifier with the approved SPDX string.
   https://github.com/natcap/pygeoprocessing/issues/466
-* Fixed a bug where ``convolve_2d`` was not handling ``nan`` NoData
-  values correctly.
-  https://github.com/natcap/pygeoprocessing/issues/473
 * Fixing a side effect in ``pygeoprocessing.align_and_resize_raster_stack``
   where aligning a raster stack using the bounding box mode would result in the
   input bounding box being modified in-place.
   https://github.com/natcap/pygeoprocessing/issues/471
+* Fixed a bug where ``convolve_2d`` was not handling ``nan`` NoData
+  values correctly.
+  https://github.com/natcap/pygeoprocessing/issues/473
 
 2.4.10 (2026-01-13)
 -------------------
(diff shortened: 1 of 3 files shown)


phargogh (approved):
Makes sense to me ... thanks @claire-simpson !

PR #477, opened 2026-03-27
Handle rasters without a defined `NoData` value in `warp_raster`

### LINKED ISSUE #476: `align_and_resize_raster_stack` should ensure output rasters have a defined NoData value
When `align_and_resize_raster_stack` is called on a raster that does not have a NoData value defined and whose extent is smaller than the target bounding box, the function pads the output raster with 0 values. This behavior can silently introduce incorrect data, as these 0 pixels are treated as valid values rather than NoData.

Instead, if an input raster does not have a set NoData value, this function should assign an appropriate one (e.g., using `choose_nodata`) prior to warping. This ensures that, when warping the raster, any edge pixels (where the input raster didn't overlap the target extent) will be assigned this NoData value (which will then correctly be treated as NoData in downstream tasks). I think it would also be useful to log when an input has no defined NoData value and/or is smaller than the target bounding box.

For more context on this issue see https://github.com/natcap/invest/pull/2392#discussion_r2941870672

claire-simpson:
The initial impetus for this PR was: if a raster input to `align_and_resize_raster_stack` does not have a defined `NoData` value, PGP should automatically pick one. We decided to handle this within `warp_raster`, which is called by `align_and_resize_raster_stack`. `warp_raster` now also logs a warning if any band in a raster input has an undefined NoData value. For multiband rasters, if one band in the raster has a NoData value of `None`, all bands will be reassigned a NoData value picked by `pygeoprocessing.choose_nodata`.

~~## Steps:~~
~~1. Detects rasters with missing `NoData` values~~
~~2. Creates a temporary VRT for those rasters and assigns a `NoData` value using `choose_nodata` (this is to avoid modifying a user's original data)~~
~~3. Uses the temporary VRT path during warping/alignment~~
~~4. Logs different warnings if (a) a raster has no `NoData` defined on any band, or (b) a multiband raster has `NoData` defined on some bands but not others, in which case a single `NoData` value is applied to all bands in the temporary VRT~~

Fixes: #476

### DIFF
--- a/src/pygeoprocessing/geoprocessing.py
+++ b/src/pygeoprocessing/geoprocessing.py
@@ -1102,6 +1102,49 @@ def align_and_resize_raster_stack(
                 n_pixels * align_pixel_size[index] +
                 align_bounding_box[index])
 
+    # Set a nodata value if the raster does not have one defined. This
+    # ensures that a raster that is smaller than the target extent will
+    # not get filled with 0s that could be confused with real data.
+    # We do this by creating a temporary VRT with an explicit nodata value
+    # for rasters that are missing one.
+    temp_working_dir_nodata = None
+    fixed_base_raster_path_list = list(base_raster_path_list)
+
+    rasters_missing_nodata = []
+    for raster_info, base_path in zip(raster_info_list, base_raster_path_list):
+        nodata_list = raster_info.get('nodata', [])
+        band_nodata = nodata_list[0] if nodata_list else None
+        if band_nodata is None:
+            rasters_missing_nodata.append(base_path)
+
+    if rasters_missing_nodata:
+        temp_working_dir_nodata = tempfile.mkdtemp(dir=working_dir, prefix='align-nodata-')
+
+        for index, (raster_info, base_path) in enumerate(
+                zip(raster_info_list, base_raster_path_list)):
+            nodata_list = raster_info.get('nodata', [])
+            band_nodata = nodata_list[0] if nodata_list else None
+            if band_nodata is not None:
+                continue
+
+            raster = gdal.OpenEx(base_path, gdal.OF_RASTER)
+            chosen_nodata = choose_nodata(raster_info['numpy_type'])
+
+            vrt_path = os.path.join(temp_working_dir_nodata,
+                                    f'input_with_nodata_{index}.vrt')
+            vrt = gdal.Translate(vrt_path, raster, format='VRT',
+                                 noData=chosen_nodata)
+            vrt = None
+            band = None
+            raster = None
+
+            fixed_base_raster_path_list[index] = vrt_path
+
+            LOGGER.warning(
+                "Input raster %s had no defined nodata value. Using "
+                "%s as nodata in a temporary VRT before warping.",
+                base_path, chosen_nodata)
+
     if mask_options:
         # Create a warped VRT.
         # This allows us to cheaply figure out the dimensions, projection, etc.
@@ -1110,7 +1153,7 @@ def align_and_resize_raster_stack(
         temp_working_dir = tempfile.mkdtemp(dir=working_dir, prefix='mask-')
         warped_vrt = os.path.join(temp_working_dir, 'warped.vrt')
         warp_raster(
-            base_raster_path=base_raster_path_list[0],
+            base_raster_path=fixed_base_raster_path_list[0],
             target_pixel_size=target_pixel_size,
             target_raster_path=warped_vrt,
             resample_method='near',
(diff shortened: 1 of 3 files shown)


src/pygeoprocessing/geoprocessing.py
```diff
@@ -1102,6 +1102,44 @@ def align_and_resize_raster_stack(
...
+        nodata_list = raster_info['nodata']
+        if None in nodata_list:
+            if temp_working_dir_nodata is None:
+                temp_working_dir_nodata = tempfile.mkdtemp(dir=working_dir,
+                                               prefix='align-nodata-')
+            raster = gdal.OpenEx(base_path, gdal.OF_RASTER)
+            chosen_nodata = choose_nodata(raster_info['numpy_type'])
+
+            if set(nodata_list) != {None}:
+                LOGGER.warning(
```
claire-simpson: From what I can tell, gdal does not create geotiffs that have different NoData values (even if you try to explicitly set different NoData values for different bands, it won't work), so I think this would be a very rare situation and therefore not super important or easy to handle gracefully (i.e., by setting NoData only for the bands that lack a defined NoData value, since gdal doesn't support this for tiffs). I also decided not to raise an error as a raster that has the pygeoprocessing-approved NoData value for its datatype (e.g., `32767` for an `int16` raster) would be fine.

dcdenu4: I _think_ gdal handles different nodata values for different bands. See the "UNIFIED_SRC_NODATA" option from [gdalwarpoptions](https://gdal.org/en/stable/api/gdalwarp_cpp.html#_CPPv415GDALWarpOptions).

src/pygeoprocessing/geoprocessing.py
```diff
@@ -1102,6 +1102,44 @@ def align_and_resize_raster_stack(
...
+                    "and resizing.", base_path, chosen_nodata)
+            else:
+                LOGGER.warning(
+                    "Input raster %s has no defined nodata value. Setting "
+                    "%s as nodata in a temporary VRT before aligning and "
+                    "resizing.", base_path, chosen_nodata)
+
+            vrt_path = os.path.join(temp_working_dir_nodata,
+                                    f'input_with_nodata_{index}.vrt')
+            vrt = gdal.Translate(vrt_path, raster, format='VRT',
```
dcdenu4: I'm wondering if the creation of the VRT is necessary or if we could take advantage of `gdal_warp_options` and the `dstnodata` (?) argument. Like the resample list, could we keep a list of destination nodata values and pass those along through the gdal_warp_options?

I'm not sure it's a better approach than the VRT one other than it reduces a call to `gdal.Translate`.

dcdenu4 (commented):
Thanks @claire-simpson , I had a comment about taking a look at gdalwarpoptions, to see if that also could be a clean way to pass along the desired destination nodata value to `gdal.Warp`.

src/pygeoprocessing/geoprocessing.py
claire-simpson: Unfortunately that doesn't seem to work! Not sure why, because I can't seem to find documentation saying this doesn't work for geotiffs, but I do get this warning `Warning 1: multi_diffND.tif, band 2: Setting nodata to 255 on band 2, but band 1 has nodata at 9999. The TIFFTAG_GDAL_NODATA only support one value per dataset. This value of 255 will be used for all bands on re-opening`

src/pygeoprocessing/geoprocessing.py
claire-simpson: Related to my comment above, I've tried using the `UNIFIED_SRC_NODATA` and the `srcNodata` and `dstNodata` warp options, and have tried to explicitly set no data for each band via `band1.SetNoDataValue(255); band2.SetNoDataValue(9999)`, but both of these result in an output tif that has the same nodata value for each band. Using a VRT was the only way I could create a file with different nodata values for each band.

dcdenu4: That's interesting. If you think this is some kind of GDAL bug and had a toy example for replicating, I'd encourage you to make an issue on the GDAL project. Emily S. has done this a few times and gotten quick feedback.

>Using a VRT was the only way I could create a file with different nodata values for each band.

But it looks like we're taking the same approach of applying one NoData value to the VRT? At least, the first `warning` message is communicating this. IMO, I think it would be fine for this function to treat all bands in a raster with the same NoData value, as long as it's clearly communicated in the docstring.

If we're not worried about preserving various band NoData values, do you think it'd be worth using the built in nodata setting option via `gdal.Warp`?

dcdenu4 (commented):
Just a clarifying question / comment on whether we are or want to support unique NoData values for multiple bands in a raster.

src/pygeoprocessing/geoprocessing.py
claire-simpson: Here is the documentation I was looking for saying that all bands in a geotiff must have the same nodata value: https://gdal.org/en/stable/drivers/raster/gtiff.html#nodata-value. It does seem geotiff specific but that is what we encounter the most (not to mention the case of different NoData values in a raster should be uncommon), so I think its fine if we just support a single NoData value

dcdenu4 (approved):
Thanks Claire!

PR #482, opened 2026-04-29
Constrain gdal to one minor version in wheel build

### LINKED ISSUE #481: GDAL lib version can conflict
If a pygeoprocessing wheel is installed in an environment with a different version of libgdal than it was built against, importing the compiled modules will fail with an error like
```
ImportError: dlopen(/Users/emily/miniforge3/envs/test/lib/python3.14/site-packages/pygeoprocessing/geoprocessing_core.cpython-314-darwin.so, 0x0002): Library not loaded: @rpath/libgdal.36.dylib
  Referenced from: <0695A1F7-F0B6-388C-AE40-6A1791BBA087> /Users/emily/miniforge3/envs/test/lib/python3.14/site-packages/pygeoprocessing/geoprocessing_core.cpython-314-darwin.so
  Reason: tried: '/Users/emily/miniforge3/envs/test/lib/libgdal.36.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/emily/miniforge3/envs/test/lib/libgdal.36.dylib' (no such file), '/Users/emily/miniforge3/envs/test/lib/libgdal.36.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/emily/miniforge3/envs/test/lib/libgdal.36.dylib' (no such file), '/Users/emily/miniforge3/envs/test/bin/../lib/libgdal.36.dylib' (no such file), '/Users/emily/miniforge3/envs/test/bin/../lib/libgdal.36.dylib' (no such file)
```
This is the same as what's happening in natcap/invest#2206, and I propose fixing in the same way as I described in [this PR](https://github.com/natcap/invest/pull/2487).

emlys:
Fixes #481 

Update the GHA wheel build process to add a GDAL constraint by unpacking the wheel after it's built, modifying the metadata, and repacking it.

This achieves the goals of requiring the same minor version of GDAL that the wheel was built against, without applying that constraint to _all_ possible distributions, and without triggering a change to the setuptools_scm version.

I first tried the approach I proposed in https://github.com/natcap/invest/pull/2487#issuecomment-4238916610 (overwriting `prepare_metadata_for_build_wheel`). But it seems there's no guarantee that the build frontend actually invokes `prepare_metadata_for_build_wheel` when building a wheel: `pip` does, `build` does not. I think you would have to overwrite the `build_wheel` function itself, and I didn't find a straightforward way to implement that.

This approach achieves the same result, except that it's isolated to our build process in Github Actions, so won't affect a wheel build anywhere else. That could be a good or bad thing.


phargogh (approved):
Wow, that is a nicely elegant solution, independent of whether a build backend honors our hooks!!  I like this a _lot_.  Well done and thank you @emlys !

PR #487, opened 2026-07-22
Add `pygeoprocessing.include_dir()` and lazily import the dependency tree

### LINKED ISSUE #480: Provide a `get_include` function
`numpy` provides a `get_include()` function, which returns the path to the directory containing numpy header files that you might want to compile against. We call `numpy.get_include()` in the `natcap.invest` setup.py in order to pass that include directory to the compiler.

We're now doing a similar thing with `pygeoprocessing`, compiling against `ManagedRaster`, which currently looks like `os.path.join(pygeoprocessing.__path__[0], 'extensions')`. It would be nice if pygeoprocessing provided its own `get_include` function so that other libraries don't need to know anything about pygeoprocessing's directory structure.

Time estimate 1 - 3 hours.

phargogh:
This PR adds a new function, `pygeoprocessing.get_include()` that returns the path to the folder that should be passed to a compiler when compiling projects like `natcap.invest` against pygeoprocessing.

Although this feature by itself is useful, the experience of working in cross-compiled environments like conda-forge (https://github.com/conda-forge/natcap.invest-feedstock/pull/60) has also made it clear that it is _very_ useful to avoid importing pygeoprocessing's whole dependency tree at compile time.  To try to resolve this, I have updated the import functionality in this PR so that imports are lazily loaded only when attributes are actually called.  This behavior is discouraged in the python docs because import-time errors won't show up as readily, but I think the tradeoff here is worth it.   From an outside view, no functionality should change as a result of this PR except when, exactly, a module is imported.

Fixes #480

### DIFF
--- a/src/pygeoprocessing/__init__.py
+++ b/src/pygeoprocessing/__init__.py
@@ -9,12 +9,9 @@ import os
 import importlib
 import functools
 
-try:
-    from importlib.metadata import PackageNotFoundError
-    from importlib.metadata import version
-except ImportError:
-    from importlib_metadata import version
-    from importlib_metadata import PackageNotFoundError
+from importlib.metadata import PackageNotFoundError
+from importlib.metadata import version
+import lazy_loader
 
 LOGGER = logging.getLogger(__name__)
 LOGGER.addHandler(logging.NullHandler())  # silence logging by default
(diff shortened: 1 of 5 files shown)


src/pygeoprocessing/__init__.py
```diff
@@ -60,26 +31,122 @@
...
-                and attrname not in exclude_set):
-            __all__ += (attrname,)
-    except AttributeError:
-        pass
+def _lazy_import(modulename):
+    """Lazily import a module.
 
-LOGGER = logging.getLogger(__name__)
-LOGGER.addHandler(logging.NullHandler())  # silence logging by default
+    The returned modbule object will not actually be imported until a module
```
emlys:
```suggestion
    The returned module object will not actually be imported until a module
```

emlys:
@phargogh I came across this [lazy loading document](https://scientific-python.org/specs/spec-0001/) and corresponding [`lazy_loader` package](https://github.com/scientific-python/lazy-loader) and I was wondering if it might be worth using a preexisting solution like that? It's nice that it's backed by a scientific python organization and it's used by some major packages. On the other hand, it's relatively new, and the implementation is pretty simple and similar to what you have here already.

I also noticed their README says,
> We recommend using `lazy-loader` with Python >= 3.11. If using Python 3.11, please upgrade to 3.11.9 or later. If using Python 3.12, please upgrade to 3.12.3 or later. These versions [avoid](https://github.com/python/cpython/pull/114781) a [known race condition](https://github.com/python/cpython/issues/114763).

I wonder if that race condition issue also applies to the implementation in this PR (just to be aware of)?

phargogh:
Thanks @emlys !  I rather like that `lazy-loader` will handle a bunch of the complexity for us, so I went ahead and moved the `__init__.py` file to use that approach instead.  One minor consequence of this is that we can now only support python 3.9+, which shouldn't really be a big deal since that's the earliest version we're currently testing against anyways and 3.9 is out of the support window.

>I also noticed their README says,
>
>>We recommend using lazy-loader with Python >= 3.11. If using Python 3.11, please upgrade to 3.11.9 or later. If using >>Python 3.12, please upgrade to 3.12.3 or later. These versions https://github.com/python/cpython/pull/114781 a >>https://github.com/python/cpython/issues/114763.
>
>I wonder if that race condition issue also applies to the implementation in this PR (just to be aware of)?

I haven't tested this specifically, but just reading through the issue and PR I think yes, this would also be important for us to be aware of and is likely to be an issue in the specified versions of python.  It looks like the issue is the underlying LazyLoader object, which we were also using in our implementation.

emlys (commented):
It looks like there are a few imports in `__init__.py` that are now unused. Looks good other than that!

phargogh:
Boy, there sure are!  I've been fiddling with my editor, so I really need to get in-nvim linting back up and running to catch this kind of issue.  Patched in b69c9118

emlys (approved):
👍

